PR #9, opened 2023-01-31
FM-8: FVM and built-in actor dependencies

### LINKED ISSUE #8: FVM and built-in actors dependency
Add a dependency to `ref-fvm` and the `builtin-actors`. We probably won't need all builtin actors, but most things depend on things like the `InitActor`, so I don't think we can _not_ load an actor bundle.

aakoshh:
Closes #8 

Added `ref-fvm` stuff and the `next` branch of `builtin-actors` (a.k.a. actors V10) as dependencies and tried if they can be loaded into RocksDB.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,52 @@
 version = 3
 
 [[package]]
+name = "addr2line"
+version = "0.17.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


adlrocha (approved):
Nice! I am working on a modified bundle of builtin-actors including the IPC actors. Right now I am building them manually, but I am thinking about writing a `Makefile` or something so they can be built and pulled easily.

aakoshh:
Yes I agree, they should be pulled as a CAR from somewhere because building everything takes ages.

PR #12, opened 2023-02-01
FM-11: Message model

### LINKED ISSUE #11: Self-contained transaction model
Self-contained transactions are what we call `Message` in Filecoin. They can be passed to the FVM for execution more or less as they are. The non-self-contained transactions will come in Milestone 2 where it will be possible to put transactions into that contain CIDs which need to be resolved first, which is where we'll make use of ABCI++. For Milestone 1, we stick to the vanilla use case. 

Look at what a `Message` looks like in Forest and bring it over to this project, with all the related data model such as addresses and signatures.

aakoshh:
Closes #11 

The PR brings over parts of the [message](https://github.com/ChainSafe/forest/tree/main/vm/message) crate from Forest. Only the `SignedMessage` was copied more or less as-is; the `ChainMessage` also got an equivalent, but its contents are different.

The location mirrors the `vm` directory in Forest, which I thought might provide some familiar structure, with the eventual `interpreter` dealing with the FVM. But, those Forest modules are not something which I can reuse: they are built against FVM v2 (as opposed to v3 which we are targeting here) and already reflect some parts of Expected Consensus.

NB the `Cbor` trait has been removed from the `builtin-actors` and `ref-fvm`.

### DIFF
--- a/fendermint/vm/message/src/chain_message.rs
+++ b/fendermint/vm/message/src/chain_message.rs
@@ -1,0 +1,32 @@
+// Copyright 2022-2023 Protocol Labs
+// SPDX-License-Identifier: Apache-2.0, MIT
+use serde::{Deserialize, Serialize};
+
+use crate::signed_message::SignedMessage;
+
+/// The different kinds of messages that can appear in blocks, ie. the transactions
+/// we can receive from Tendermint through the ABCI.
+///
+/// Unlike Filecoin, we don't have `Unsigned` messages here. In Filecoin, the messages
+/// signed by BLS signatures are aggregated to the block level, and their original
+/// signatures are stripped from the messages, to save space. Tendermint Core will
+/// not do this for us (perhaps with ABCI++ Vote Extensions we could do it), though.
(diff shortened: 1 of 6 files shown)


fendermint/vm/message/src/chain_message.rs
```diff
@@ -0,0 +1,32 @@
...
+
+/// The different kinds of messages that can appear in blocks, ie. the transactions
+/// we can receive from Tendermint through the ABCI.
+///
+/// Unlike Filecoin, we don't have `Unsigned` messages here. In Filecoin, the messages
+/// signed by BLS signatures are aggregated to the block level, and their original
+/// signatures are stripped from the messages, to save space. Tendermint Core will
+/// not do this for us (perhaps with ABCI++ Vote Extensions we could do it), though.
+#[derive(Clone, Debug, Serialize, Deserialize, Hash)]
+#[serde(untagged)]
```
aakoshh, line 15: This will probably have to change if we want to distinguish between the future CID-based cases.

PR #23, opened 2023-02-07
FM-16: Storage abstraction

aakoshh:
Related to #16 

This PR brings over some code I developed a while ago when I wanted to use RocksDB with transactions, and column families. It creates a `KVStore` abstraction that we can implement, with `KVReadable` and `KVWriteable` traits to show whether something requires read-only or read-write access. Further to that, it defines `Encode`, `Decode` and `Codec` traits to do the actual serialization. 

Currently in Forest they use a `DB: Blockstore + Store` generic parameter to access IPLD data under the default namespace and also to store arbitrary key-value pairs as bytes, for example [here](https://github.com/ChainSafe/forest/blob/v0.6.0/blockchain/chain/src/store/chain_store.rs#L704). 

```rust
pub fn set_genesis<DB>(db: &DB, header: &BlockHeader) -> Result<Cid, Error>
where
    DB: Blockstore + Store,
{
    db.write(GENESIS_KEY, header.marshal_cbor()?)?;
    db.put_obj(&header, Blake2b256)
        .map_err(|e| Error::Other(e.to_string()))
}
```

I wanted to achieve the following things with this PR:
* Be able to store things like this genesis header, or the tip of the chain, or the application state, separately from the IPLD data, in a different column family. It seems cleaner and the IPLD store can grow arbitrarily large.
* Be able to use transactions, when overwriting multiple entries. This is not an issue with the IPLD part because it's content addressed and append-only, but in other cases, if the application crashed midway, it might be left in an inconsistent state. 
* Provide a strongly typed interface to the data store using the `KVCollection` type, to avoid having to serialize data from bytes in the call site. 

The idea is that a function like `set_genesis` would look something like this:

```rust
pub fn set_genesis<DB>(db: &DB, header: &BlockHeader) -> Result<Cid, Error>
where
    DB: Blockstore + KVStore<S, Namespace=&str> + KVWriteable<DB> + Encode<BlockHeader> + Encode<&str>,
{
    let mut tx = db.write();
    tx.put("genesis", "header", header)?;
    let cid = db.put_obj(&header, Blake2b256)
        .map_err(|e| Error::Other(e.to_string()))?;
    tx.prepare_and_commit();
   Ok(cid)
}
```

It is more cumbersome in this small context; the boilerplate would be manageable by using `KVCollection` and starting transactions further out further out, which allows multiple operations to be chained together into an atomic isolated execution.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -503,7 +503,7 @@ dependencies = [
  "core2",
  "multibase",
  "multihash",
- "quickcheck",
+ "quickcheck 0.9.2",
  "rand 0.7.3",
  "serde",
  "serde_bytes",
@@ -1065,6 +1065,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "env_logger"
+version = "0.8.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a19187fea3ac7e84da7dacf48de0c45d63c6a76f9490dae389aead16c243fce3"
+dependencies = [
+ "log",
+ "regex",
+]
+
+[[package]]
 name = "errno"
 version = "0.2.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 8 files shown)


adlrocha (approved):
skimmed through it, treat it as a rubber stump, but it looks really nice. Thanks!

PR #24, opened 2023-02-09
FM-16: RocksDB

aakoshh:
Related to #16 

Implements `KVStore` for RocksDB. Unfortunately I could not use the Forest RocksDB library any more, because even if I specified the `multi-threaded-cf` feature which is required for `.transaction()`, I had to create an `OptimisticTransactionDB`, not a `DB`, so I ended up copying a lot of the Forest code. 
 
The PR changes `App` to use `KVStore`. 

In this instance, after so much work, it doesn't make a difference because we don't modify the app state concurrently, nor do we perform multiple steps during writes/reads. Still, it felt right to put the data into a different column family, and at least this way we can use the same abstraction elsewhere in the application, like for relayers. Another area is where it can help is supporting queries by maintaining a list of the past N state roots. If it's slow, we can always go back. 

I will add tests in a follow up PR.

### DIFF
--- a/fendermint/app/src/store.rs
+++ b/fendermint/app/src/store.rs
@@ -1,27 +1,21 @@
-use std::{borrow::Cow, hash::Hash, marker::PhantomData};
+use std::borrow::Cow;
 
 use fendermint_storage::{Codec, Decode, Encode, KVError, KVResult, KVStore};
 use fvm_ipld_encoding::{de::DeserializeOwned, serde::Serialize};
 
 #[derive(Clone)]
-pub struct AppStore<NS> {
-    _ns: PhantomData<NS>,
-}
+pub struct AppStore;
 
-impl<NS> KVStore for AppStore<NS>
-where
-    NS: Eq + Hash + Clone,
-{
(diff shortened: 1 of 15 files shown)


fendermint/app/src/store.rs
```diff
@@ -0,0 +1,38 @@
...
+    }
+}
+
+/// CBOR deserialization.
+impl<T> Decode<T> for AppStore
+where
+    T: DeserializeOwned,
+{
+    fn from_repr(repr: &Self::Repr) -> KVResult<T> {
+        fvm_ipld_encoding::from_slice(repr).map_err(|e| KVError::Codec(Box::new(e)))
```
aakoshh, line 36: @adlrocha I can't remember now if the restriction on String keys in maps only affects HAMT fields or everything you ever want to store.

aakoshh: I think it's just the HAMT which serializes as CBOR but deserializes as IPLD. Just confusing that we are using `fvm_ipld_encoding` here.

PR #26, opened 2023-02-10
FM-25: Remove KVPreparedTransaction

### LINKED ISSUE #25: Simplify the prepare/commit phases to just try_commit
A followup on #16 

According to the RocksDB documentation, for `OptimisticTransactionDB`:
* `commit`: will commit changes unless there is a conflict, in which case it will return `Busy`
* `prepare`: writes to the WAL, so commit is simple, as the cost of making `rollback` more costly; it doesn't say that it would return `Busy` if there are conflicts

See:
* https://www.bookstack.cn/read/rocksdb-6.14-en/4b5d7163a69af0cb.md?wd=2nd
* https://docs.rs/rocksdb/latest/rocksdb/struct.Transaction.html#method.commit

If `commit` is the one which returns `Busy` then this is at odds with how [STM expects it](https://github.com/aakoshh/async-stm-rs/blob/master/src/transaction.rs#L133) to work. Also, in STM we do the database prepare first, and the in-memory key checks second. We could do the other way around, though: 
1. locks the keys in the STM transaction first, detect conflicts there
2. if a conflict is found, rollback the DB
3. if there are no conflicts, then `try_commit` the database
4. if `try_commit` returned `true`, write to the locked in-memory places
5. if `try_commit` returned `false`, release the locks and try again

aakoshh:
Closes #25

The PR removes the `prepare()` method because it works differently in RocksDB than I imagined. Unlike `put`, `delete` and `commit`, the docs do not mention that it could return `Busy`, so I imagine that it does not check for conflicts. It might do, but it doesn't say, just that it prepares for two phase commit by writing changes to the WAL. 

I got rid of it in favor of allowing `commit` to return `KVResult::Conflict` to signal that it failed. I will update STM to expect this workflow.

### DIFF
--- a/fendermint/app/src/app.rs
+++ b/fendermint/app/src/app.rs
@@ -7,9 +7,7 @@ use std::sync::{Arc, Mutex};
 use async_trait::async_trait;
 use cid::Cid;
 use fendermint_abci::Application;
-use fendermint_storage::{
-    Codec, Encode, KVRead, KVReadable, KVStore, KVTransaction, KVWritable, KVWrite,
-};
+use fendermint_storage::{Codec, Encode, KVRead, KVReadable, KVStore, KVWritable, KVWrite};
 use fendermint_vm_interpreter::bytes::BytesMessageApplyRet;
 use fendermint_vm_interpreter::chain::ChainMessageApplyRet;
 use fendermint_vm_interpreter::fvm::{FvmApplyRet, FvmState};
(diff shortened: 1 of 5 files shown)


aakoshh:
STM updated in https://github.com/aakoshh/async-stm-rs/commit/bba36f7123985059d26c414597bc1248f15bf886

PR #31, opened 2023-02-15
FM-30: Binary format tests

### LINKED ISSUE #30: Message format tests
The messages need to be serialized to a binary format, which in our case is IPLD. 

We need tests that demonstrate that:
* arbitrary messages can be serialized and deserialized with complete fidelity, that the format isn't lossy
* save examples as test vectors or golden files that we can point to as reference and protect against regressions

At this point we only handle the signed messages, however, to make convincing tests for chain messages, it would be good to include the other two planned examples (just reject them in handlers), because some of the serde annotations tell it to use untagged types, and I'm not sure that will work if there are more than one possible types.

aakoshh:
Related to #30 

The PR contains `Arbitrary` instances for messages and a quickcheck test to make sure a `ChainMessage` can be serialized to/from binary as intended with IPLD.

#### Invalid arbitrary data from upstream

I had to fix some of the arbitrary data generated by the upstream libraries: 
* `TokenAmount` has a `BigInt` that can be too large to serialize according to the Filecoin spec.
* `DelegatedAddress` has a `length` field that had a value inconsistent with how many bytes it had filled with non-zero values. 

I don't know if these were done like this deliberately as "completely arbitrary" or they are a result of oversight.

#### Reserved message types

I added two `ChainMessage` types that I do not intend to handle in Milestone 1, but I wanted to see if they work with the serialization.

#### Copied arbitrary CID generation

There is a version conflict with the `cid` and `multihash` libraries in that the versions used by `ref-fvm` rely on an earlier version of `quickcheck`, which makes those fancy `Arbitrary` instances unusable. I copied them so that I can generate messages with CIDs. 

Only after did I notice that Forest also has a [simpler arbitrary CID](https://github.com/ChainSafe/forest/blob/v0.6.0/blockchain/blocks/src/lib.rs) that only generates one version and doesn't try to do weighted sampling. Oh well.

### DIFF
--- a/fendermint/vm/message/src/chain.rs
+++ b/fendermint/vm/message/src/chain.rs
@@ -1,3 +1,4 @@
+use cid::Cid;
 // Copyright 2022-2023 Protocol Labs
 // SPDX-License-Identifier: Apache-2.0, MIT
 use serde::{Deserialize, Serialize};
@@ -12,7 +13,6 @@ use crate::signed::SignedMessage;
 /// signatures are stripped from the messages, to save space. Tendermint Core will
 /// not do this for us (perhaps with ABCI++ Vote Extensions we could do it), though.
 #[derive(Clone, Debug, Serialize, Deserialize, Hash, PartialEq)]
-#[serde(untagged)]
 pub enum ChainMessage {
     /// A message that can be passed on to the FVM as-is.
     Signed(SignedMessage),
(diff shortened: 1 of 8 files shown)


fendermint/vm/message/src/chain.rs
```diff
@@ -11,8 +12,7 @@ use crate::signed::SignedMessage;
 /// signed by BLS signatures are aggregated to the block level, and their original
 /// signatures are stripped from the messages, to save space. Tendermint Core will
 /// not do this for us (perhaps with ABCI++ Vote Extensions we could do it), though.
-#[derive(Clone, Debug, Serialize, Deserialize, Hash)]
-#[serde(untagged)]
```
aakoshh, line 15: This has been removed, because if we leave it we get:
```
Error { description: "Msg(\"Only bytes can be deserialized into a CID\")", protocol: Cbor }
```
This was expected, and exactly why I wanted to add these types even though I won't use them yet. 

What I didn't expect was that I'd get this error even if there was only one variant wrapping a CID. I thought serde [will try one, then the other](https://serde.rs/enum-representations.html#untagged), and return whichever variant works. 

I am not sure what the tagging mechanism will look like when combined with IPLD. I can generate examples but can't write a "spec".

PR #97, opened 2023-05-10
FM-95: Automated smoke test

### LINKED ISSUE #95: Automated smoke test
Use `cargo-make` similar to how it's used in `tendermint-rs` to stand up a Tendermint container, a Fendermint container, run the simplecoin example, then tear down the containers.

aakoshh:
Closes #95 

Adds a `fendermint/testing/smoke-test` crate with [cargo-make](https://github.com/sagiegurari/cargo-make) that sets up Fendermint and Tendermint as docker containers and runs the simplecoin example.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -65,7 +65,7 @@ jobs:
           profile: minimal
           target: wasm32-unknown-unknown
           toolchain: ${{ matrix.rust }}
-          components: rustfmt,clippy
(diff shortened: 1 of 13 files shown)


aakoshh:
Maybe it would be worth adding a condition to only run this test if a label is applied to the PR: https://stackoverflow.com/questions/62325286/run-github-actions-when-pull-requests-have-a-specific-label

PR #109, opened 2023-05-18
FM-106: Upgrade FVM version to 3.4

### LINKED ISSUE #106: Upgrade ref-fvm to 3.3.x

aakoshh:
Closes #106

We can upgrade to 3.2, but to go to 3.3 or 3.4 is thwarted by `frc42_dispatch` depending on 3.2

We can try to remove that dependency by hardcoding the hash, like the FVM integration tests do :persevere: 

We only have 1 method to hash, so this isn't too bad at the moment, but if in the future there would be more, I would switch back to using the hasher.

Perhaps it was a one-off, but the test times were faster without this dependency (the 3.2 and 3.4 tests were running concurrently, so it's not down to caching).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,5 +1,14 @@
 [workspace]
-members = ["fendermint/abci", "fendermint/app", "fendermint/rocksdb", "fendermint/rpc", "fendermint/storage", "fendermint/testing", "fendermint/testing/*-test", "fendermint/vm/*"]
+members = [
+  "fendermint/abci",
+  "fendermint/app",
+  "fendermint/rocksdb",
+  "fendermint/rpc",
+  "fendermint/storage",
+  "fendermint/testing",
+  "fendermint/testing/*-test",
+  "fendermint/vm/*",
+]
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -46,20 +55,28 @@ tracing-subscriber = "0.3"
...
+
 # Using 0.8 because of ref-fvm.
 # 0.9 would be better because of its updated quickcheck dependency.
 # 0.10 breaks some API.
 cid = { version = "0.8", features = ["serde-codec", "std"] }
 
-# The following are on crates.io but they might be pre-releases.
-fvm = { version = "3.1", default-features = false }     # no opencl feature or it fails on CI
-fvm_shared = { version = "3.1", features = ["crypto"] }
+# Depending on the release cycle, this dependency might want an earlier version of the FVM.
```
adlrocha, line 70: I don't know if we will regret it in the long-run, but something we could do is to bring all the frc42 that we need to https://github.com/consensus-shipyard/fvm-utils, so we can upgrade accordingly. This is the approach we are following for actor runtimes. We also use extensively frc42 in actors, so maybe this helps. Instead of using the external dependency we can fork the relevant code that we need into our `fvm-utils`.

(Been there, and I hate this dependency dance, is a time drain)

aakoshh: Hm, I didn't know you use it that much. It was released last month, so it doesn't seem to lag _that_ much. Would it be possible to jump in and do a release, could they just be busy? 

But yeah, if there is pain, we have that option too. I don't follow what features there are in the FVM releases, so I'm not sure we need to be on the bleeding edge. For now I think we can save ourselves the extra burden.

PR #157, opened 2023-07-20
Add ETH API launch doc

cryptoAtwill:
Add more doc on how to launch ETH api.

### DIFF
--- a/docs/running.md
+++ b/docs/running.md
@@ -410,6 +410,8 @@ Note that the first block execution is very slow because we have to load the Was
 but after that the blocks come in fast, one per second.
 
 ### Run ETH API
+If we want to use `evm` related API, such as running `fendermint/eth/api/examples/ethers.rs`, we need to start ETH API process. 
+
 The ETH RPC api runs on top of cometbft. Make sure you have cometbft running properly. The architecture is as follows:
 ```
 +---------------------------+


aakoshh (approved):
Thanks!

adlrocha (approved):

docs/running.md
```diff
@@ -409,6 +409,62 @@ I[2023-05-19|10:13:54.116] indexed block exents                         module=t
...
+|   |                   |   |
+|   | :8545             |   |
+|   +-------------------+   |
+|     |                     |
+| :8545                     |
+-----------------------------
+```
+To start the ethereum RPC api with:
+```
+cargo run -p fendermint_app --release -- eth run
```
adlrocha, line 451: In the future I suggest having a `make build`  that outputs the binaries into a `./bin` directory so we can just `./eth run` or something like this. But we can have a follow-up with this once we prepare for the first release.

aakoshh: The readme lists a couple of ways you can run this. I used `cargo run` because that is one less step that needs to happen when I'm doing iterative debugging (no need to make build which might do more than I need, and then run).

aakoshh: But we can agree to have more streamlined commands in the README. In the demo I used the cargo install version, so there's no need for a bin directory that is mostly unused. An alias works too.

adlrocha: Sounds good, no strong opinions here. For me this is good to merge.

PR #158, opened 2023-07-20
FM-156: IPC actor bindings

### LINKED ISSUE #156: Add IPC actor bindings to the repo
IPC Solidity actors live here: https://github.com/consensus-shipyard/ipc-solidity-actors

The IPC Agent has copies of the generated ABI here: https://github.com/consensus-shipyard/ipc-agent/tree/main/contracts

These are used with `abigen!` [here](https://github.com/consensus-shipyard/ipc-agent/blob/v0.4.0/src/manager/evm/manager.rs#L38-L41). 

In my experience this doesn't work with Rust Analyzer, we just get `{unknown}` and no code completion. If we actually generate the Rust artifacts, it is more helpful. Since these contracts are more essential to what we are doing than, say, the SimpleCoin contract I used in examples, I think it's worth adding them to the repo.

aakoshh:
Closes #156 

* Created a `make ipc-actors-abi` entry to clone the `ipc-solidity-actors` repo, compile the ABI artifacts, then create Rust bindings
* Created a `fendermint_vm_ipc_actors` crate with a `build.rs` file that generates Rust modules from the above. 
* Copying the ABI artifacts into the Docker image, so that we can later deploy them during genesis

Unfortunately common data structures like `SubnetID` are repeated across all three modules. Not sure what can be done about that. Perhaps we can generate bindings for more of the contract files, and add `From` instances manually or with macros for all the mirrored ones. (This would require removing the auto-generation of `lib.rs`).


Makefile
```diff
@@ -73,3 +81,31 @@ $(BUILTIN_ACTORS_BUNDLE): $(BUILTIN_ACTORS_CODE)
...
+# Clone the IPC Solidity actors if necessary and compile the ABI, putting down a marker at the end.
+$(IPC_ACTORS_ABI): $(IPC_ACTORS_CODE) | forge
+	if [ ! -d $(IPC_ACTORS_DIR) ]; then \
+		mkdir -p $(IPC_ACTORS_DIR) && \
+		cd $(IPC_ACTORS_DIR) && \
+		cd .. && \
+		git clone https://github.com/consensus-shipyard/ipc-solidity-actors.git; \
+	fi
+	cd $(IPC_ACTORS_DIR) && \
+	git checkout fm-156-ipc-solidity-actors && \
```
adlrocha, line 99: Should we point to main or not checkout at all? (I guess you are doing this for testing, right?)
```suggestion
	git checkout main&& \
```

Makefile
```diff
@@ -73,3 +81,31 @@ $(BUILTIN_ACTORS_BUNDLE): $(BUILTIN_ACTORS_CODE)
 	git pull && \
 	rustup target add wasm32-unknown-unknown && \
 	cargo run --release -- -o output/$(shell basename $@)
+
+
+# Compile the ABI artifacts and Rust bindings for the IPC Solidity actors.
+ipc-actors-abi: $(IPC_ACTORS_ABI)
```
adlrocha, line 87: @cryptoAtwill, should add a similar makefile rule to the agent?

fendermint/vm/ipc_actors/build.rs
```diff
@@ -0,0 +1,87 @@
...
+// SPDX-License-Identifier: Apache-2.0, MIT
+
+use std::io::Write;
+use std::path::{Path, PathBuf};
+
+/// Generate Rust bindings from the IPC Solidity Actors ABI artifacts.
+///
+/// These are built by `make ipc-actors-abi`, here we just add the final step
+/// so we have better code completion with Rust Analyzer.
+fn main() {
```
adlrocha, line 11: @cryptoAtwill, actually, what about removing the type bindings from the agent and leveraging the ones in Fendermint (the same way that we leverage the ones from `ipc-actors` to interact with the native actors?

adlrocha (approved):
I added a few questions to @cryptoAtwill, and a potential typo inherited from testing, but LGTM. Thanks!

Makefile
aakoshh: Oh, you merged the fix?

aakoshh: As far as I see the 3 methods are still in `IGateway.sol`, because this hasn't been merged: https://github.com/consensus-shipyard/ipc-solidity-actors/pull/138

Until that's merged, this branch here has the 3 methods you mentioned commented out, so the size stays below 25kB and doesn't fail the entire build.

PR #160, opened 2023-07-25
transfer RPC value in full FIL

adlrocha:
It was a bit counter-intuitive to have to send 10^18 to send a single Filecoin that can be interpreted  by Metamask and the Ethereum tooling. This PR accepts a float as the input for `value` in `transfer`. 

Optionally, we should consider  accepting also full FIL in the argument of genesis creation (I have a weaker opinion here, so no need to address it immediately). 

The PR also adds `cometbft` and `test-network` in `.gitignore` to prevent devs from mistakenly pushing these dirs.


fendermint/app/src/options/rpc.rs
```diff
@@ -145,8 +147,8 @@ pub struct TransArgs {
     /// Name of chain the for which the message will be signed.
     #[arg(long, short, env = "FM_CHAIN_NAME")]
     pub chain_name: String,
-    /// Amount of tokens to send, in atto.
-    #[arg(long, short, value_parser = parse_token_amount, default_value = "0")]
+    /// Amount of tokens to send, in full FIL, not atto.
```
aakoshh, line 150: When I read "full FIL" I imagine it will be an _integer_ that contains whole FIL amounts, 1, 2, 3, etc. Maybe "fractional FIL"? But if "full FIL" is the adopted nomenclature it's okay, now I know.

aakoshh (commented):
Nice, definitely more usable than atto!

Wouldn't it make sense to replace all CLI parsers with this?

aakoshh:
I see you got the full Gateway contract to compile as well :+1:

fendermint/app/src/options/rpc.rs
adlrocha: I usually call `full_fil` because it is not subdivided into `atto` or other subunits, but weak opinion weakly held. If you think it is clearer as `fractional_FIL` I am happy to change it.

adlrocha:
> Wouldn't it make sense to replace all CLI parsers with this?

Maybe you are right, I didn't want to mess up with the rest of the commands, but I can change it for everything to be in FIL instead of atto.

adlrocha:
@aakoshh, new review before merging? I decided to have full fil units for balances, while using atto for gas parameters. Let me know if you have any comments. Thanks!

docs/demos/milestone-1/fendermint-demo.sh
```diff
@@ -25,7 +25,7 @@ cat test-network/keys/alice.pk
 fendermint \
   genesis --genesis-file test-network/genesis.json \
   add-account --public-key test-network/keys/alice.pk \
-              --balance 1000000000000000000
+              --balance 1000
```
aakoshh, line 29: I'm not sure I would alter this demo script to be honest. The README in the milestone directory says these commands reflect the state as it was back then, although I was too lazy to include a commit, but it should point for example to 7f4c155f35b374327a33c7fccec6df0e71870f07 which is one before the switch to CometBFT.

I wanted this script to be as it appeared in the video, a snapshot at Milestone 1, for reference, and only keep the `running.md` file up to date.

fendermint/app/src/options/genesis.rs
```diff
@@ -72,8 +72,8 @@ pub struct GenesisAddMultisigArgs {
     /// Path to the Secp256k1 public key exported in base64 format, one for each signatory.
     #[arg(long, short)]
     pub public_key: Vec<PathBuf>,
-    /// Initial balance in atto.
-    #[arg(long, short, value_parser = parse_token_amount)]
+    /// Initial balance in full FIL units.
+    #[arg(long, short, value_parser = parse_full_fil)]
     pub balance: TokenAmount,
     /// Number of signatures required.
     #[arg(long, short)]
```
aakoshh, line 79: What about `base_fee`?

aakoshh (commented):
The end-to-end tests fail now, but I'm not sure why. 

Do you want to keep gas stuff in ATTO so that you have 9 extra decimal places? For example a minimum of 1 nano FIL would be too much for gas premium or base fee? I have no idea what are realistic values here, just wondering if it would be confusing for users to remember which CLI parameter is in what unit.

docs/running.md
```diff
@@ -469,21 +469,21 @@ gVUBRpGWKyQvYeoOY3OJROSyogmA3ys=
 
 The simplest transaction we can do is to transfer tokens from one account to another.
```
aakoshh, line 470: The `balance` in Genesis in this file hasn't been updated.

adlrocha: Good catch, thanks!

aakoshh (approved):
Cool, thanks for reverting the demo script. If this is the approach in Lotus then let's do it!

PR #161, opened 2023-07-25
Fix estimate gas

cryptoAtwill:
When sending transactions, we cannot estimate gas due to invalid serialisation error. This is because `ethers-rust` requires the `transaction type` to be zero padded hex string while hardhat does not enforce zero padding. 

We propose a temporary solution as follows:
Since we only use `Eip1559TransactionRequest`, we created a new `TypedTransaction` that uses only `Eip1559TransactionRequest`. When deserializing, we use alias that handles both zero padding and non zero padding.

See https://github.com/gakonst/ethers-rs/pull/2518 for `ethers` fix for this issue.


aakoshh (approved):
Thank you for the fix!

adlrocha:
Sorry, folks. I didn't know this PR was already a thing and I was working on this same branch to build upon @cryptoAtwill changes to `TypedTransaction`, I didn't know they were final. I added some changes to support `LegacyTransactions` (nedded for Metamask and Remix support)

aakoshh (changes requested):
[call](https://github.com/consensus-shipyard/fendermint/blob/ad6797904ba837c75fcac76794f76d7d0e320b40/fendermint/eth/api/src/apis/eth.rs#L523) will need this fix as well.

fendermint/eth/api/src/conv/from_eth.rs
```diff
@@ -7,23 +7,17 @@ use anyhow::Context;
...
-            Ok(fendermint_vm_message::conv::from_eth::to_fvm_message(&tx)?)
-        }
-        TypedTransaction::Legacy(_) | TypedTransaction::Eip2930(_) => error(
-            ExitCode::USR_ILLEGAL_ARGUMENT,
-            "unexpected transaction type",
-        ),
-    }
+    Ok(fendermint_vm_message::conv::from_eth::to_fvm_message(
+        &tx.into(),
+    )?)
```
aakoshh, line 22: I'm not sure this will work during signature checks. 

This `tx.into()` calls [TypedTransaction::into_eip1559](https://github.com/gakonst/ethers-rs/blob/ethers-v2.0.8/ethers-core/src/types/transaction/eip2718.rs#L562) which doesn't set all the fields (leaves out `max_fee_per_gas` and  `max_priority_fee_per_gas` which the FVM actually uses) and leaves them on default values. Which in itself might be okay, we send the transaction with 0 fees, however, this happens _after_ the transaction has been signed by the client, so let's see how we check the signature.

To figure out what to check the signature against, we turn the FVM transaction back into a `TypedTransaction` and call [sighash](https://github.com/consensus-shipyard/fendermint/blob/master/fendermint/vm/message/src/signed.rs#L102) on it. Since we only expect Eip1559 transactions, that's exactly what we're are [going to use](https://github.com/consensus-shipyard/fendermint/blob/master/fendermint/vm/message/src/conv/from_fvm.rs#L107).

Then if we look at [TypedTransaction::sighash](https://github.com/gakonst/ethers-rs/blob/ethers-v2.0.8/ethers-core/src/types/transaction/eip2718.rs#L363) it calls [TypedTransaction::rlp](https://github.com/gakonst/ethers-rs/blob/ethers-v2.0.8/ethers-core/src/types/transaction/eip2718.rs#L337) which treats all variants differently, and in our case [prepends](https://github.com/gakonst/ethers-rs/blob/ethers-v2.0.8/ethers-core/src/types/transaction/eip2718.rs#L348) `0x2` to the RLP representation. 

Therefore if the transaction from the client _wasn't_ an Eip1559 request, you might turn it into one here, but it will fail the signature check because what has been signed by the client had a different, or no prefix.

It would be better to reject this transaction when we still understand why we are doing it.

aakoshh: And I know that with the current workaround these other transactions would be rejected by the API, but if we treat this as a temporary solution, when ethers is released with the fix, the above scenario will be a possibility.

adlrocha:
> [call](https://github.com/consensus-shipyard/fendermint/blob/ad6797904ba837c75fcac76794f76d7d0e320b40/fendermint/eth/api/src/apis/eth.rs#L523) will need this fix as well.

I had it in the backlog, but if you don't mind, I will solve this one on top of https://github.com/consensus-shipyard/fendermint/pull/163 so I can test the end-to-end with Remix.

docs/running.md
```diff
@@ -409,6 +409,62 @@ I[2023-05-19|10:13:54.116] indexed block exents                         module=t
...
+2023-07-20T12:30:48.385026Z  INFO fendermint::cmd: reading configuration path="/home/admin/.fendermint/config"
+2023-07-20T12:30:48.435387Z  INFO fendermint_eth_api: bound Ethereum API listen_addr=127.0.0.1:8545
+```
+
+</details>
+
+We can try query the chain id by:
+```shell
+curl -X POST -i   -H 'Content-Type: application/json'   -d '{"jsonrpc":"2.0","id":0,"method":"eth_chainId","params":[]}'   http://localhost:8545
+```
```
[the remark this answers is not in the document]
adlrocha: @cryptoAtwill, do you mind picking up this one?

fendermint/eth/api/src/conv/from_eth.rs
adlrocha: Resolved in the latest commit.

Issue #29, opened 2023-02-13
Enforce gas_fee_cap during checking


aakoshh:
Followup to https://github.com/consensus-shipyard/fendermint/pull/28 

The FVM allows the user to send messages with a fee cap which is lower than the base fee, even zero; the miner pays the difference. A validator will probably not want to foot all the bills, though, so the `FvmMessageInterpreter` should be able to reject messages which are not profitable. 

Look at what Forest is doing in the mempool and implement something similar.

raulk:
> The FVM allows the user to send messages with a fee cap which is lower than the base fee, even zero; the miner pays the difference.

Yes, but you seem to be tracking that in #147. Let's keep this issue about enforcing (a) correct inclusion behaviour, and (b) mempool queuing logic for when the base fee is higher than the gas fee cap?

At the mpool, this looks like implementing a bounded queue with some heuristics to size it appropriately (e.g. a mixture of count and message size). Today you can assume that most wallets will re-broadcast if a monitored tx gets dropped from the mpool, but to maximise QoS delivered by the netweork, you want to do your best to keep a set of messages around with reasonable criteria (e.g. that could be eligible for inclusion in a future window of N blocks if base fee consistently goes down, up to a max count and size each).

Following what Forest does makes sense, but at this stage we can probably afford to simplify if it's not easy to port the logic over.

aakoshh:
@raulk you are right, #147 is the issue for implementing the logic where the validator should pay the difference, with the caveat of having to specify what to do if they can't. 

In this ticket I just wanted to capture that the nodes should have a way of keeping unwanted transactions out of blocks in the first place, with a combination of `check_tx` and `prepare_proposal`. We can either reject transactions in `check_tx` so they don't enter the mempool of the node they were sent to (but other validators can still add them to the block, unless we explicitly don't vote for them, as discussed in #147), or we can select the best to be included in `prepare_proposal` - the rest will be retried by Tendermint later, they aren't removed from the mempool.

PR #163, opened 2023-07-28
Improve gas estimation logic

adlrocha:
Depends on: #161 

This PR improves the gas estimation logic. Instead of directly pre-executing the message to get the estimated gas of the call we: 
- Estimate the exact gas used for the execution. 
- Apply an overestimation. 
- Perform a gas search to see if the gas limit needs to be increased over the previous overestimation. 
Without this, tools like Metamask and Remix didn't estimate the gas correctly and message never went through. This approach is not as accurate as the fine-grained one used in Lotus, but for our use case is good enough to ensure that the transaction is included in the next block.  

Additionally, the gas estimation also populates the gas premium and gas limit of the transaction when not set. 

For the gas estimation logic, I had to include a new `call_with_cache` method to allow enabling and disabling the cache to determine if in the subsequent execution of message the results should be conveniently buffered or not.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,68 +19,69 @@ edition = "2021"
 license = "MIT OR Apache-2.0"
 license-file = "LICENSE-APACHE"
 
-
 [workspace.dependencies]
-anyhow = "1"
-arbitrary = { version = "1", features = ["derive"] }
-arbtest = "0.2"
-async-stm = "0.2"
-async-trait = "0.1"
-axum = { version = "0.6", features = ["ws"] }
-base64 = "0.21"
-blake2b_simd = "1.0"
-bytes = "1.4"
-clap = { version = "4.1", features = ["derive", "env"] }
-config = "0.13"
-dirs = "5.0"
-ethers = { version = "2.0", features = ["abigen", "ws"] }
-ethers-core = { version = "2.0" }
-fnv = "1.0"
-futures = "0.3"
-hex = "0.4"
-jsonrpc-v2 = { version = "0.11", default-features = false, features = ["bytes-v10"] }
-k256 = "0.11"                                                                         # Same as tendermint-rs
-lazy_static = "1.4"
-libsecp256k1 = "0.7"
-multihash = { version = "0.16.1", default-features = false }
-num-traits = "0.2"
-paste = "1"
-prost = { version = "0.11" }
-quickcheck = "1"
-quickcheck_macros = "1"
-rand = "0.8"
-rand_chacha = "0.3"
-regex = "1"
-serde = { version = "1", features = ["derive"] }
-serde_json = { version = "1" }
-serde_tuple = "0.5"
-serde_with = "2.3"
-tempfile = "3.3"
-thiserror = "1"
-tokio = { version = "1", features = ["rt-multi-thread", "macros"] }
-tracing = "0.1"
-tracing-subscriber = "0.3"
-
+anyhow = "1" 
+arbitrary = {version = "1", features = ["derive"]} 
+arbtest = "0.2" 
+async-stm = "0.2" 
+async-trait = "0.1" 
+axum = {version = "0.6", features = ["ws"]} 
+base64 = "0.21" 
+blake2b_simd = "1.0" 
+bytes = "1.4" 
+clap = {version = "4.1", features = ["derive", "env"]} 
+config = "0.13" 
+dirs = "5.0" 
+ethers = {version = "2.0", features = ["abigen", "ws"]} 
+ethers-core = {version = "2.0"} 
+fnv = "1.0" 
+futures = "0.3" 
+hex = "0.4" 
+jsonrpc-v2 = {version = "0.11", default-features = false, features = ["bytes-v10"]} 
+k256 = "0.11" # Same as tendermint-rs
+lazy_static = "1.4" 
+libsecp256k1 = "0.7" 
+multihash = {version = "0.16.1", default-features = false} 
+num-traits = "0.2" 
+paste = "1" 
+prost = {version = "0.11"} 
+quickcheck = "1" 
+quickcheck_macros = "1" 
+rand = "0.8" 
+rand_chacha = "0.3" 
+regex = "1" 
+serde = {version = "1", features = ["derive"]} 
+serde_json = {version = "1"} 
+serde_tuple = "0.5" 
+serde_with = "2.3" 
+tempfile = "3.3" 
+thiserror = "1" 
+tokio = {version = "1", features = ["rt-multi-thread", "macros"]} 
+tracing = "0.1" 
+tracing-subscriber = "0.3" 
 
 # Stable FVM dependencies from crates.io
-fvm = { version = "=3.3", default-features = false }     # no opencl feature or it fails on CI
+fvm = {version = "=3.3", default-features = false}# no opencl feature or it fails on CI
 fvm_ipld_blockstore = "0.1"
-fvm_ipld_encoding = "0.3"
 fvm_ipld_car = "0.6"
+fvm_ipld_encoding = "0.3"
 fvm_ipld_hamt = "0.6"
-fvm_shared = { version = "=3.3", features = ["crypto"] }
+fvm_shared = {version = "=3.3", features = ["crypto"]}
 
 # Using 0.8 because of ref-fvm.
 # 0.9 would be better because of its updated quickcheck dependency.
 # 0.10 breaks some API.
-cid = { version = "0.8", features = ["serde-codec", "std"] }
+cid = {version = "0.8", features = ["serde-codec", "std"]}
 
 # Depending on the release cycle, this dependency might want an earlier version of the FVM.
 # We can work around it by hardcoding the method hashes; currently there is only one.
 # frc42_dispatch = "3.2"
 
 # Using the same tendermint-rs dependency as tower-abci. From both we are interested in v037 modules.
-tower-abci = { version = "0.7" }
-tendermint = { version = "0.31", features = ["secp256k1"] }
-tendermint-rpc = { version = "0.31", features = ["secp256k1", "http-client", "websocket-client"] }
-tendermint-proto = { version = "0.31" }
+tendermint = {version = "0.31", features = ["secp256k1"]}
+tendermint-proto = {version = "0.31"}
+tendermint-rpc = {version = "0.31", features = ["secp256k1", "http-client", "websocket-client"]}
+tower-abci = {version = "0.7"}
+
+[patch.crates-io]
+jsonrpc-v2 = {git = "https://github.com/cryptoAtwill/jsonrpc-v2", branch = "expose_inner_error"}
--- a/fendermint/app/src/app.rs
+++ b/fendermint/app/src/app.rs
@@ -539,6 +539,10 @@ where
                 }
             },
         };
+        println!(
+            "-- Deliver Response status: {:?} - info: {:?}",
+            response.code, response.info
+        );
 
         Ok(response)
     }
--- a/fendermint/app/src/app.rs
+++ b/fendermint/app/src/app.rs
@@ -539,10 +539,14 @@ where
                 }
             },
         };
-        println!(
-            "-- Deliver Response status: {:?} - info: {:?}",
-            response.code, response.info
-        );
+
+        if response.code != 0.into() {
+            tracing::error!(
+                "deliver_tx failed: {:?} - {:?}",
+                response.code,
+                response.info
+            );
+        }
 
         Ok(response)
     }
--- a/fendermint/eth/api/src/handlers/http.rs
+++ b/fendermint/eth/api/src/handlers/http.rs
@@ -20,8 +20,7 @@ pub async fn handle(
 
     // NOTE: Any authorization can come here.
 
-    tracing::debug!("RPC request: {request:?}");
-    println!("** RPC request: {request:?}");
+    tracing::info!("RPC request: {request:?}");
 
     let id = request.id_ref().map(id_to_string).unwrap_or_default();
     let method = request.method_ref().to_owned();
@@ -38,13 +37,12 @@ pub async fn handle(
         Ok(result) => {
             tracing::debug!(method, id, result, "RPC call success");
 
-            println!("-- RPC response success!: {method:?} - {result:?}");
+            tracing::debug!("RPC response success: {method:?} - {result:?}");
             (StatusCode::OK, response_headers, result)
         }
         Err(err) => {
             let msg = err.to_string();
             tracing::error!(method, id, msg, "RPC call failure");
-            println!("xx RPC response failure!: {method:?} - {msg:?}");
             (StatusCode::INTERNAL_SERVER_ERROR, response_headers, msg)
         }
     }
--- a/fendermint/eth/api/src/handlers/http.rs
+++ b/fendermint/eth/api/src/handlers/http.rs
@@ -21,6 +21,7 @@ pub async fn handle(
     // NOTE: Any authorization can come here.
 
     tracing::debug!("RPC request: {request:?}");
+    println!("** RPC request: {request:?}");
 
     let id = request.id_ref().map(id_to_string).unwrap_or_default();
     let method = request.method_ref().to_owned();
@@ -36,11 +37,14 @@ pub async fn handle(
     match call_rpc_str(&state.rpc_server, request).await {
         Ok(result) => {
             tracing::debug!(method, id, result, "RPC call success");
+
+            println!("-- RPC response success!: {method:?} - {result:?}");
             (StatusCode::OK, response_headers, result)
         }
         Err(err) => {
             let msg = err.to_string();
             tracing::error!(method, id, msg, "RPC call failure");
+            println!("xx RPC response failure!: {method:?} - {msg:?}");
             (StatusCode::INTERNAL_SERVER_ERROR, response_headers, msg)
         }
     }
@@ -62,3 +66,8 @@ async fn call_rpc_str(
     let response = server.handle(request).await;
     Ok(serde_json::to_string(&response)?)
 }
+
+// RPC request: RequestObject { jsonrpc: V2, method: "eth_estimateGas", params: Some(Value(Array [Object {"data": Null, "from": String("0x6be1ccf648c74800380d0520d797a170c808b624"), "gasPrice": String("0x0"), "value": String("0x0")}])), id: Some(Some(Num(1314076190650))) }
+// RPC request: RequestObject { jsonrpc: V2, method: "eth_blockNumber", params: Some(Value(Array [])), id: Some(Some(Num(4225860438222755))) }
+// RPC request: RequestObject { jsonrpc: V2, method: "eth_getBlockByNumber", params: Some(Value(Array [String("0x3cd"), Bool(false)])), id: Some(Some(Num(4225860438222756))) }
+// RPC request: RequestObject { jsonrpc: V2, method: "eth_feeHistory", params: Some(Value(Array [String("0x5"), String("0x3cd"), Array [Number(10), Number(20), Number(30)]])), id: Some(Some(Num(4225860438222757))) }
--- a/fendermint/vm/interpreter/src/fvm/query.rs
+++ b/fendermint/vm/interpreter/src/fvm/query.rs
@@ -3,7 +3,9 @@
 use async_trait::async_trait;
 use fendermint_vm_message::query::{ActorState, FvmQuery, GasEstimate, StateParams};
 use fvm_ipld_blockstore::Blockstore;
-use fvm_shared::{ActorID, BLOCK_GAS_LIMIT};
+use fvm_shared::{
+    bigint::BigInt, econ::TokenAmount, error::ExitCode, message::Message, ActorID, BLOCK_GAS_LIMIT,
+};
 
 use crate::QueryInterpreter;
 
@@ -64,24 +66,19 @@ where
                 FvmQueryRet::Call(ret)
             }
             FvmQuery::EstimateGas(mut msg) => {
-                // TODO: Figure out how to relate the token balance to a gas limit.
-                //       Do we look at `state.state_params.base_fee`? What about `gas_premium` and `gas_limit`?
-
-                // XXX: This value is for Filecoin, and it's not even used by the FVM, but at least it should not have a problem with it.
+                // Setting BlockGasLimit as initial limit for gas estimation
                 msg.gas_limit = BLOCK_GAS_LIMIT;
 
-                // TODO: This actually fails if the caller doesn't have enough gas.
-                //       Should we modify the state tree up front to give it more?
-                let apply_ret = state.call(*msg)?;
+                // Populate gas message parameters.
+                // If message fails with BLOCK_GAS_LIMIT as the message gas limit
+                // it means that there is an error in the execution of the message
+                // we could optionally propagate that error if needed as done in Lotus.
+                // (but why would you want to estimate the gas of a message that can't
+                // be executed?)
+                let mut msg = self.estimate_gassed_msg(&state, msg)?;
 
-                let est = GasEstimate {
-                    exit_code: apply_ret.msg_receipt.exit_code,
-                    info: apply_ret
-                        .failure_info
-                        .map(|x| x.to_string())
-                        .unwrap_or_default(),
-                    gas_limit: apply_ret.msg_receipt.gas_used,
-                };
+                // perform a gas search for an accurate value
+                let est = self.gas_search(&state, &mut msg)?;
 
                 FvmQueryRet::EstimateGas(est)
             }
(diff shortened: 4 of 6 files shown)


Cargo.toml
```diff
@@ -19,68 +19,69 @@ edition = "2021"
...
-tendermint = { version = "0.31", features = ["secp256k1"] }
-tendermint-rpc = { version = "0.31", features = ["secp256k1", "http-client", "websocket-client"] }
-tendermint-proto = { version = "0.31" }
+tendermint = {version = "0.31", features = ["secp256k1"]}
+tendermint-proto = {version = "0.31"}
+tendermint-rpc = {version = "0.31", features = ["secp256k1", "http-client", "websocket-client"]}
+tower-abci = {version = "0.7"}
+
+[patch.crates-io]
+jsonrpc-v2 = {git = "https://github.com/cryptoAtwill/jsonrpc-v2", branch = "expose_inner_error"}
```
adlrocha, line 87: @cryptoAtwill, @aakoshh do we wait for Will's patch to be accepted upstream, or should we fork this library in the end?

cryptoAtwill: That PR is merged, I think we can switch to upstream.

fendermint/vm/interpreter/src/fvm/query.rs
```diff
@@ -99,3 +96,131 @@ where
...
+        &self,
+        state: &FvmQueryState<DB>,
+        msg: &Message,
+        limit: u64,
+    ) -> anyhow::Result<Option<GasEstimate>> {
+        let mut msg = msg.clone();
+        msg.gas_limit = limit;
+        // set message nonce to zero so the right one is picked up
+        msg.sequence = 0;
+        println!("message being applied: {:?}", msg);
```
cryptoAtwill, line 200: maybe we should use log::info here instead.

cryptoAtwill (approved):

fendermint/vm/interpreter/src/fvm/query.rs
```diff
@@ -99,3 +96,131 @@ where
...
+    // for a valid gas limit for the current message in isolation.
+    fn gas_search(
+        &self,
+        state: &FvmQueryState<DB>,
+        msg: &mut Message,
+    ) -> anyhow::Result<GasEstimate> {
+        let mut curr_limit = msg.gas_limit;
+
+        while {
+            if let Some(ret) = self.estimation_call_with_limit(state, msg, curr_limit)? {
```
cryptoAtwill, line 164: Just checking, the `state` is using `state.call_with_cache(msg, false)`, this call is borrowing state as `mut`, does it update the state? If we run it in a loop, the state would be modified with each loop?

fendermint/vm/interpreter/src/fvm/query.rs
adlrocha: Definitely, I left it behind while I was testing :see_no_evil:

Cargo.toml
aakoshh: I think we can use the fork until that point.

Cargo.toml
```diff
@@ -19,68 +19,69 @@ edition = "2021"
...
-serde_tuple = "0.5"
-serde_with = "2.3"
-tempfile = "3.3"
-thiserror = "1"
-tokio = { version = "1", features = ["rt-multi-thread", "macros"] }
-tracing = "0.1"
-tracing-subscriber = "0.3"
-
+anyhow = "1" 
+arbitrary = {version = "1", features = ["derive"]}
```
aakoshh, line 24: It looks like your tom formatter removes extra spaces. I wonder if we'll have a constant churn on this.

fendermint/app/src/app.rs
```diff
@@ -540,6 +540,14 @@ where
             },
         };
 
+        if response.code != 0.into() {
+            tracing::error!(
```
aakoshh, line 544: I wouldn't necessarily use `error` in relation to user transactions. Who knows why they failed. I worked on a project where the sys admins explained their preference for log levels:
* `error`: They get a call during the middle of the night to go fix it immediately. 
* `warn`: It's not an immediate problem but something they have to deal with before long, e.g. disk space running low. 
* `info`: For your everyday monitoring
* `debug`: Something they can enable for troubleshooting IIRC.

So I would use `info` for this, since for the operator looking at the logs, this maybe interesting but probably nothing they can do anything about. `debug` would be too verbose at the moment and hinder our own development most of the time.

fendermint/eth/api/src/handlers/http.rs
```diff
@@ -20,7 +20,7 @@ pub async fn handle(
 
     // NOTE: Any authorization can come here.
 
-    tracing::debug!("RPC request: {request:?}");
+    tracing::info!("RPC request: {request:?}");
```
aakoshh, line 23: Do we really want to log every request in full glory in everyday scenarios?

fendermint/eth/api/src/handlers/http.rs
```diff
@@ -36,6 +36,8 @@ pub async fn handle(
     match call_rpc_str(&state.rpc_server, request).await {
         Ok(result) => {
             tracing::debug!(method, id, result, "RPC call success");
+
+            tracing::debug!("RPC response success: {method:?} - {result:?}");
```
aakoshh, line 40: Aren't these two logs equivalent? The top version is preferred because it does semantic logging with keywords you can filter on, once we do log aggregation. If you want debug format, you can do one of these:

```rust
tracing::debug!(method, id, ?result, foo = ?bar, "RPC call success");
```

fendermint/vm/interpreter/src/fvm/query.rs
```diff
@@ -99,3 +96,131 @@ where
...
+    const DEFAULT_GAS_PREMIUM: u64 = 20000;
+    /// Gas search step increase used to find the optimal gas limit.
+    const GAS_SEARCH_STEP: f64 = 1.2;
+
+    fn estimate_gassed_msg(
+        &self,
+        state: &FvmQueryState<DB>,
+        msg: Box<Message>,
+    ) -> anyhow::Result<Message> {
+        let mut out = (*msg).clone();
```
aakoshh, line 119: Why not take ownership of the message and return the estimate-filled value? 

I think it's a code smell to take a value by reference only to clone it immediately. You should take it by value and let the caller decide if they want to clone it, or give ownership over theirs to you.

fendermint/vm/interpreter/src/fvm/query.rs
```diff
@@ -99,3 +96,131 @@ where
...
+        // estimate the gas limit and assign it to the message
+        // do not reuse the cache
+        let ret = state.call_with_cache(*msg.clone(), false)?;
+        if ret.msg_receipt.exit_code != ExitCode::OK {
```
aakoshh, line 123:
```suggestion
        if !ret.msg_receipt.exit_code.is_success() {
```

fendermint/vm/interpreter/src/fvm/query.rs
```diff
@@ -99,3 +96,131 @@ where
...
+            ));
+        }
+        out.gas_limit = (ret.msg_receipt.gas_used as f64 * Self::GAS_OVERESTIMATION_RATE) as u64;
+        if out.gas_premium.is_zero() {
+            // TODO: Instead of assigning a default value here, we should analyze historical
+            // blocks from the current height to estimate an accurate value for this premium.
+            // To achieve this we would need to perform a set of ABCI queries.
+            // In the meantime, this value should be good enough to make sure that the
+            // message is included in a block.
+            out.gas_premium = TokenAmount::from_nano(BigInt::from(Self::DEFAULT_GAS_PREMIUM));
```
aakoshh, line 136: ethers-rs [does this](https://github.com/gakonst/ethers-rs/blob/ethers-v2.0.8/ethers-providers/src/rpc/provider.rs#L440), do we really have to do it here? This is forcing rewards onto users via estimates. Shouldn't it be up to them and the miners?

fendermint/vm/interpreter/src/fvm/query.rs
```diff
@@ -99,3 +96,131 @@ where
...
+        }
+        if out.gas_fee_cap.is_zero() {
+            // Compute the fee cap from gas premium and applying an additional overestimation.
+            let overestimated_limit = (out.gas_limit as f64 * Self::GAS_OVERESTIMATION_RATE) as u64;
```
aakoshh, line 140: This is now doubly over estimated. Is that intentional?

fendermint/vm/interpreter/src/fvm/query.rs
```diff
@@ -99,3 +96,131 @@ where
...
+            out.gas_fee_cap = std::cmp::min(
+                TokenAmount::from_atto(BigInt::from(overestimated_limit)) + &out.gas_premium,
```
aakoshh, line 142: Wait the `gas_fee_cap` is a _price`, not a gas amount, isn't it? And why would it contain both the fee _and_ the premium, when there is the separate premium field on its own?

Look [here](https://github.com/filecoin-project/ref-fvm/blob/fvm%40v3.5.0/fvm/src/executor/default.rs#L453): 
```rust
let gas_cost: TokenAmount = msg.gas_fee_cap.clone() * msg.gas_limit;
```

What you are doing would result in the `gas_limit` being multiplied by itself.

aakoshh: I thought `gas_premium` was also some kind of price as it gets added together with the fee cap [here](https://github.com/filecoin-project/ref-fvm/blob/fvm%40v3.5.0/fvm/src/gas/outputs.rs#L40).

fendermint/vm/interpreter/src/fvm/query.rs
```diff
@@ -99,3 +96,131 @@ where
...
+
+            curr_limit = (curr_limit as f64 * Self::GAS_SEARCH_STEP) as u64;
+            if curr_limit > BLOCK_GAS_LIMIT {
+                return Ok(GasEstimate {
+                    exit_code: ExitCode::OK,
+                    info: "".to_string(),
+                    gas_limit: BLOCK_GAS_LIMIT,
+                });
+            }
+            true
```
aakoshh, line 176: What is this `true` for?

aakoshh: Oh, I see it's like:

```rust
while {
   <stuff>
   true
} {} // empty block
```

I reckon this would be cleaner:

```rust 
loop {
  <stuff>
}
```

fendermint/vm/interpreter/src/fvm/query.rs
```diff
@@ -99,3 +96,131 @@ where
...
+
+        // TODO: For a more accurate gas estimation we could track the low and the high
+        // of the search and make higher steps (e.g. `GAS_SEARCH_STEP = 2`).
+        // Once an interval is found of [low, high] for which the message
+        // succeeds, we make a finer-grained within that interval.
+        // At this point, I don't think is worth being that accurate as long as it works.
+
+        Err(anyhow::anyhow!(
+            "gas search failed. no valid gas limit found"
+        ))
```
aakoshh, line 187: This will never be called because you never break out of the loop, only return.

fendermint/vm/interpreter/src/fvm/query.rs
```diff
@@ -99,3 +96,131 @@ where
...
+    }
+
+    fn estimation_call_with_limit(
+        &self,
+        state: &FvmQueryState<DB>,
+        msg: &Message,
+        limit: u64,
+    ) -> anyhow::Result<Option<GasEstimate>> {
+        let mut msg = msg.clone();
```
aakoshh, line 196:
```suggestion
        mut msg: Message,
        limit: u64,
    ) -> anyhow::Result<Option<GasEstimate>> {
```
Same as before, let the caller decide how they want to satisfy your need for owning a message.

fendermint/vm/interpreter/src/fvm/query.rs
aakoshh: I reckon it can be removed.

fendermint/vm/interpreter/src/fvm/query.rs
```diff
@@ -64,24 +66,19 @@ where
...
-                let est = GasEstimate {
-                    exit_code: apply_ret.msg_receipt.exit_code,
-                    info: apply_ret
-                        .failure_info
-                        .map(|x| x.to_string())
-                        .unwrap_or_default(),
-                    gas_limit: apply_ret.msg_receipt.gas_used,
-                };
+                // perform a gas search for an accurate value
+                let est = self.gas_search(&state, &mut msg)?;
```
aakoshh, line 81: I don't think using `?` here is the right call, because further down you return an error if the call fails for any reason other than OUT_OF_GAS, which would kill the proces. 

This is why previously I returned the `ApplyRet` from the `call` and used that to return a `GasEstimate` with both the gas used _and_ the fault code.

fendermint/vm/interpreter/src/fvm/query.rs
```diff
@@ -99,3 +96,131 @@ where
...
+        if ret.exit_code == ExitCode::OK {
+            return Ok(Some(ret));
+        }
+
+        if ret.exit_code != ExitCode::SYS_OUT_OF_GAS {
+            return Err(anyhow::anyhow!(
+                "message execution failed in gas search with error code {:?}",
+                ret.exit_code
+            ));
+        }
```
aakoshh, line 221: So this is what we have to be very careful with, the way you use it will kill the process. 

If I were you I would add an `is_out_of_gas` method to the `GasEstimate` struct and decide on that in `gas_search`. It would simplify the decision logic 
* from `Ok(Some) | Ok(None) | Err` 
* to `Ok(est) | Err` and looking at `est.is_out_of_gas()` which is a bit more readable as well on the caller side.

fendermint/vm/interpreter/src/fvm/state/query.rs
```diff
@@ -143,14 +149,25 @@ where
     /// The results are never going to be flushed, so it's semantically read-only,
     /// but it might write into the buffered block store the FVM creates. Running
     /// multiple such messages results in their buffered effects stacking up.
-    pub fn call(&self, mut msg: FvmMessage) -> anyhow::Result<ApplyRet> {
+    pub fn call(&self, msg: FvmMessage) -> anyhow::Result<ApplyRet> {
+        // the default behavior of call is to reuse the cache and stack the results.
+        // for isolated execution of message without cache, use directly `call_with_cache`
+        self.call_with_cache(msg, true)
```
aakoshh, line 155: So currently if one mixes `true` and `false` calls, then whether effects stack up depends on the order of execution, e.g. `call; call_with_cache(false); call` would see the cache reverted for the last call, while `call; call;` would all stack up. 

I would perhaps use a simpler semantic where the `reuse_cache: false` option would skip overwriting whatever was in the cache and leave it untouched, just working with a fresh engine, so the `true` and `false` values are completely isolated. I might be wrong on this, but it's always possible to go to an enum if one wants to specify the exact behaviour. 

I just find it odd that we have `call` which hides this from us. Maybe a `call` and a `call_nocache` with no flag could make it clear that one is stacking, the other isn't?

fendermint/vm/interpreter/src/fvm/query.rs
```diff
@@ -99,3 +96,130 @@ where
...
+{
+    /// Overestimation rate applied to gas to ensure that the
+    /// message goes through in the gas estimation.
+    const GAS_OVERESTIMATION_RATE: f64 = 1.25;
+    /// Default gas premium value. Inferred through a quick search through
+    /// InvokeEVM messages in filfox. The default value is only used if
+    /// the user hasn't specified a gas premium.
+    const DEFAULT_GAS_PREMIUM: u64 = 20000;
+    /// Gas search step increase used to find the optimal gas limit.
+    const GAS_SEARCH_STEP: f64 = 1.2;
```
aakoshh, line 112: These should be added as constructor arguments to the `FvmMessageInterpreter`.

aakoshh (commented):
Great stuff, finally we can work out what these damn gas fields are for! 

I still don't fully understand why we need this. 

If we set `gas_fee_cap` to 0, then the FVM does not do a balance check (miner pays the penalty) and one can execute a message to its fullest, and just take the `gas_used` as the limit. Maybe do 10% over estimation. 

You said that the estimated limits weren't enough for you, so I assume you must have set the `gas_fee_cap` to some non-zero value, in which case you can run out of balance as well as the gas limit (but that we set to the BLOCK_GAS_LIMIT during estimation). 

Couldn't we just set the fees to 0 and return the gas, and let the client worry about coming up with a fee they are willing to pay?

fendermint/vm/interpreter/src/fvm/query.rs
```diff
@@ -99,3 +96,130 @@ where
...
+    ) -> anyhow::Result<Message> {
+        let mut out = (*msg).clone();
+        // estimate the gas limit and assign it to the message
+        // do not reuse the cache
+        let ret = state.call_with_cache(*msg.clone(), false)?;
+        if ret.msg_receipt.exit_code != ExitCode::OK {
+            return Err(anyhow::anyhow!(
+                "message execution failed with error code {}",
+                ret.msg_receipt.exit_code
+            ));
```
aakoshh, line 127: No, this also cannot be treated as an error and then [used with ? operator](https://github.com/consensus-shipyard/fendermint/pull/163/files#diff-5be56262407092392424776bca11de06a243db9770516d04402d1f7d6961ab6bR78). You have to return it as `Ok` and inspect the exit code, or you have to do a case analysis on the result, otherwise the process crashes.

fendermint/vm/interpreter/src/fvm/query.rs
```diff
@@ -99,3 +96,130 @@ where
...
+        out.gas_limit = (ret.msg_receipt.gas_used as f64 * Self::GAS_OVERESTIMATION_RATE) as u64;
+        if out.gas_premium.is_zero() {
+            // TODO: Instead of assigning a default value here, we should analyze historical
+            // blocks from the current height to estimate an accurate value for this premium.
+            // To achieve this we would need to perform a set of ABCI queries.
+            // In the meantime, this value should be good enough to make sure that the
+            // message is included in a block.
+            out.gas_premium = TokenAmount::from_nano(BigInt::from(Self::DEFAULT_GAS_PREMIUM));
+        }
+        if out.gas_fee_cap.is_zero() {
```
aakoshh, line 138: I don't fully understand the semantics of this method. 

If the `gas_fee_cap` was non-zero, the `call` could have failed with OUT_OF_GAS, in which case we don't get here. 

If the `gas_fee_cap` was zero, then it will most likely not have an OUT_OF_GAS error because earlier we set the gas limit to the whole BLOCK_GAS_LIMIT, however if we then set the `gas_fee_cap` then it's terra incognita again. 

What can we claim about the output here?

fendermint/vm/interpreter/src/fvm/query.rs
```diff
@@ -99,3 +96,130 @@ where
...
+    fn estimation_call_with_limit(
+        &self,
+        state: &FvmQueryState<DB>,
+        msg: &Message,
+        limit: u64,
+    ) -> anyhow::Result<Option<GasEstimate>> {
+        let mut msg = msg.clone();
+        msg.gas_limit = limit;
+        // set message nonce to zero so the right one is picked up
+        msg.sequence = 0;
```
aakoshh, line 199: Shouldn't the `estimate_gassed_msg` also have this?

fendermint/vm/interpreter/src/fvm/query.rs
```diff
@@ -64,24 +66,19 @@ where
                 FvmQueryRet::Call(ret)
             }
             FvmQuery::EstimateGas(mut msg) => {
-                // TODO: Figure out how to relate the token balance to a gas limit.
-                //       Do we look at `state.state_params.base_fee`? What about `gas_premium` and `gas_limit`?
-
-                // XXX: This value is for Filecoin, and it's not even used by the FVM, but at least it should not have a problem with it.
+                // Setting BlockGasLimit as initial limit for gas estimation
                 msg.gas_limit = BLOCK_GAS_LIMIT;
```
aakoshh, line 70: Move into `estimate_gassed_message`?

fendermint/vm/interpreter/src/fvm/query.rs
```diff
@@ -64,24 +66,19 @@ where
...
-                // TODO: This actually fails if the caller doesn't have enough gas.
-                //       Should we modify the state tree up front to give it more?
-                let apply_ret = state.call(*msg)?;
+                // Populate gas message parameters.
+                // If message fails with BLOCK_GAS_LIMIT as the message gas limit
+                // it means that there is an error in the execution of the message
+                // we could optionally propagate that error if needed as done in Lotus.
+                // (but why would you want to estimate the gas of a message that can't
+                // be executed?)
+                let mut msg = self.estimate_gassed_msg(&state, msg)?;
```
aakoshh, line 78: I don't understand: this method will put in the exact gas used with a 25% over estimation in the message. Why should we have to perform a `gas_search` after this, which further increases it by 20% in each iteration? Why would it suddenly fail to execute on the same exact state, when gas is a deterministic value?

aakoshh (changes requested):
Just a formal request for changing the error handling to avoid killing the process in case a transaction returns an error from its execution.

Cargo.toml
adlrocha: Cool, I bumped the version in this PR to pick up the latest one.

fendermint/vm/interpreter/src/fvm/query.rs
adlrocha: Correct. If `cache = true` the state is updated in each iteration and changes to the state are buffered, but these changes are not committed in the blockstore (they only apply to this `FvmQueryState` instance.

Cargo.toml
adlrocha: I disabled the toml formatter to avoid this from happening again. Sorry about that.

fendermint/app/src/app.rs
adlrocha: I like that categorization :) Fine by me. I was thinking about using `error` or  `warn` to make errors standout the normal operation logs. But as long as we have a common  criteria I am fine :) Fixed.

fendermint/eth/api/src/handlers/http.rs
adlrocha: Totally, this is just me being sloppy. I forgot to remove that one (which initially was a `println!`). Thanks!

fendermint/eth/api/src/handlers/http.rs
adlrocha: In other RPC APIs from my past life it was useful to have all of the requests explicitly logged even if it was noisy, but that might just be my personal preference. I'll revert and once we deploy in production we can figure out the right balance.

fendermint/vm/interpreter/src/fvm/query.rs
adlrocha: That makes total sense. Thank you for the feedback. I think this was a consequence of me not knowing Rust well. I thought I couldn't not use `Box` as raw reference. Will be fixed in the same commit.

fendermint/vm/interpreter/src/fvm/query.rs
adlrocha: In Lotus they have a local estimation of fees also in case the user didn't set the `gas_premium`. The reason for that is that you can use Lotus to send non-EVM messages directly and the user may not set the premium. I guess having a constant premium here as a sanity-check should be OK for us because we will rely on `ethers` and other tooling to compute this for us, so we should expect almost all messages to have `gas_premium != 0`. 

Weak opinion weakly held. I can also bypass `gas_premium` here even if it is zero if you think this is the correct way of doing it.

fendermint/vm/interpreter/src/fvm/query.rs
adlrocha: I think you are right, I completely misunderstood what they are doing in Lotus. I will review this part. Thanks!

fendermint/vm/interpreter/src/fvm/query.rs
adlrocha: Fixed.

fendermint/vm/interpreter/src/fvm/query.rs
adlrocha: Fixed.

fendermint/vm/interpreter/src/fvm/query.rs
aakoshh: `Box` is basically just a thing you own that points at something on the heap, so that you can have a pointer instead of including the bytes in your data structures. I think you can pass `Box<T>` as a raw reference `&T` with `&` or `.as_ref()`, and also take ownership of its contents by moving them onto the stack, which consumes the `Box`.

Cargo.toml
aakoshh: Cheers! I also use a TOML formatter but I couldnt' find a whitespace setting for this one, only arrays and such.

fendermint/vm/interpreter/src/fvm/query.rs
aakoshh: I don't know, I didn't think there was a correct way because the FVM happily executes with 0 fees and 0 premiums, and it should be up to the validator to make sure they only include transactions which are profitable, see #29 

Personally I would not set anything that isn't necessary to be set, especially since this value is not returned from the API, so even now I'm not sure what effect we have by setting it.

fendermint/vm/interpreter/src/fvm/query.rs
aakoshh: Hm, I might be wrong, but we have this on line 129:

```rust
        out.gas_limit = (ret.msg_receipt.gas_used as f64 * Self::GAS_OVERESTIMATION_RATE) as u64;
```
and then on line 140:
```rust
let overestimated_limit = (out.gas_limit as f64 * Self::GAS_OVERESTIMATION_RATE) as u64;
```

Which gives:
```
 out.gas_fee_cap = out.gas_limit * GAS_OVERESTIMATION_RATE * GAS_OVERESTIMATION_RATE  + &out.gas_premium,
```

That's `GAS_OVERESTIMATION_RATE` squared.

fendermint/vm/interpreter/src/fvm/query.rs
aakoshh: Glad it's not just me! :smiley:

fendermint/vm/interpreter/src/fvm/query.rs
adlrocha: Now I get how it works :) Fixed in the next commit. Thanks!

fendermint/vm/interpreter/src/fvm/query.rs
adlrocha: Fixed.

fendermint/vm/interpreter/src/fvm/query.rs
adlrocha: After reading the spec and tinkering with it, I think you are right, is better to just estimate the `gas_limit` for now, and then compute the `gas_premium` and `gas_fee_cap` optimally once we figure out the `base_fee` story. Unless you are against, let's defer for the future.

fendermint/vm/interpreter/src/fvm/query.rs
adlrocha: This was supposed to give additional fee cap space in case the estimation was too conservative. We are not setting a fee cap anymore so this comment doesn't apply anymore.

fendermint/vm/interpreter/src/fvm/query.rs
adlrocha: So the idea here is that if the overestimation fails (i.e. we haven´t set enough gas yet), we may need to search for a bigger value.

I reduced the overestimation and steps to 5% now, and now that I removed the `gas_fee_cap` it may always get it at the first attempt, but in case this changes in the future I think we should keep the `gas_search`.

fendermint/app/src/cmd/run.rs
```diff
@@ -20,7 +22,8 @@ cmd! {
 }
 
 async fn run(settings: Settings) -> anyhow::Result<()> {
-    let interpreter = FvmMessageInterpreter::<NamespaceBlockstore>::new();
+    let interpreter =
+        FvmMessageInterpreter::<NamespaceBlockstore>::new(DEFAULT_GAS_RATE, DEFAULT_GAS_RATE);
```
aakoshh, line 26: You can add them to `Settings` if it's already out here.

fendermint/vm/interpreter/src/fvm/query.rs
```diff
@@ -99,3 +96,112 @@ where
...
+            // the case but after a lot of testing, setting this value to zero rejects the transaction)
+            msg.gas_premium = TokenAmount::from_nano(BigInt::from(1));
+        }
+        if msg.gas_fee_cap.is_zero() {
+            // Compute the fee cap from gas premium and applying an additional overestimation.
+            let overestimated_limit = (msg.gas_limit as f64 * self.gas_overestimation_rate) as u64;
+            msg.gas_fee_cap = std::cmp::min(
+                TokenAmount::from_atto(BigInt::from(overestimated_limit)) + &msg.gas_premium,
+                TokenAmount::from_atto(BLOCK_GAS_LIMIT),
+            );
```
aakoshh, line 138: This looks wrong again:
1. It's not a price but a gas amount (potentially equals the block gas limit!)
2. It's `gas_limit * gas_overestimation_rate ** 2`

fendermint/vm/interpreter/src/fvm/state/query.rs
aakoshh: We agreed to change this to have a single `call` method and add a `use_cache: bool` parameter to it. If `use_cache` is false it will not touch the cache, including not resetting it. This way it's always clear from the call site how it's going to work.

fendermint/app/src/settings.rs
```diff
@@ -36,6 +36,16 @@ pub struct DbSettings {
...
 
+#[derive(Debug, Deserialize)]
+pub struct FvmSettings {
+    /// Overestimation rate applied to gas estimations to ensure that the
+    /// message goes through
+    pub gas_overestimation_rate: f64,
+    /// Gas search step increase used to find the optimal gas limit.
+    /// It determines how fine-grained we want the gas estimation to be.
+    pub gas_search_step: f64,
+}
```
aakoshh, line 47: You'll need to add default values to https://github.com/consensus-shipyard/fendermint/blob/b1493d05b46891c72f43653d747b2ace790c0958/fendermint/app/config/default.toml

fendermint/vm/interpreter/src/fvm/query.rs
```diff
@@ -99,3 +96,111 @@ where
...
+
+        // estimate the gas limit and assign it to the message
+        // do not reuse the cache
+        let ret = state.call(msg.clone(), false)?;
+        if !ret.msg_receipt.exit_code.is_success() {
+            // if the message fail we can't estimate the gas.
+            return Ok(Some(GasEstimate {
+                exit_code: ret.msg_receipt.exit_code,
+                info: ret.failure_info.map(|x| x.to_string()).unwrap_or_default(),
+                gas_limit: 0,
```
aakoshh, line 120: Should this be `gas_used`? In the [bottom](https://github.com/consensus-shipyard/fendermint/pull/163/files#diff-5be56262407092392424776bca11de06a243db9770516d04402d1f7d6961ab6bR194) it is, even in the error case.

fendermint/app/src/settings.rs
adlrocha: Hmm... I did. I don't know why they haven't been pushed. Let me check

fendermint/vm/interpreter/src/fvm/state/query.rs
```diff
@@ -142,15 +146,17 @@ where
...
+    /// multiple such messages results in their buffered effects stacking up if
+    /// `use_cache` is enabled. If `use_cache` is not enabled, the results of the
+    /// execution are not cached.
+    pub fn call(&self, mut msg: FvmMessage, use_cache: bool) -> anyhow::Result<ApplyRet> {
         // If the sequence is zero, treat it as a signal to use whatever is in the state.
         if msg.sequence.is_zero() {
             if let Some((_, state)) = self.actor_state(&msg.from)? {
                 msg.sequence = state.sequence;
             }
         }
```
aakoshh, line 158: For correctness this should be inside the `with_exec_state`, otherwise the state root can be different once `use_cache: true` has been used to execute some prior transactions. It's my bug and we can fix it in another PR.

fendermint/vm/interpreter/src/fvm/state/query.rs
```diff
@@ -109,7 +111,9 @@ where
         .context("error creating execution state")?;
 
         let res = f(&mut exec_state);
-        *cache = Some(exec_state);
+        if use_cache {
```
aakoshh, line 114: This should also be used when deciding whether to [use the cache](https://github.com/consensus-shipyard/fendermint/blob/b1493d05b46891c72f43653d747b2ace790c0958/fendermint/vm/interpreter/src/fvm/state/query.rs#L101) or not.

fendermint/vm/interpreter/src/fvm/state/query.rs
aakoshh: https://github.com/consensus-shipyard/fendermint/issues/176

fendermint/vm/interpreter/src/fvm/query.rs
adlrocha: I used this to make explicit that the message errored and we shouldn't use the gas estimation, but no strong opinion, we can return the `gas_used` even if it failed.

fendermint/vm/interpreter/src/fvm/state/query.rs
adlrocha: You are right! Up to you I can fix it here if needed (although it may be cleaner to just start a new PR and get this over with :) )

aakoshh (approved):
:tada

PR #166, opened 2023-07-31
FM-162: Genesis config for IPC

### LINKED ISSUE #162: IPC: Add IPC configuration to Genesis
https://github.com/consensus-shipyard/fendermint/pull/159 includes a [TODO](https://github.com/consensus-shipyard/fendermint/blob/a64196b1c953e88a3bd4fb08f6536e7f068d7d6f/fendermint/vm/interpreter/src/fvm/genesis.rs#L246) to move the hardcoded dummy IPC parameters into the Genesis file, which includes things like the parent network ID.

aakoshh:
Closes #162 

Builds on #148 

Adds a `fendermint genesis ipc gatway` command to set the options related to the creation of the `Gateway` actor in the `genesis.json` file.


Cargo.toml
```diff
@@ -97,3 +103,6 @@ tower-abci = { version = "0.7" }
 tendermint = { version = "0.31", features = ["secp256k1"] }
 tendermint-rpc = { version = "0.31", features = ["secp256k1", "http-client", "websocket-client"] }
 tendermint-proto = { version = "0.31" }
+
+# Using the IPC SDK without the `fil-actor` feature so as not to depend on the actor `Runtime`.
+ipc-sdk = { git = "https://github.com/consensus-shipyard/ipc-actors.git", default-features = false, branch = "fm-162-fil-actor-feature" }
```
adlrocha, line 108: Should we release a tag or point to `main` before merging?
```suggestion
ipc-sdk = { git = "https://github.com/consensus-shipyard/ipc-actors.git", default-features = false, branch = "main" }
```

fendermint/vm/actor_interface/src/ipc.rs
```diff
@@ -28,6 +31,37 @@ pub mod gateway {
...
+                route.push(H160::from(addr.0))
+            }
+            Ok(Self {
+                network_name: SubnetID {
+                    root: value.subnet_id.root_id(),
+                    route,
+                },
+                bottom_up_check_period: value.bottom_up_check_period,
+                top_down_check_period: value.top_down_check_period,
+                // XXX: Ignoring any error resulting from larger fee than what fits into U256. This is in genesis after all.
```
adlrocha, line 58: Actually, should we panic here if this happens? As it is in genesis is not part of the core operation and it would allow us to catch potential misconfigurations.

fendermint/vm/genesis/src/lib.rs
```diff
@@ -16,6 +16,26 @@ use fendermint_vm_encoding::IsHumanReadable;
...
 
+/// The genesis data structure we serialize to JSON and start the chain with.
+#[serde_as]
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
+pub struct Genesis {
+    /// The name of the blockchain.
+    ///
+    /// It will be used to derive a chain ID as well as being
+    /// the network name in the `InitActor`.
+    pub chain_name: String,
```
adlrocha, line 27: This `String` is actually expected to be the stringinfied representation of a `SubnetID` right?

adlrocha (approved):
A few minor comments but LGTM. Thanks!

fendermint/vm/actor_interface/src/ipc.rs
aakoshh: Thanks, added a test to show that it indeed panics.

PR #167, opened 2023-08-01
Export snapshot

cryptoAtwill:
Initial attempt to add snapshot functionalities to fendermint. The ideas are as follows:
1. We are only taking the latest `StateParams`, `BlockHeight` and `StateTree`. (Maybe we can consider snapshot some past StateParams?)
2. Snapshot are directly written to CAR file by a generic IPLD walker that loops over all the cids and their values in the underlyiing blockstore.
3. Taking snapshot of the latest states can be triggered from cron job at the start/end of some blocks as a async process running in the background.
4. To interface with tendermint `list_snapshots` call, we can return the path of the snapshot files in metadata field and chunk the car file accordingly.

This PR is starting to address point 1 and 2.


fendermint/vm/interpreter/src/fvm/state/snapshot.rs
```diff
@@ -0,0 +1,312 @@
...
+
+        Ok(Self {
+            state_tree_root,
+            state_tree,
+            block_state_params_bytes,
+            block_state_params_cid,
+        })
+    }
+
+    /// For V1 snapshot, we are putting two components into the CAR file: the state tree and latest
```
adlrocha, line 117: I don't know if it makes sense in Fendermint (or if we even need them) because maybe Tendermint already includes a way to recover from a snapshot or genesis, but in Lotus snapshots include also all transactions from genesis (or a previous snapshot) so one could verify if the state tree is correct by re-executing all messages.

We should maybe consider adding the option to include messages in a snapshot?

fendermint/vm/interpreter/src/fvm/state/snapshot.rs
```diff
@@ -0,0 +1,312 @@
...
+    pub fn new(
+        store: DB,
+        state_params: FvmStateParams,
+        block_height: BlockHeight,
+    ) -> anyhow::Result<Self> {
+        let state_tree =
+            StateTree::new_from_root(ReadOnlyBlockstore::new(store), &state_params.state_root)?;
+        let state_tree_root = state_params.state_root;
+
+        let block_state_params_bytes = fvm_ipld_encoding::to_vec(&(state_params, block_height))?;
```
adlrocha, line 105: Here you are serializing a tuple of `state_params`, `height`, right? Wouldn't it be better to create a clear data structure that include both so its de/serialization is clearer?

fendermint/vm/interpreter/src/fvm/state/snapshot.rs
```diff
@@ -0,0 +1,312 @@
...
+
+            car.write_stream_async(&mut Pin::new(&mut write), &mut streamer)
+                .await
+                .unwrap();
+
+            let mut streamer = tokio_stream::iter(vec![(
+                self.block_state_params_cid,
+                self.block_state_params_bytes,
+            )]);
+            car.write_stream_async(&mut Pin::new(&mut write), &mut streamer)
```
adlrocha, line 142: According to the docs, `write_strea_async` writes the header and stream of data to writer. Would calling it twice write the header twice? We should double-check this to ensure that we are not writing the header twice.

If you look at the CAR spec, there is only one header with the roots at the beginning of the file: https://ipld.io/specs/transport/car/carv2/

fendermint/vm/interpreter/src/fvm/state/snapshot.rs
```diff
@@ -0,0 +1,312 @@
...
+
+impl<BlockStore> StateTreeStreamer<BlockStore> {
+    pub fn new(state_root_cid: Cid, bs: BlockStore) -> Self {
+        let mut dfs = VecDeque::new();
+        dfs.push_back(state_root_cid);
+        Self { dfs, bs }
+    }
+}
+
+impl<BlockStore: Blockstore> Stream for StateTreeStreamer<BlockStore> {
```
adlrocha, line 170: Out of a curiosity, has this implementation of the streamer been extracted from Forest? If this is the case I would add a note pointing to the original implementation (if possible).

adlrocha (commented):
A few comments, but I think we are in a good path, let me know if you need me to look deeper into anything. Thanks!

fendermint/vm/interpreter/Cargo.toml
```diff
@@ -28,11 +28,19 @@ fvm_ipld_blockstore = { workspace = true }
...
+futures-util = "0.3.28"
+libipld = { version = "0.14", default-features = false, features = ["dag-cbor"] }
 tokio = { workspace = true }
+tokio-stream = "0.1.14"
+tokio-util = {version = "0.7.8", features = ["compat"]}
+pin-project = "1.1.2"
 
+[dev-dependencies]
+quickcheck = { workspace = true }
+fvm = { version = "~3.3", default-features = false, features=["arb", "testing"] }
```
aakoshh, line 41: There definitely is an `fvm` already in the workspace. Far as I know every different feature combination results in a different compilation, so it's better to use the same. I understand this requires the `testing` feature that isn't needed in the normal context, but look, you can take the version from the workspace and add a feature here:

https://github.com/consensus-shipyard/fendermint/blob/master/fendermint/eth/api/Cargo.toml#L38

More generally, I think we should maintain all versions in the root and avoid mixing. I have a fear like it will end up being a mess in the future.

fendermint/vm/interpreter/src/fvm/state/snapshot.rs
aakoshh: We wouldn't have access to all those messages unless we store them or query them from Tendermint, but I don't think this is the intended use of snapshots in Tendermint, since it can garbage collect old blocks.

fendermint/vm/interpreter/src/fvm/state/snapshot.rs
```diff
@@ -0,0 +1,357 @@
...
+
+pub struct V1Snapshot<DB> {
+    /// The root cid of the state tree
+    state_tree_root: Cid,
+    /// The state tree of the current blockchain
+    state_tree: StateTree<ReadOnlyBlockstore<DB>>,
+    /// The latest block height with state params serialized to bytes
+    block_state_params_bytes: Vec<u8>,
+    /// Block state params cid derived from block state params bytes
+    block_state_params_cid: Cid,
```
aakoshh, line 92: There is the `fvm_ipld_blockstore::Block` type which can wrap bytes and compute the CID on demand, and then it's clear what codec it is using.

But I'm not sure why you wouldn't store just the `FvmStateParams` in this struct and only serialize to bytes when writing to a CAR, and also deserialize when reading a CAR.

fendermint/vm/interpreter/src/fvm/state/snapshot.rs
```diff
@@ -0,0 +1,357 @@
...
+            )]);
+            let mut streamer = state_tree_streamer.merge(block_state_streamer);
+
+            let mut write = write.compat_write();
+            car.write_stream_async(&mut Pin::new(&mut write), &mut streamer)
+                .await
+                .unwrap();
+        });
+
+        write_task.await?;
```
aakoshh, line 175: Will a panic in the task be turned into an `Error`?

fendermint/vm/interpreter/src/fvm/state/snapshot.rs
```diff
@@ -0,0 +1,357 @@
...
+    /// serialized block state params.
+    fn car_header_roots(&self) -> anyhow::Result<Vec<Cid>> {
+        Ok(vec![self.state_tree_root, self.block_state_params_cid])
+    }
+
+    pub fn init_chain(&self) -> anyhow::Result<()> {
+        todo!()
+    }
+
+    async fn write_car(self, car: CarHeader, write: tokio::fs::File) -> anyhow::Result<()> {
```
aakoshh, line 159: Why create the `CarHeader` here, and then the `car_header_roots` function is not needed, you can delay serializing the `state_params` to a `Block` to get the CID up to this point, and not need to store bytes and CID in the struct?

fendermint/vm/interpreter/src/fvm/state/snapshot.rs
```diff
@@ -0,0 +1,357 @@
...
+        });
+
+        write_task.await?;
+
+        Ok(())
+    }
+}
+
+#[pin_project::pin_project]
+struct StateTreeStreamer<BlockStore> {
```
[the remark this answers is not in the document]
aakoshh: It would be more intuitive when you look at `bs: BlockStore` that it's not a regular type.

fendermint/vm/interpreter/src/fvm/state/snapshot.rs
```diff
@@ -0,0 +1,357 @@
...
+    use std::collections::VecDeque;
+
+    fn prepare_state_tree(items: u64) -> (Cid, StateTree<MemoryBlockstore>) {
+        let store = MemoryBlockstore::new();
+        let mut state_tree = StateTree::new(store, StateTreeVersion::V5).unwrap();
+        let mut gen = Gen::new(16);
+
+        for i in 1..=items {
+            let state = ActorState::arbitrary(&mut gen);
+            state_tree.set_actor(i, state);
```
aakoshh, line 271: So the state will have a CID not pointing at anything, right?

fendermint/vm/interpreter/src/fvm/state/snapshot.rs
```diff
@@ -0,0 +1,357 @@
...
+    }
+
+    #[tokio::test]
+    async fn test_streamer() {
+        let (root_cid, state_tree) = prepare_state_tree(100);
+        let bs = state_tree.into_store();
+        let mut stream = StateTreeStreamer {
+            dfs: VecDeque::from(vec![root_cid.clone()]),
+            bs: bs.clone(),
+        };
```
aakoshh, line 305: Use `StateTreeStreamer::new`?

fendermint/vm/interpreter/src/fvm/state/snapshot.rs
```diff
@@ -0,0 +1,357 @@
...
+            Snapshot::V1(s) => s,
+        };
+
+        assert_eq!(state_root, loaded_snapshot.state_tree_root);
+        assert_tree2_contains_tree1(
+            &StateTree::new_from_root(bs, &state_root).unwrap(),
+            &loaded_snapshot.state_tree,
+        );
+
+        let state_params_bytes = fvm_ipld_encoding::to_vec(&(state_params, block_height)).unwrap();
```
aakoshh, line 354: I think this should live inside `read_car` so the caller doesn't need to understand that you were using this tuple. Oh you aren't even deserializing here, so even more so.

aakoshh (commented):
Very nice! 

There are some crates for me to catch up on, I have no idea about `pin_project` :) 

There shouldn't be any changes to the `ipc_actors`, unless the `dev` branched changed there. But definitely `gateway.rs` should not exist any more if you rebase.

fendermint/vm/interpreter/src/fvm/state/snapshot.rs
```diff
@@ -0,0 +1,357 @@
...
+    DB: Blockstore + Clone + 'static + Send,
+{
+    /// Creates a new V2Snapshot struct. Caller ensure store
+    pub fn new(
+        store: DB,
+        state_params: FvmStateParams,
+        block_height: BlockHeight,
+    ) -> anyhow::Result<Self> {
+        let state_tree =
+            StateTree::new_from_root(ReadOnlyBlockstore::new(store), &state_params.state_root)?;
```
aakoshh, line 106: Because `FvmStateParams` have a `state_root` inside it, and it's a `Cid`, isn't it enough to use the CID of the `(state_params, block_height)` structure as a single root CID for the CAR file? The state root is implicitly reachable through that.

fendermint/vm/interpreter/src/fvm/state/snapshot.rs
adlrocha: Amazing, then. If Tendermint handles this for us then no need to include messages in snapshots. So here what we are building is actually a "BFT-snapshot", i.e. the baseline state required for a node to sync from a height.

fendermint/vm/interpreter/src/fvm/state/snapshot.rs
cryptoAtwill: yeah, it's only written once now

fendermint/vm/interpreter/src/fvm/state/snapshot.rs
cryptoAtwill: It's not strictly a reference, but the inspiration is from https://github.com/ChainSafe/forest/blob/77d6b2b128d73900b0162e3f573ff8d63e6324b3/src/ipld/util.rs#L335

fendermint/vm/interpreter/Cargo.toml
cryptoAtwill: updated

fendermint/vm/interpreter/src/fvm/state/snapshot.rs
```diff
@@ -0,0 +1,357 @@
...
+        let snapshot = V1Snapshot::read_car(file, store).await?;
+        Ok(Self::V1(snapshot))
+    }
+
+    /// Write the snapshot to car file
+    pub async fn write_car(self, path: impl AsRef<Path>) -> anyhow::Result<()> {
+        let file = tokio::fs::File::create(path).await?;
+
+        // derive the car header roots
+        let car = CarHeader::new(self.car_header_roots()?, self.version());
```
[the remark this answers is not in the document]
cryptoAtwill: you are right, this seems to be the version of the car file after reading the spec. I changed it already.

fendermint/vm/interpreter/src/fvm/state/snapshot.rs
cryptoAtwill: right, for test only. The cid will be picked up by the ipld walker and found there is no data there and continue. We can panic if this happens in non-test env.

fendermint/vm/interpreter/src/fvm/state/snapshot.rs
cryptoAtwill: That works too

fendermint/vm/interpreter/src/fvm/state/snapshot.rs
cryptoAtwill: It's just two fields so I feel making a type alias is easier and clear.

fendermint/vm/interpreter/src/fvm/state/snapshot.rs
cryptoAtwill: I updated accordingly

fendermint/vm/interpreter/src/fvm/state/snapshot.rs
cryptoAtwill: returns the result now

fendermint/vm/interpreter/src/fvm/state/snapshot.rs
cryptoAtwill: I updated slightly to use stream instead. But originally I was thinking the caller, i.e. `Snapshot` should control the root cids.

cryptoAtwill:
@aakoshh @adlrocha I have addressed the comments above and updated the code to have better handling of versioning. Please help take a look.

fendermint/vm/interpreter/Cargo.toml
```diff
@@ -32,11 +32,19 @@ fvm_ipld_blockstore = { workspace = true }
...
 
-[dev-dependencies]
-quickcheck = { workspace = true }
+futures-core = "0.3.28"
+futures-util = "0.3.28"
+libipld = { version = "0.14", default-features = false, features = ["dag-cbor"] }
 tokio = { workspace = true }
+tokio-stream = "0.1.14"
+tokio-util = {version = "0.7.8", features = ["compat"]}
+pin-project = "1.1.2"
```
aakoshh, line 41: I would move these to the top level Cargo.toml file, even if nothing else uses them at the moment. We have a lot of crates already and it might be impossible to spot common ones later.

fendermint/vm/interpreter/src/fvm/state/snapshot.rs
```diff
@@ -0,0 +1,364 @@
...
+pub type BlockHeight = u64;
+
+/// Taking snapshot of the current blockchain state
+pub enum Snapshot<DB> {
+    V1(V1Snapshot<DB>),
+}
+
+/// Contains the overall metadata for the snapshot
+#[derive(Serialize, Deserialize)]
+struct SnapShotMetadata {
```
aakoshh, line 32:
```suggestion
struct SnapshotMetadata {
```

fendermint/vm/interpreter/src/fvm/state/snapshot.rs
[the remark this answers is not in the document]
aakoshh: You gave a thumbs up but didn't change the parameter name, just resolved the comment. I'm not sure whether you intended to change it or not :slightly_smiling_face:

fendermint/vm/interpreter/src/fvm/state/snapshot.rs
aakoshh: I usually include a comment to point the reader at the original, in case they too want to get the inspiration from the horse's mouth.

fendermint/vm/interpreter/src/fvm/state/snapshot.rs
```diff
@@ -0,0 +1,364 @@
...
+                    let ipld = from_slice::<Ipld>(&bytes).unwrap();
+                    walk_ipld_cids(ipld, &mut this.dfs);
+                    return Poll::Ready(Some((cid, bytes)));
+                }
+                Ok(None) => {
+                    tracing::warn!("cid: {cid:?} has no value in block store, skip");
+                    continue;
+                }
+                Err(e) => {
+                    tracing::warn!("cannot get from block store: {}", e.to_string());
```
aakoshh, line 228:
```suggestion
                    tracing::error!("cannot get from block store: {}", e.to_string());
```
I agree that this should ideally return a `Result`, but at least it should log markedly that this is an unexpected error that the operator should look at. The snapshot should not be considered complete if it failed to export completely.

fendermint/vm/interpreter/src/fvm/state/snapshot.rs
```diff
@@ -0,0 +1,364 @@
...
+            };
+
+            match this.bs.get(&cid) {
+                Ok(Some(bytes)) => {
+                    let ipld = from_slice::<Ipld>(&bytes).unwrap();
+                    walk_ipld_cids(ipld, &mut this.dfs);
+                    return Poll::Ready(Some((cid, bytes)));
+                }
+                Ok(None) => {
+                    tracing::warn!("cid: {cid:?} has no value in block store, skip");
```
aakoshh, line 224:
```suggestion
                    tracing::debug!("cid: {cid:?} has no value in block store, skip");
```
I think these will be part of the data model, some CIDs are just that without any reachable value in the database, but this is nothing the operator can do anything about. I would reserve `warn` to things they should keep their eyes on, and demote this to `debug` for times when you are wondering what is going on. There could be unpredictable numbers of these, so we don't want the logs to start sending emails to operators by the thousands every time a snapshot is made.

fendermint/vm/interpreter/src/fvm/state/snapshot.rs
aakoshh: I wouldn't panic because CIDs without data were for example part of the atomic execution protocol at some point.

aakoshh: I was only asking because I was wondering if that CID following is indeed exercised by this code as it is (the HAMT can store all data in a single node if there aren't too many elements), but it looks fine to me :crossed_fingers:

aakoshh (commented):
Awesome! I have only very minor remarks left :clap:

fendermint/vm/interpreter/src/fvm/state/snapshot.rs
aakoshh: The way it is now, with the ability to inspect `version` before trying to deserialize the actual payload, should make it possible to migrate later :+1:

fendermint/vm/interpreter/src/fvm/state/snapshot.rs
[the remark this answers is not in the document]
cryptoAtwill: Wait, I changed them all. Oh, I know, the commit was lost when I was trying to rebase the branch. Will add them back.

fendermint/vm/interpreter/src/fvm/state/snapshot.rs
cryptoAtwill: updated

fendermint/vm/interpreter/src/fvm/state/snapshot.rs
cryptoAtwill: updated

aakoshh (approved):
Thanks for following up on all the comments, great job!

fendermint/vm/interpreter/src/fvm/state/snapshot.rs
```diff
@@ -0,0 +1,364 @@
...
+    /// Write the snapshot to car file
+    pub async fn write_car(self, path: impl AsRef<Path>) -> anyhow::Result<()> {
+        let file = tokio::fs::File::create(path).await?;
+
+        // derive the metadata for the car file, so that the snapshot version can be recorded.
+        let (metadata, snapshot_streamer) = self.into_streamer()?;
+        let (metadata_cid, metadata_bytes) = derive_cid(&metadata)?;
+
+        // create the target car header with the metadata cid as the only root
+        let car = CarHeader::new(vec![metadata_cid], 1);
```
adlrocha, line 90: So the root of our snapshot is the metadata (i.e. the only object identified in the header) and from the metadata we get the root cid of the actual snapshot, right? I think it may be useful to document this, one may have also expected the header to include two cids, one for the metadata and one for the snapshot data.

(That being said, I think I like our approach more, but let's document so it's clear what we are doing)

adlrocha (approved):
Amazing job! A minor comment but I think this is good to go.

fendermint/vm/interpreter/src/fvm/state/snapshot.rs
cryptoAtwill: Comments added!

PR #170, opened 2023-08-01
FM-170: add transaction compat to call query

adlrocha:
Depends on #163 

In order to finalize the tooling integration (and for Remix to be able to make calls to a contract), we had to also accept legacy transaction transactions for call queries.


aakoshh (approved):
:+1:

Issue #168, opened 2023-08-01
IPC: Adapt the genesis deployment of the gateway to use diamond implementation


adlrocha:
We recently implemented a diamond architecture for the gateway (see [deployment](https://github.com/consensus-shipyard/ipc-solidity-actors/blob/dev/scripts/deploy-gateway.template)), and we should adapt the genesis deployment of the actor implemented in https://github.com/consensus-shipyard/fendermint/pull/159 to work with this new contract architecture.

aakoshh:
@adlrocha I'll hopefully be able to figure it out from the new deployment script, which I have not looked at under a magnifier, but I'm trying to understand how people are supposed to deploy subnets with the diamond. 

The `SubnetActor` wasn't deployed in Genesis because it will be a different actor for each subnet users create, even though the code is the same. What happens now that it's been split into a diamond and two facets? Do you deploy 3 contracts? Or should the facets be deployed in Genesis? How does it work with addresses?

PR #174, opened 2023-08-02
FM-168: IPC solidity diamond

aakoshh:
Closes https://github.com/consensus-shipyard/fendermint/issues/168

Builds on https://github.com/consensus-shipyard/fendermint/pull/166

The PR deals with the diamond pattern during the deployment of IPC contracts in genesis. The change is that top level contracts can have facets, which cannot be discovered from the ABI, so they are hardcoded in the Genesis interpreter at the moment. The facets are taken into account as libraries that can have dependencies, and then passed to the top level diamond contracts as constructor arguments.


fendermint/vm/actor_interface/src/diamond.rs
```diff
@@ -0,0 +1,28 @@
...
+// See https://medium.com/@MarqyMarq/how-to-implement-the-diamond-standard-69e87dae44e6
+
+use std::collections::HashMap;
+
+use ethers::abi::Abi;
+use fvm_shared::ActorID;
+
+#[derive(Clone, Debug)]
+pub struct EthFacet {
+    pub name: &'static str,
```
adlrocha, line 14: Noob Rust question, why use a `'static` here?

fendermint/vm/actor_interface/src/ipc.rs
```diff
@@ -5,14 +5,70 @@
...
 define_id!(SUBNETREGISTRY { id: 65 });
 
+lazy_static! {
+    pub static ref IPC_CONTRACTS: EthContractMap = {
+        [
+            (
+                "GatewayDiamond",
+                EthContract {
+                    actor_id: GATEWAY_ACTOR_ID,
+                    abi: ia::gateway_diamond::GATEWAYDIAMOND_ABI.to_owned(),
```
adlrocha, line 24: @aakoshh, just a quick note, @dnkolegov was working on [an additional facet](https://github.com/consensus-shipyard/ipc-solidity-actors/pull/171/files) to the gateway to reduce even more the size of the contract. I will review the PR next, I don't know if you want to include it already in this PR or have a follow-up once is merged.

adlrocha (approved):
LGTM. This is great, we should try to deploy a child subnet with Fendermint and connect to the agent see what happens. Thanks!

fendermint/vm/actor_interface/src/diamond.rs
aakoshh: I could use `String` which would be owned by the `EthFacet` and could be anything, and I'd have to declare it like `name: "Gateway".to_owned(), ` or `.into()` or `.to_string()`, or I can use what you normally see as `&str` and declare it like `name: "Gateway", `, but in that case because this isn't a parameter to a function, I have to say what the lifetime of this reference is, and `&'static` means the lifetime is that of the entire program. This means that these are statically defined names typed into the editor, not coming from a file or something. 

We can change this if we have to, even use macros to construct this thing to have a nice DSL, but this was my first idea: that you have to describe in code what these contract's look like, like you do in your deployment scripts, so `&'static` is fine.

fendermint/vm/actor_interface/src/ipc.rs
aakoshh: I would merge this as it is and open a new PR. That PR will make it much easier to see what are the typical changes needed when something like this happens.

PR #177, opened 2023-08-04
Cron snapshot the chain

cryptoAtwill:
This PR builds on top of #167.
It is taking the snapshot of the chain at the end of a block.


fendermint/app/src/app.rs
```diff
@@ -151,6 +155,35 @@ where
...
+    ) -> Result<Self> {
+        let app = Self {
+            db: Arc::new(db),
+            state_store: Arc::new(state_store),
+            multi_engine: Arc::new(MultiEngine::new(1)),
+            builtin_actors_bundle,
+            snapshot: Some(SnapshotManager::new(snapshot_config)),
             namespace: app_namespace,
             state_hist: KVCollection::new(state_hist_namespace),
             state_hist_size,
```
aakoshh, line 189: How about instead of repeating the whole constructor and potentially doing all sorts of combinations for future optional features, you leave `new` as it is, and add separate `with_snapshot` method that takes the snapshot config and initiates the manager, overwriting the previous one, if that is acceptable. It can consume an return `self`, so you can only do this during setup.

fendermint/app/src/app.rs
```diff
@@ -562,6 +595,14 @@ where
             .await
             .context("end failed")?;
 
+        if let Some(snapshot) = &self.snapshot {
+            if snapshot.should_run(request.height as BlockHeight) {
+                let (state_params, block_height) =
+                    self.state_params_at_height(Height::try_from(request.height)?)?;
```
aakoshh, line 601: I don't think at this point the state parameters for `request.height` are already saved to history, so this will most likely end up returning the last committed state. 

I think a better place to kick off snapshotting is in the `commit` method, after the state has been saved to the database. Technically in `end_block` Tendermint hasn't told us to save the changes yet, so we should not be snapshotting just yet.

In `commit` these parameters are available without lookup as well.

fendermint/app/src/snapshot.rs
```diff
@@ -0,0 +1,77 @@
...
+                return false;
+            }
+        };
+
+        let path = self.config.path.clone();
+        let lock = self.lock.clone();
+        tokio::spawn(async move {
+            let r = snapshot.write_car(path).await;
+            lock.store(!STARTED, Ordering::SeqCst);
+            r
```
aakoshh, line 72: Some logs will go a long way here.

fendermint/app/src/snapshot.rs
```diff
@@ -0,0 +1,77 @@
...
+    pub fn new(config: SnapshotConfig) -> Self {
+        Self {
+            config,
+            lock: Arc::new(AtomicBool::new(false)),
+        }
+    }
+
+    pub fn should_run(&self, height: BlockHeight) -> bool {
+        height % self.config.period == 0
+    }
```
aakoshh, line 36: One issue with the way this is that it will create snapshots even while a node is trying to catch up with others by applying historical transactions. In that scenario this will just slow things down, so I'd recommend making this method asynchronous, and if the period says there should be a snapshot make, then first check with the Tendermint API what the sync status is, like [here](https://github.com/consensus-shipyard/fendermint/blob/main/fendermint/eth/api/src/apis/eth.rs#L649-L651), and only return `true` if this node is in sync with the others.

For that the `SnapshotManager` will need a Tendermint RPC client pointing at the current node, but we'll need this kind of thing for other purposes as well (broadcasting transactions), so this is natural.

fendermint/app/src/snapshot.rs
aakoshh, line 72: Why return anything from this task if it's just dropped?

fendermint/app/src/snapshot.rs
```diff
@@ -0,0 +1,77 @@
...
+            }
+        };
+
+        let path = self.config.path.clone();
+        let lock = self.lock.clone();
+        tokio::spawn(async move {
+            let r = snapshot.write_car(path).await;
+            lock.store(!STARTED, Ordering::SeqCst);
+            r
+        });
```
aakoshh, line 73: How about freeing up older snapshots? They recommend keeping 2, but this will keep filling up the disk.

aakoshh (commented):
Good start! I wonder how this will work with garbage collection. We need to keep in mind not to delete from the blockstore stuff that is being snapshotted, so maybe the manager can have a function to answer if there is a certain block height it is snapshotting. 

I'd also consider whether it's better to have an optional manager, or one that is non-optional, but doesn't do snapshotting if e.g. the `period` is set to `0`. I usually use `0` to disable periodic stuff, such as the history pruning.

fendermint/app/src/app.rs
```diff
@@ -102,6 +103,9 @@ where
     ///
     /// Only loaded once during genesis; later comes from the [`StateTree`].
     builtin_actors_bundle: PathBuf,
+    /// Manages snapshot taking for the blockchain. If None, it will not trigger snapshot of the
```
adlrocha, line 106:
```suggestion
    /// Manages the creation of periodic snapshots for the network. If None, it will not trigger the creation of snapshots for the
```

fendermint/app/src/app.rs
adlrocha: ^ +1

fendermint/app/src/app.rs
adlrocha: Agree with @aakoshh, we should trigger the snapshot either when we commit the state, or at the beginning of the next block (I personally think that the former makes more sense).

adlrocha (changes requested):
Amazing job, @cryptoAtwill . I think @aakoshh's review mostly covers everything. I guess the two critical things to address before getting this merge are: 
* The use of an async snapshot generation while a peer is syncing the chain from scratch (or maybe even disabling snapshot generation, and just make a snapshot for the latest state after syncing from scratch). I think this is really important to prevent the kind of issues Lotus is currently facing while syncing peers.
* And the garbage collection of snapshots (keeping the latest 2 as Akosh suggests would make sense, although I would propose making it configurable if possible).

fendermint/app/src/snapshot.rs
```diff
@@ -0,0 +1,77 @@
...
+use crate::BlockHeight;
+use fendermint_vm_interpreter::fvm::state::{FvmStateParams, Snapshot};
+use fvm_ipld_blockstore::Blockstore;
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::Arc;
+
+/// The config params for snapshot
+#[derive(Clone)]
+pub struct SnapshotConfig {
+    path: String,
```
aakoshh, line 13:
```suggestion
    path: PathBuf,
```

I encourage that we store the most specific type in this config and push out validation to the time where the app config is parsed, rather than maintain the possibility that it's an invalid path all along the way.

The other remark is that it could be made clear by the name whether this is pointing at a directory or what.

fendermint/app/src/snapshot.rs
```diff
@@ -0,0 +1,77 @@
...
+            Ok(_) => {}
+            Err(_) => {
+                tracing::info!(
+                    "triggering snapshot at height: {block_height} fails as it's already running"
+                );
+                return false;
+            }
+        };
+
+        let path = self.config.path.clone();
```
aakoshh, line 67: This `path` seems like it's going to be the same for every snapshot. I see from the [implementation](https://github.com/consensus-shipyard/fendermint/blob/5faf85d841b94d79cd6b936ab922c1005525b2df/fendermint/vm/interpreter/src/fvm/state/snapshot.rs#L82-L83) that it is a path a file, not a directory, which means that every snapshot will overwrite the previous one. 

The problem is that this happens as soon as the `write_car` method starts, which has multiple negative consequences:
* If anything goes wrong, we lost our only working snapshot
* The file is created and it looks like a ready snapshot even though it's still being written to
* If anything goes wrong we might end up with a partial, invalid snapshot
* It's impossible to tell from the file name what the snapshot is for

Ideally the path would point at a directory, where we can store snapshots that bear the name of the epoch for which they were made. When we start writing a snapshot, it should go into a temporary file, which is renamed to its final form after it's ready. It should be easy to filter and sort snapshots in the ABCI methods to only offer the latest complete ones, and to remove the oldest ones when a new one is ready. We should also be able to store checksum files next to the snapshots in the directory.

aakoshh:
Going to work on this in https://github.com/consensus-shipyard/fendermint/issues/353

PR #178, opened 2023-08-04
Use exec state

cryptoAtwill:
This addresses issue #176.

This PR removes the state_tree option and just use exec_state, which also exposes a state_tree_mut() method now, to avoid any discrepancy, and then you can chain call and actor_state calls freely.

### DIFF
--- a/fendermint/vm/interpreter/src/fvm/state/query.rs
+++ b/fendermint/vm/interpreter/src/fvm/state/query.rs
@@ -8,7 +8,7 @@ use anyhow::{anyhow, Context};
 use cid::Cid;
 use fendermint_vm_actor_interface::system::SYSTEM_ACTOR_ADDR;
 use fendermint_vm_message::query::ActorState;
-use fvm::{engine::MultiEngine, executor::ApplyRet, state_tree::StateTree};
+use fvm::{engine::MultiEngine, executor::ApplyRet};
 use fvm_ipld_blockstore::Blockstore;
 use fvm_shared::{address::Address, clock::ChainEpoch, ActorID};
 use num_traits::Zero;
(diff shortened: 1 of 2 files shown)


fendermint/vm/interpreter/src/fvm/state/query.rs
```diff
@@ -125,8 +103,13 @@ where
...
 
     /// Get the state of an actor, if it exists.
-    pub fn actor_state(&self, addr: &Address) -> anyhow::Result<Option<(ActorID, ActorState)>> {
-        self.with_state_tree(|state_tree| {
+    pub fn actor_state(
+        &self,
+        use_cache: bool,
+        addr: &Address,
+    ) -> anyhow::Result<Option<(ActorID, ActorState)>> {
+        self.with_exec_state(use_cache, |exec_state| {
```
adlrocha, line 111: I will probably let @aakoshh have the final call in this one because I may be missing something, so let's wait for his review, but do we really need `use_cache` as a public argument of `actor_state`? Isn't `use_cache` always `false` in this case? 
```suggestion
        self.with_exec_state(false, |exec_state| {
```

adlrocha (approved):
Thanks! Let's wait for @aakoshh review before merging.

fendermint/vm/interpreter/src/fvm/state/query.rs
aakoshh: I think it makes sense to offer it, the calls can now influence the result, and you might want to query it taking the last call into account.

fendermint/vm/interpreter/src/fvm/state/query.rs
```diff
@@ -154,7 +137,7 @@ where
     pub fn call(&self, mut msg: FvmMessage, use_cache: bool) -> anyhow::Result<ApplyRet> {
         // If the sequence is zero, treat it as a signal to use whatever is in the state.
         if msg.sequence.is_zero() {
-            if let Some((_, state)) = self.actor_state(&msg.from)? {
+            if let Some((_, state)) = self.actor_state(use_cache, &msg.from)? {
```
aakoshh, line 140: It this was moved inside the with_exec_state then it wouldn’t have to initialise the execution state twice.

aakoshh (approved):
Nice, thanks for picking this up Will!

fendermint/vm/interpreter/src/fvm/state/query.rs
cryptoAtwill: yeah, @aakoshh, I removed the `actor_state` call and directly used the `exec_state`.

PR #181, opened 2023-08-08
Top down discussion

cryptoAtwill:
Migrate the top down message execution to new design with Fendermint.

### DIFF
--- a/fendermint/app/src/app.rs
+++ b/fendermint/app/src/app.rs
@@ -9,7 +9,8 @@ use async_trait::async_trait;
 use bytes::Bytes;
 use cid::Cid;
 use fendermint_abci::{AbciResult, Application};
-use fendermint_ipc::pof::ProofOfFinality;
+use fendermint_ipc::parent::PollingParentSyncer;
+use fendermint_ipc::pof::ParentFinalityProvider;
 use fendermint_ipc::IPCMessage;
 use fendermint_storage::{
     Codec, Encode, KVCollection, KVRead, KVReadable, KVStore, KVWritable, KVWrite,
@@ -130,7 +131,7 @@ where
     /// Zero means unlimited.
     state_hist_size: u64,
     /// The top down parent finality.
-    parent_finality: ProofOfFinality,
+    parent_finality: ParentFinalityProvider<PollingParentSyncer>,
 }
 
 impl<DB, SS, S, I> App<DB, SS, S, I>
@@ -299,24 +300,25 @@ where
         Ok((state.state_params, state.block_height))
     }
 
-    async fn verify_ipc_message(&self, ipc_message: IPCMessage) -> bool {
+    fn verify_ipc_message(&self, ipc_message: IPCMessage) -> bool {
         match ipc_message {
-            IPCMessage::TopDown(finality) => self.parent_finality.check_finality(&finality).await,
+            IPCMessage::TopDown(finality) => self.parent_finality.check_finality(&finality),
             IPCMessage::BottomUp => unimplemented!(),
         }
     }
 
     /// Verifies the list of transactions to see if they are valid. Returns true if all the txns
     /// are valid and false if any one of the txns is invalid.
-    async fn verify_messages(&self, txns: &Vec<Bytes>) -> bool {
+    fn verify_messages(&self, txns: &Vec<Bytes>) -> bool {
         for tx in txns {
             let is_ok = match serde_json::from_slice::<ChainMessage>(tx) {
                 Err(_) => {
+                    // FIXME: maybe we can reject the proposal when the transaction cannot be deserialized
                     tracing::info!("cannot deserialized transaction: {tx:?}");
                     continue;
                 }
                 Ok(message) => match message {
-                    ChainMessage::IPC(ipc) => self.verify_ipc_message(ipc).await,
+                    ChainMessage::IPC(ipc) => self.verify_ipc_message(ipc),
                     _ => continue,
                 },
             };
@@ -527,6 +529,8 @@ where
         Ok(response)
     }
 
+    /// 1. Prepare txns
+    /// 2. Prepare parent finality proposal
     async fn prepare_proposal(
         &self,
         request: request::PrepareProposal,
@@ -542,10 +546,14 @@ where
             txs.push(tx);
         }
 
-        let proof = self.parent_finality.get_finality();
-        let bytes = serde_json::to_vec(&ChainMessage::IPC(IPCMessage::TopDown(proof)))
-            .expect("should not have failed");
-        txs.push(bytes.into());
+        if let Some(finality_proposal) = self
+            .parent_finality
+            .next_finality_proposal()
+            .map_err(|e| anyhow!("cannot get next finality to propose due to: {e}"))? {
+
+            let bytes = serde_json::to_vec(&ChainMessage::IPC(IPCMessage::TopDown(finality_proposal)))?;
+            txs.push(bytes.into());
+        }
 
         Ok(response::PrepareProposal { txs })
     }
@@ -554,7 +562,7 @@ where
         &self,
         request: request::ProcessProposal,
     ) -> AbciResult<response::ProcessProposal> {
-        if !self.verify_messages(&request.txs).await {
+        if !self.verify_messages(&request.txs) {
             Ok(response::ProcessProposal::Reject)
         } else {
             Ok(response::ProcessProposal::Accept)
@@ -609,6 +617,7 @@ where
                     );
                     to_deliver_tx(ret)
                 }
+                _ => todo!(),
             },
         };
 
--- a/fendermint/app/src/app.rs
+++ b/fendermint/app/src/app.rs
@@ -549,9 +549,10 @@ where
         if let Some(finality_proposal) = self
             .parent_finality
             .next_finality_proposal()
-            .map_err(|e| anyhow!("cannot get next finality to propose due to: {e}"))? {
-
-            let bytes = serde_json::to_vec(&ChainMessage::IPC(IPCMessage::TopDown(finality_proposal)))?;
+            .map_err(|e| anyhow!("cannot get next finality to propose due to: {e}"))?
+        {
+            let bytes =
+                serde_json::to_vec(&ChainMessage::IPC(IPCMessage::TopDown(finality_proposal)))?;
             txs.push(bytes.into());
         }
 
--- a/fendermint/app/src/app.rs
+++ b/fendermint/app/src/app.rs
@@ -6,6 +6,7 @@ use std::sync::{Arc, Mutex};
 
 use anyhow::{anyhow, Context, Result};
 use async_trait::async_trait;
+use bytes::Bytes;
 use cid::Cid;
 use fendermint_abci::{AbciResult, Application};
 use fendermint_storage::{
@@ -297,6 +298,36 @@ where
         let state = self.committed_state()?;
         Ok((state.state_params, state.block_height))
     }
+
+    async fn verify_ipc_message(&self, ipc_message: IPCMessage) -> bool {
+        match ipc_message {
+            IPCMessage::TopDown(finality) => {
+                self.parent_finality.check_finality(&finality).await
+            }
+            IPCMessage::BottomUp => unimplemented!()
+        }
+    }
+
+    /// Verifies the list of transactions to see if they are valid. Returns true if all the txns
+    /// are valid and false if any one of the txns is invalid.
+    async fn verify_messages(&self, txns: &Vec<Bytes>) -> bool {
+        for tx in txns {
+            let is_ok = match serde_json::from_slice::<ChainMessage>(tx) {
+                Err(_) => {
+                    tracing::info!("cannot deserialized transaction: {tx:?}");
+                    continue
+                },
+                Ok(message) => match message {
+                    ChainMessage::IPC(ipc) => self.verify_ipc_message(ipc).await,
+                    _ => continue,
+                }
+            };
+
+            if !is_ok { return false; }
+        }
+
+        true
+    }
 }
 
 // NOTE: The `Application` interface doesn't allow failures at the moment. The protobuf
@@ -511,12 +542,23 @@ where
             txs.push(tx);
         }
 
-        let proof = self.parent_finality.get_proof();
+        let proof = self.parent_finality.get_finality();
         let bytes = serde_json::to_vec(&ChainMessage::IPC(IPCMessage::TopDown(proof)))
             .expect("should not have failed");
         txs.push(bytes.into());
 
         Ok(response::PrepareProposal { txs })
+    }
+
+    async fn process_proposal(
+        &self,
+        request: request::ProcessProposal,
+    ) -> AbciResult<response::ProcessProposal> {
+        if !self.verify_messages(&request.txs).await {
+            Ok(response::ProcessProposal::Reject)
+        } else {
+            Ok(response::ProcessProposal::Accept)
+        }
     }
 
     /// Signals the beginning of a new block, prior to any `DeliverTx` calls.
--- a/fendermint/app/src/app.rs
+++ b/fendermint/app/src/app.rs
@@ -34,6 +34,9 @@ use serde::{Deserialize, Serialize};
 use tendermint::abci::request::CheckTxKind;
 use tendermint::abci::{request, response};
 use tendermint::block::Height;
+use fendermint_ipc::IPCMessage;
+use fendermint_ipc::pof::ProofOfFinality;
+use fendermint_vm_message::chain::ChainMessage;
 
 use crate::{tmconv::*, VERSION};
 use crate::{BlockHeight, APP_VERSION};
@@ -125,6 +128,8 @@ where
     ///
     /// Zero means unlimited.
     state_hist_size: u64,
+    /// The top down parent finality.
+    parent_finality: ProofOfFinality,
 }
 
 impl<DB, SS, S, I> App<DB, SS, S, I>
@@ -157,6 +162,7 @@ where
             interpreter: Arc::new(interpreter),
             exec_state: Arc::new(Mutex::new(None)),
             check_state: Arc::new(tokio::sync::Mutex::new(None)),
+            parent_finality: (),
         };
         app.init_committed_state()?;
         Ok(app)
@@ -488,6 +494,29 @@ where
         };
 
         Ok(response)
+    }
+
+    async fn prepare_proposal(
+        &self,
+        request: request::PrepareProposal,
+    ) -> AbciResult<response::PrepareProposal> {
+        let max_tx_bytes: usize = request.max_tx_bytes.try_into().unwrap();
+        let mut size: usize = 0;
+        let mut txs = Vec::new();
+        for tx in request.txs {
+            if size.saturating_add(tx.len()) > max_tx_bytes {
+                break;
+            }
+            size += tx.len();
+            txs.push(tx);
+        }
+
+        let proof = self.parent_finality.get_proof();
+        let bytes = serde_json::to_vec(&ChainMessage::IPC(IPCMessage::TopDown(proof)))
+            .expect("should not have failed");
+        txs.push(bytes.into());
+
+        Ok(response::PrepareProposal { txs })
     }
 
     /// Signals the beginning of a new block, prior to any `DeliverTx` calls.
--- a/fendermint/app/src/app.rs
+++ b/fendermint/app/src/app.rs
@@ -9,6 +9,8 @@ use async_trait::async_trait;
 use bytes::Bytes;
 use cid::Cid;
 use fendermint_abci::{AbciResult, Application};
+use fendermint_ipc::pof::ProofOfFinality;
+use fendermint_ipc::IPCMessage;
 use fendermint_storage::{
     Codec, Encode, KVCollection, KVRead, KVReadable, KVStore, KVWritable, KVWrite,
 };
@@ -25,6 +27,7 @@ use fendermint_vm_interpreter::signed::InvalidSignature;
 use fendermint_vm_interpreter::{
     CheckInterpreter, ExecInterpreter, GenesisInterpreter, QueryInterpreter,
 };
+use fendermint_vm_message::chain::ChainMessage;
 use fvm::engine::MultiEngine;
 use fvm_ipld_blockstore::Blockstore;
 use fvm_shared::chainid::ChainID;
@@ -35,9 +38,6 @@ use serde::{Deserialize, Serialize};
 use tendermint::abci::request::CheckTxKind;
 use tendermint::abci::{request, response};
 use tendermint::block::Height;
-use fendermint_ipc::IPCMessage;
-use fendermint_ipc::pof::ProofOfFinality;
-use fendermint_vm_message::chain::ChainMessage;
 
 use crate::{tmconv::*, VERSION};
 use crate::{BlockHeight, APP_VERSION};
@@ -301,10 +301,8 @@ where
 
     async fn verify_ipc_message(&self, ipc_message: IPCMessage) -> bool {
         match ipc_message {
-            IPCMessage::TopDown(finality) => {
-                self.parent_finality.check_finality(&finality).await
-            }
-            IPCMessage::BottomUp => unimplemented!()
+            IPCMessage::TopDown(finality) => self.parent_finality.check_finality(&finality).await,
+            IPCMessage::BottomUp => unimplemented!(),
         }
     }
 
@@ -315,15 +313,17 @@ where
             let is_ok = match serde_json::from_slice::<ChainMessage>(tx) {
                 Err(_) => {
                     tracing::info!("cannot deserialized transaction: {tx:?}");
-                    continue
-                },
+                    continue;
+                }
                 Ok(message) => match message {
                     ChainMessage::IPC(ipc) => self.verify_ipc_message(ipc).await,
                     _ => continue,
-                }
+                },
             };
 
-            if !is_ok { return false; }
+            if !is_ok {
+                return false;
+            }
         }
 
         true
--- a/fendermint/ipc/src/cache.rs
+++ b/fendermint/ipc/src/cache.rs
@@ -1,0 +1,104 @@
+// Copyright 2022-2023 Protocol Labs
+// SPDX-License-Identifier: Apache-2.0, MIT
+
+use num_traits::{PrimInt};
+use std::cmp::{max, min};
+use std::collections::HashMap;
+use std::hash::Hash;
+
+type Bounds<T> = (T, T);
+
+/// The key value cache. The key must be numeric and falls within a range/bound.
+pub struct RangeKeyCache<Key, Value> {
+    /// Stores the data in a hashmap.
+    data: HashMap<Key, Value>,
+    /// The lower and upper bound of keys stored in data
+    bounds: Option<Bounds<Key>>,
+}
+
+impl<Key: PrimInt + Hash, Value> RangeKeyCache<Key, Value> {
+    pub fn new() -> Self {
+        Self {
+            data: Default::default(),
+            bounds: None,
+        }
+    }
+
+    pub fn upper_bound(&self) -> Option<Key> {
+        self.within_bounds(|(_, upper)| *upper)
+    }
+
+    pub fn get_value(&self, key: Key) -> Option<&Value> {
+        self.within_bounds(|(lower_bound, upper_bound)| {
+            if *lower_bound > key || *upper_bound < key {
+                return None;
+            }
+            return self.data.get(&key);
+        })
+        .flatten()
+    }
+
+    pub fn values_within_range(&self, start: Key, end: Option<Key>) -> Vec<&Value> {
+        self.within_bounds(|(lower_bound, upper_bound)| {
+            let start = max(*lower_bound, start);
+            let end = min(*upper_bound, end.unwrap_or(*upper_bound));
+
+            let mut r = vec![];
+            let mut i = start;
+            while i <= end {
+                if let Some(v) = self.get_value(i) {
+                    r.push(v);
+                }
+
+                i = i + Key::one();
+            }
+
+            r
+        })
+        .unwrap_or(vec![])
+    }
+
+    /// Removes the block hashes stored till the specified height, exclusive.
+    pub fn remove_key_till(&mut self, key: Key) {
+        if let Some((lower_bound, upper_bound)) = self.bounds.as_mut() {
+            if *lower_bound > key || *upper_bound < key {
+                return;
+            }
+
+            let mut i = *lower_bound;
+            while i < key {
+                self.data.remove(&i);
+                i = i + Key::one();
+            }
+
+            *lower_bound = key;
+        }
+    }
+
+    /// Insert the block hash at the next height
+    pub fn insert_after_lower_bound(&mut self, key: Key, val: Value) {
+        match &mut self.bounds {
+            None => {
+                self.data.insert(key, val);
+                self.bounds.replace((key, key));
+            }
+            Some((upper, lower)) => {
+                if *lower > key {
+                    return;
+                }
+
+                self.data.insert(key, val);
+                if *upper < key {
+                    *upper = key;
+                }
+            }
+        }
+    }
+
+    fn within_bounds<F, T>(&self, f: F) -> Option<T>
+    where
+        F: Fn(&(Key, Key)) -> T,
+    {
+        self.bounds.as_ref().map(f)
+    }
+}
--- a/fendermint/ipc/src/lib.rs
+++ b/fendermint/ipc/src/lib.rs
@@ -2,17 +2,55 @@
 // SPDX-License-Identifier: Apache-2.0, MIT
 //! Interfacing with IPC, provides utility functions
 
+mod agent;
 mod message;
-mod parent;
+pub mod parent;
 pub mod pof;
 
+use async_trait::async_trait;
+use ipc_sdk::cross::CrossMsg;
+use ipc_sdk::ValidatorSet;
 pub use message::IPCMessage;
+use crate::parent::PollingParentSyncer;
+use crate::pof::IPCParentFinality;
 
 #[derive(Debug, Clone)]
 pub struct Config {
     /// The number of blocks to delay reporting when creating the pof
-    chain_head_delay: u64,
+    chain_head_delay: BlockHeight,
+    /// The lower bound for the chain head height in parent view
+    chain_head_lower_bound: BlockHeight,
 
     /// Parent syncing cron period, in seconds
     polling_interval: u64,
+}
+
+type BlockHeight = u64;
+
+/// Obtain the latest state information required for IPC from the parent subnet
+pub trait ParentViewProvider {
+    /// Fetch the latest chain head
+    fn latest_height(&self) -> Option<BlockHeight>;
+    /// Fetch the block hash at target height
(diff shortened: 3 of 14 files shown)


fendermint/ipc/src/pof.rs
```diff
@@ -0,0 +1,150 @@
...
+        }
+    };
+}
+
+async fn sync_with_parent<T: ParentSyncer>(
+    config: Config,
+    parent_syncer: Arc<T>,
+    lock: LockedProof,
+) -> anyhow::Result<()> {
+    let mut interval = tokio::time::interval(Duration::from_secs(config.polling_interval));
```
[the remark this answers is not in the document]
adlrocha: I think this functionality shouldn't be here but part of the `ParentSyncer` implementation, right? (maybe I missed something)

adlrocha (commented):
I just did a first  pass to see if it helps with the implementation. I think we should decouple the `ParentView`, i.e. the perceived finality of the parent seen by the peer, from the actual proof of finality committed and agreed upon by all validators. I think these concepts are not decoupled in the code and is quite confusing to follow. 

My suggestion:
* `ParentSyncer` is the piece of code responsible for syncing with the parent and keeping an up to date view of the perceived parent finality of the peer. Any module that needs to perform queries about what is the current finality we see in the parent queries this module.
*  `IPCParentFinality` is the piece responsible for orchestrating the creation of PoFs and keeps track of what is the latest finality all the validators have so far agreed upon.

Let me know if this makes sense, we can also discuss sync if you think is easier.

fendermint/ipc/src/pof.rs
```diff
@@ -0,0 +1,150 @@
...
+    // top_down_msgs: Vec<CrossMsg>,
+    // /// latest configuration information from the parent.
+    // config: MembershipSet,
+}
+
+/// The proof of finality util struct
+pub struct ProofOfFinality<T> {
+    config: Config,
+    started: Arc<AtomicBool>,
+    latest_proof: LockedProof,
```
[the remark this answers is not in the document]
cryptoAtwill: maybe I should rename this to `latest_parent_view`

fendermint/app/src/app.rs
```diff
@@ -490,6 +527,40 @@ where
...
 
+    async fn prepare_proposal(
+        &self,
+        request: request::PrepareProposal,
+    ) -> AbciResult<response::PrepareProposal> {
+        let max_tx_bytes: usize = request.max_tx_bytes.try_into().unwrap();
+        let mut size: usize = 0;
+        let mut txs = Vec::new();
+        for tx in request.txs {
+            if size.saturating_add(tx.len()) > max_tx_bytes {
```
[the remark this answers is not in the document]
aakoshh: I would think so.

fendermint/app/src/app.rs
```diff
@@ -291,6 +299,37 @@ where
         let state = self.committed_state()?;
         Ok((state.state_params, state.block_height))
     }
+
+    fn verify_ipc_message(&self, ipc_message: IPCMessage) -> bool {
```
aakoshh, line 303: This can now be moved into the `ChainMessageInterpreter` in the implementation of the `ProposalInterpreter` added in https://github.com/consensus-shipyard/fendermint/pull/185

My goal with `app.rs` was to keep its size and scope to a minimum, it's arguably too big already. Ideally it should only be an implementation of the `Application` trait and dispatch to other components. Everything it does is in support of this, mainly state management and conversions to Tendermint format. 

It is for this reason that the top-level interpreter which deals with transactions and genesis receives binary _input_: to make `app.rs` oblivious to the actual payloads we are working with. It has to know the _output_ format in order to be able to translate it to Tendermint types. 

I'd like us to maintain this separation of concerns and move the finality checks into the interpreter. This can be achieved in multiple ways:

The _parent view_ or `parent_finality` can be added to the `ChainMessageInterpreter` during application construction, instead of the `App` struct, but ideally only if it's a _stateless_ component. If parent finality checking needs some state, it should be lifted into the `App` for safekeeping, and passed to the interpreter during invocation. I would try using STM for this, so the interpreter can read the state of the parent chain from data structures maintained by another background component, which can be started `run.rs`. This same STM enabled componet can also be a shared with the transaction processing where we can add things CIDs to look up, for another background process to handle.

fendermint/app/src/app.rs
```diff
@@ -291,6 +299,37 @@ where
...
+            IPCMessage::TopDown(finality) => self.parent_finality.check_finality(&finality),
+            IPCMessage::BottomUp => unimplemented!(),
+        }
+    }
+
+    /// Verifies the list of transactions to see if they are valid. Returns true if all the txns
+    /// are valid and false if any one of the txns is invalid.
+    fn verify_messages(&self, txns: &Vec<Bytes>) -> bool {
+        for tx in txns {
+            let is_ok = match serde_json::from_slice::<ChainMessage>(tx) {
```
aakoshh, line 314: If you take a look at the `BytesMessageInterpreter` you will see that we use IPLD binary format for messages, not JSON. The Cosmos SDK uses JSON, which may be why you thought Tendermint does as well, but it's actually agnostic, and I thought we should stick to the Filecoin convention of IPLD (not that I don't think JSON wouldn't be more convenient). 

So this check should not be here, or exist in this format, because checking this is exactly the purpose of `BytesMessageInterpreter`. I added this check in https://github.com/consensus-shipyard/fendermint/pull/185

fendermint/app/src/app.rs
```diff
@@ -490,6 +529,47 @@ where
...
+        let max_tx_bytes: usize = request.max_tx_bytes.try_into().unwrap();
+        let mut size: usize = 0;
+        let mut txs = Vec::new();
+        for tx in request.txs {
+            if size.saturating_add(tx.len()) > max_tx_bytes {
+                break;
+            }
+            size += tx.len();
+            txs.push(tx);
+        }
```
aakoshh, line 547: I added this as a utility function to https://github.com/consensus-shipyard/fendermint/pull/185 so there's no need to repeat it in `app.rs`.

fendermint/app/src/app.rs
```diff
@@ -490,6 +529,47 @@ where
...
+
+        if let Some(finality_proposal) = self
```
aakoshh, line 549: Let's move this logic into the `ChainMessageInterpreter`.

fendermint/app/src/app.rs
```diff
@@ -490,6 +529,47 @@ where
...
+            .parent_finality
+            .next_finality_proposal()
+            .map_err(|e| anyhow!("cannot get next finality to propose due to: {e}"))?
```
aakoshh, line 552: How can this method fail? It's only reading data from memory, not doing any IO, right? 

Be very careful when you use `?` in `app.rs` because a failure will crash Fendermint and Tendermint will stop as well.

fendermint/app/src/app.rs
```diff
@@ -490,6 +529,47 @@ where
...
+        {
+            let bytes =
+                serde_json::to_vec(&ChainMessage::IPC(IPCMessage::TopDown(finality_proposal)))?;
```
aakoshh, line 555: Once moved into the `ChainMessageInterpreter` you won't have to convert to binary (not JSON) because that is take care of by the `BytesMessageInterpreter`. 

In any case, it was wrong to put this _after_ the stuff above that summed up all the bytes, because this could have brought the total above the allowed limit. This will not be a problem if you rebase onto https://github.com/consensus-shipyard/fendermint/pull/185 because there the check is done after the interpreters have finished. 

For this reason (the possibility of this check removing a message we wanted to include in the proposal) it is crucial that for example `next_finality_proposal()` actually does things in a read-only fashion, which is why I indicated in the `ProposalInterpreter` that it does not return a modified state (it's a `Reader` monad, not a `State` monad :nerd_face: ), because our proposal might not be voted on, might time out, or may be removed. We have to be prepared to re-propose it. Only the transaction execution should cause changes, e.g. when we see that our proposal went through, then we can move on to the next candidate.

If we used STM, we might actually want to maintain two regions there: a "transient" region modified during `deliver_tx`, and a "permanent" region where it's merged during `commit`.

fendermint/app/src/app.rs
```diff
@@ -490,6 +529,47 @@ where
...
+        }
+
+        Ok(response::PrepareProposal { txs })
+    }
+
+    async fn process_proposal(
+        &self,
+        request: request::ProcessProposal,
+    ) -> AbciResult<response::ProcessProposal> {
+        if !self.verify_messages(&request.txs) {
```
aakoshh, line 566: Move to `ChainMessageInterpreter`.

fendermint/ipc/src/cache.rs
```diff
@@ -0,0 +1,104 @@
...
+
+use num_traits::{PrimInt};
+use std::cmp::{max, min};
+use std::collections::HashMap;
+use std::hash::Hash;
+
+type Bounds<T> = (T, T);
+
+/// The key value cache. The key must be numeric and falls within a range/bound.
+pub struct RangeKeyCache<Key, Value> {
```
aakoshh, line 12:
```suggestion
pub struct RangeKeyCache<K, V> {
```

I think these won't be harder to read, and are well established?

fendermint/ipc/src/cache.rs
```diff
@@ -0,0 +1,104 @@
...
+    /// Stores the data in a hashmap.
+    data: HashMap<Key, Value>,
```
aakoshh, line 14: You could use a `BTreeMap` to store data in-order, since it looks like you will be iterating over it.

fendermint/ipc/src/cache.rs
```diff
@@ -0,0 +1,104 @@
...
+        self.within_bounds(|(lower_bound, upper_bound)| {
+            if *lower_bound > key || *upper_bound < key {
+                return None;
+            }
+            return self.data.get(&key);
+        })
+        .flatten()
+    }
+
+    pub fn values_within_range(&self, start: Key, end: Option<Key>) -> Vec<&Value> {
```
aakoshh, line 41: https://doc.rust-lang.org/std/collections/struct.BTreeMap.html#method.range

fendermint/ipc/src/cache.rs
```diff
@@ -0,0 +1,104 @@
...
+                i = i + Key::one();
+            }
+
+            r
+        })
+        .unwrap_or(vec![])
+    }
+
+    /// Removes the block hashes stored till the specified height, exclusive.
+    pub fn remove_key_till(&mut self, key: Key) {
```
aakoshh, line 62: https://doc.rust-lang.org/std/collections/struct.BTreeMap.html#method.split_off

fendermint/ipc/src/cache.rs
```diff
@@ -0,0 +1,104 @@
...
+                self.data.remove(&i);
+                i = i + Key::one();
+            }
+
+            *lower_bound = key;
+        }
+    }
+
+    /// Insert the block hash at the next height
+    pub fn insert_after_lower_bound(&mut self, key: Key, val: Value) {
```
aakoshh, line 79: I reckon the implementation of this cache can be simplified to a struct which has a `BTreeMap` and a minimum established height. As I understand those are the operations you want: to clear out data up to a lower bound, and never to re-insert it. The upper bound seems like accidental complexity.

fendermint/app/src/app.rs
```diff
@@ -291,6 +299,37 @@ where
...
+    }
+
+    /// Verifies the list of transactions to see if they are valid. Returns true if all the txns
+    /// are valid and false if any one of the txns is invalid.
+    fn verify_messages(&self, txns: &Vec<Bytes>) -> bool {
+        for tx in txns {
+            let is_ok = match serde_json::from_slice::<ChainMessage>(tx) {
+                Err(_) => {
+                    // FIXME: maybe we can reject the proposal when the transaction cannot be deserialized
+                    tracing::info!("cannot deserialized transaction: {tx:?}");
```
aakoshh, line 317: This was also my first intuition, but then I thought we can also let the proposal pass and punish the validator during execution.

fendermint/ipc/src/message.rs
```diff
@@ -0,0 +1,19 @@
+// Copyright 2022-2023 Protocol Labs
+// SPDX-License-Identifier: Apache-2.0, MIT
+
+//! IPC messages
+
+use crate::pof::IPCParentFinality;
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+pub enum IPCMessage {
```
aakoshh, line 10: Let's move the IPC related messages (and messages only) into the `fendermint_vm_message` crate, so that doesn't have to include references to crates with implementation such as the IPC Agent proxy. It should be a crate that anyone can add as a dependency to be able to inspect chain messages, until we figure out a way to make them more generic.

fendermint/ipc/src/message.rs
```diff
@@ -0,0 +1,16 @@
...
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub enum IPCMessage {
+    TopDown(IPCParentFinality),
+    BottomUp,
+}
+
+impl From<IPCMessage> for Vec<u8> {
+    fn from(value: IPCMessage) -> Self {
+        serde_json::to_vec(&value).expect("should not happen")
```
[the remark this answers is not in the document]
aakoshh: Ah, this is super confusing actually. I read it as _parsing_ `Vec<u8>` like 3 times already. 

So yes, this will (probably) not fail because we can always convert to JSON, and this isn't the case when we're running the risk of a malicious validator passing us some untrustworthy byte content. However, I'd argue this implementation should not exist: it is not an inherent property of `IPCMessage` that it's transformed to binary via JSON. Case in point: we actually use IPLD. Let's not decide for everyone what they should use, and leave it up to the application implementer.

fendermint/ipc/src/message.rs
aakoshh: On a side note: I always struggle to decide between using acronyms in PascalCasing names, but I usually go with only capitalising the first letter, e.g. `ProductId`, `XmlHttpRequest`, `JsonRpcServer`, rather than `ProductID`, `XMLHTTPRequest`, `JSONRPCServer`. I see there are all sorts of variations out there, but I found this to be a perhaps less aesthetically pleasing, but at least consistent way. 

Have you guys established the `IPC*` prefix?

fendermint/ipc/src/parent.rs
```diff
@@ -0,0 +1,202 @@
...
+        })
+    }
+
+    fn membership(&self) -> Option<ValidatorSet> {
+        self.read_cache(|cache| cache.membership.clone())
+    }
+
+    fn on_finality_committed(&self, finality: &IPCParentFinality) {
+        let mut cache = self.cache.write().unwrap();
+        cache.block_hash.remove_key_till(finality.height);
```
aakoshh, line 67: What about the other caches, when do they get cleared?

fendermint/ipc/src/parent.rs
```diff
@@ -0,0 +1,202 @@
...
+        interval.tick().await;
+
+        let hashes = fetch_block_hashes(&config, &agent_proxy, &lock).await?;
+        update_top_down_msgs(&config, &agent_proxy, &lock).await?;
+        update_membership(&config, &agent_proxy, &lock).await?;
+
+        let mut cache = lock.write().unwrap();
+        for r in hashes {
+            cache.block_hash.insert_after_lower_bound(r.0, r.1);
+        }
```
aakoshh, line 139: Is there a reason why hashes are inserted after all the other updates? Why not have `update_block_hashes` instead of `fetch_block_hashes`?

fendermint/ipc/src/parent.rs
```diff
@@ -0,0 +1,202 @@
...
+    lock: LockedCache,
+) -> anyhow::Result<()> {
+    let mut interval = tokio::time::interval(Duration::from_secs(config.polling_interval));
+
+    loop {
+        interval.tick().await;
+
+        let hashes = fetch_block_hashes(&config, &agent_proxy, &lock).await?;
+        update_top_down_msgs(&config, &agent_proxy, &lock).await?;
+        update_membership(&config, &agent_proxy, &lock).await?;
```
aakoshh, line 134: I recommend using `.context(...)` liberally to attach some message about which of these failed.

fendermint/ipc/src/parent.rs
```diff
@@ -0,0 +1,202 @@
...
+            Err(_) => return Err(anyhow!("already started")),
+        }
+
+        let parent_syncer = self.ipc_agent_proxy.clone();
+        let config = self.config.clone();
+        let cache = self.cache.clone();
+
+        let handle =
+            tokio::spawn(async move { sync_with_parent(config, parent_syncer, cache).await });
+        self.handle = Some(handle);
```
aakoshh, line 116: It looks like if the syncer fails, the `handle` will contain the error, but there will be no feedback anywhere in the logs about the error, I assume until the application is shut down and _potentially_ the `handle` is awaited upon. 

I'd say if it is allowed to fail, it should at least log the error here, but even better would be not allowing it to fail and retrying later.

fendermint/ipc/src/parent.rs
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+    let starting_height = {
+        let cache = lock.read().unwrap();
+        // if cache.latest_height() is None, it means we have not started fetching any heights
+        // we just use the latest height minus a lower bound as the parent view
+        let starting_height = cache.block_hash.upper_bound().unwrap_or(max(
+            1,
+            latest_height.saturating_sub(config.chain_head_lower_bound),
+        ));
```
aakoshh, line 186: It's great that you thought about where to start from. I do think though that this behaviour could be improved. 

After a restart, syncing with the parent should resume from where it was left off, that is, before the syncer is started, we should look up the state of our contract that ultimately tracks the application of top-down messages and retrieve the last known final block height and nonce, and go from there. Otherwise you might stop your node for maintenance, bring it back later, and miss some of the messages simply because more than the `chain_head_lower_bound` amount of time has elapsed.

Another adverse scenario is when you stop your node, restart quickly, and then it replays messages which you already had by effectively resetting the lower bound to an earlier value!

fendermint/ipc/src/lib.rs
```diff
@@ -0,0 +1,58 @@
...
+use ipc_sdk::cross::CrossMsg;
+use ipc_sdk::ValidatorSet;
+pub use message::IPCMessage;
+
+#[derive(Debug, Clone)]
+pub struct Config {
+    /// The number of blocks to delay reporting when creating the pof
+    chain_head_delay: BlockHeight,
+    /// The lower bound for the chain head height in parent view
+    chain_head_lower_bound: BlockHeight,
```
aakoshh, line 22: It would be helpful to explain here whether this is an _absolute_ value or a _relative_ one compared to the latest height.

fendermint/ipc/src/parent.rs
```diff
@@ -0,0 +1,202 @@
...
+    config: &Config,
+    agent_proxy: &Arc<AgentProxy>,
+    lock: &LockedCache,
+) -> anyhow::Result<Vec<(BlockHeight, Vec<u8>)>> {
+    let latest_height = match agent_proxy.get_chain_head_height().await {
+        Ok(h) => h,
+        Err(e) => {
+            tracing::warn!("cannot fetch parent chain head due to {e}");
+
+            // not throw errors, caller will retry
```
aakoshh, line 169: I see, then maybe it would be better if this method wasn't allowed to return an error at all, to remove all ambiguity about whether it can fail or not.

fendermint/ipc/src/pof.rs
```diff
@@ -0,0 +1,85 @@
...
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+pub struct IPCParentFinality {
+    /// The latest chain height
+    pub height: u64,
+    /// The block hash. For FVM, it is a Cid. For Evm, it is bytes32.
+    pub block_hash: Vec<u8>,
+    /// new top-down messages finalized in this PoF
+    pub top_down_msgs: Vec<CrossMsg>,
+    /// latest configuration information from the parent.
+    pub config: ValidatorSet,
```
aakoshh, line 23:
```suggestion
    /// latest configuration information from the parent.
    pub validator_set: ValidatorSet,
```
You already call it `membership` in the interface, so that's a good name too, but `config` is such an overloaded term, and most of the time in this codebase it's about static settings of a component, not something as specific as the dynamic validator powers. I understand it's called _configuration_ in the theory papers, but in this context it's not helpful IMO.

fendermint/ipc/src/pof.rs
```diff
@@ -0,0 +1,85 @@
...
+use ipc_sdk::ValidatorSet;
+use serde::{Deserialize, Serialize};
+use std::sync::Arc;
+
+/// The proof for POF.
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+pub struct IPCParentFinality {
+    /// The latest chain height
+    pub height: u64,
+    /// The block hash. For FVM, it is a Cid. For Evm, it is bytes32.
```
aakoshh, line 18: I'm not sure why we mention EVM here. Fendermint will have to deal with Tendermint parents, but not `geth`.

fendermint/ipc/src/parent.rs
```diff
@@ -0,0 +1,202 @@
...
+        self.read_cache(|cache| cache.block_hash.upper_bound())
+    }
+
+    fn block_hash(&self, height: BlockHeight) -> Option<Vec<u8>> {
+        self.read_cache(|cache| cache.block_hash.get_value(height).map(|v| v.to_vec()))
+    }
+
+    fn top_down_msgs(&self, _height: BlockHeight, nonce: Nonce) -> Vec<CrossMsg> {
+        self.read_cache(|cache| {
+            let v = cache.top_down_message.values_within_range(nonce, None);
```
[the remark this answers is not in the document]
aakoshh: Is it in the `CrossMsg` itself? Must be. Sorry I'm not familiar with the SDK yet.

fendermint/ipc/src/pof.rs
```diff
@@ -0,0 +1,85 @@
...
+    pub top_down_msgs: Vec<CrossMsg>,
+    /// latest configuration information from the parent.
+    pub config: ValidatorSet,
+}
+
+/// The parent finality information provider. It provides the latest finality information about the
+/// parent. Also it perform validation on incoming finality.
+#[derive(Clone)]
+pub struct ParentFinalityProvider<T> {
+    config: Config,
```
aakoshh, line 30: Case in point: here's another `config` field, less than 10 lines away, with a completely different meaning.

fendermint/ipc/src/lib.rs
aakoshh: https://github.com/consensus-shipyard/fendermint/pull/181/files#diff-088fbb2a5e5eeef38c5648a53f47e659eecd568fe77bb93e6b262cfa08801b42R185 

suggests it's a delta, but then what's the difference between this and the "delay" field? :thinking:

fendermint/ipc/src/pof.rs
```diff
@@ -0,0 +1,85 @@
...
+        Ok(())
+    }
+
+    /// The next finality to propose
+    pub fn next_finality_proposal(&self) -> anyhow::Result<Option<IPCParentFinality>> {
+        let latest_height = match self.parent_view_provider.latest_height() {
+            Some(h) => h,
+            None => return Ok(None),
+        };
+        if latest_height < self.config.chain_head_delay {
```
aakoshh, line 48: I tought `chain_head_delay` said it was "number of blocks". How can it be compared with an absolute height?

aakoshh: Ah, this is when the parent chain is lower then the minimum. A comment would help here.

fendermint/ipc/src/parent.rs
```diff
@@ -0,0 +1,202 @@
...
+        tracing::debug!("polling parent from {starting_height} till {latest_height}");
+
+        starting_height
+
+        // ensure cache is dropped
+    };
+
+    // FIXME: make the fetching batching and concurrent
+    let mut results = vec![];
+    for h in starting_height..=latest_height {
```
aakoshh, line 196: So basically this is re-reading all the last 900 block heights? There must be a way to handle re-orgs in a more relaxed way, push notifications, maybe compare the last height read with what you have on record to see if there was a change?

fendermint/ipc/src/pof.rs
```diff
@@ -0,0 +1,85 @@
...
+            return Ok(None);
+        }
+
+        let confident_height = latest_height - self.config.chain_head_delay;
+        self.finality_proposal_at_height(confident_height)
```
aakoshh, line 53: It looks like we never access the data in anything above `latest_height - chain_head_delay`, so I wonder why the syncer even fetches those? It could just ask the parent chain always about the level which it thinks is final, and save itself the trouble of querying the IPC Agent at every poll interval for the last 900 or heights, only to look at the 900th and nothing else.

fendermint/ipc/src/pof.rs
```diff
@@ -0,0 +1,85 @@
...
+            top_down_msgs: vec![],
+            config: Default::default(),
+        }))
+    }
+
+    pub fn check_finality(&self, other_finality: &IPCParentFinality) -> bool {
+        let this_finality = match self.finality_proposal_at_height(other_finality.height) {
+            Ok(Some(finality)) => finality,
+            _ => {
+                tracing::info!("cannot create next finality, check return false");
```
aakoshh, line 78:
```suggestion
                tracing::debug!("cannot create next finality, check return false");
```

I doubt you will want to see this often in the production log.

fendermint/ipc/src/pof.rs
```diff
@@ -0,0 +1,85 @@
...
+        Ok(Some(IPCParentFinality {
+            height,
+            block_hash,
+            top_down_msgs: vec![],
+            config: Default::default(),
+        }))
+    }
+
+    pub fn check_finality(&self, other_finality: &IPCParentFinality) -> bool {
+        let this_finality = match self.finality_proposal_at_height(other_finality.height) {
```
aakoshh, line 75: I may be wrong but I think this isn't entirely what we wanted. 

As I understand `check_finality` is used to ascertain that a top-down checkpoint proposed by some validator is indeed final according to our view. (It would be great if the method contained docstrings). 

When I look at `next_finality_proposal` it contains the logic to deduct from the latest height the delay, and then it simply calls `finality_proposal_at_height`, which itself just constructs the data, but doesn't check whether the height it's asked about is indeed final. 

By calling `finality_proposal_at_height` in `check_finality`, we only check the contents of the proposal, but not its actual finality, ie. not that it's been sufficiently embedded.

fendermint/ipc/src/pof.rs
```diff
@@ -0,0 +1,85 @@
...
+            Ok(Some(finality)) => finality,
```
aakoshh, line 76:
```suggestion
            Ok(Some(finality)) => finality == *other_finality,
```
And then you don't have to use `return`.

fendermint/ipc/src/parent.rs
```diff
@@ -0,0 +1,202 @@
...
+) -> anyhow::Result<ValidatorSet> {
+    Ok(ValidatorSet::default())
+}
+
+async fn fetch_block_hashes(
+    config: &Config,
+    agent_proxy: &Arc<AgentProxy>,
+    lock: &LockedCache,
+) -> anyhow::Result<Vec<(BlockHeight, Vec<u8>)>> {
+    let latest_height = match agent_proxy.get_chain_head_height().await {
```
aakoshh, line 164: Would it be technically possible for `get_chain_head_height` to return a _lower_ height than what came before? 

Say the parent was a longest chain and it was undergoing a re-org, that is, it was switching from a shorter to a longer fork, and first it rolled back the shorter fork to the greatest common ancestor, then forward on the new new fork. If we queried during rollback, it might appear to go backwards, and then it would leave some invalid block hashes in the cache. 

For example if we had this:
```
B0 - B1 - B2 - B3 - B4
   \ 
     B1' - B2' - B3' - B4' - B5'
```
1. Our client is at B4, latest height is 4
2. The parent decides to roll back to B0 and then forward to B5'
3. We query it during the rollback when it's at B3', we fetch B1', B2', B3'
4. We insert these into the cache, which now contains B0, B1', B2', B3', B4

B4 is not a descendant of B3' but it was left there because it's not included in the catch.

If you included the clearing out of the cache above the the new minimum, rather than just do inserts, this problem went away.

fendermint/ipc/src/cache.rs
aakoshh: TBH even a `VecDequeue` would be better than a `HashMap` as most likely you want your data to be contiguous as well, without inserting into gaps, and you only pop from the beginning and push/pop from the end. You can calculate the index based on the boundaries and the offset.

fendermint/vm/message/Cargo.toml
```diff
@@ -30,6 +30,7 @@ rand = { workspace = true, optional = true }
 fendermint_vm_encoding = { path = "../encoding" }
 fendermint_vm_actor_interface = { path = "../actor_interface" }
 fendermint_testing = { path = "../../testing", optional = true }
+fendermint_ipc = { path = "../../ipc" }
```
aakoshh, line 33: Please don't add a reference to something as "heavy" as the IPC crate. The `fendermint_vm_message` create is a dependency for `fendermint_rpc`, so it should be a thin DTO layer concerned only with messages and their basic validation (like signatures), and not drag in the entire tokio stack.

aakoshh (commented):
@cryptoAtwill this is a great effort at tackling the top-down story, which I probably big enough to be called an epic. There is clearly a discovery process going on here while we figure out the best way to add this to Fendermint. 

I think it would be a good idea to make a conscious effort to greatly limit the scope of these PRs. Let's think how we could break up an issue if it seems too big, and discuss early and often. 

For example a good first PR for this would have been just one that contains just the top-down finality message proposal and nothing else. Then we can discuss if we all agree that it's in the right place and has the right content. Next we can add the parent view, without hooking it into the processing, again so we can agree about its implementation, add tests, etc.

fendermint/ipc/src/cache.rs
```diff
@@ -0,0 +1,104 @@
...
+
+impl<Key: PrimInt + Hash, Value> RangeKeyCache<Key, Value> {
+    pub fn new() -> Self {
+        Self {
+            data: Default::default(),
+            bounds: None,
+        }
+    }
+
+    pub fn upper_bound(&self) -> Option<Key> {
```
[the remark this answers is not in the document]
aakoshh: I saw later that if you started with an empty cache, inserted something, then removed it, this would still remember what the upper bound used to be. If this is an intended behaviour, it would be great to capture it as an invariant, although personally I'm not sure how intuitive it is that the cache knows about a height which you cannot retrieve from it.

fendermint/ipc/src/parent.rs
```diff
@@ -0,0 +1,202 @@
...
+            // FIXME: avoid clone here, return references, let caller clone on demand
+            v.into_iter().cloned().collect()
+        })
+    }
+
+    fn membership(&self) -> Option<ValidatorSet> {
+        self.read_cache(|cache| cache.membership.clone())
+    }
+
+    fn on_finality_committed(&self, finality: &IPCParentFinality) {
```
aakoshh, line 65: So here we remove all entries from the cache up to the `finality` height, which is calculated elsewhere as `latest - delay`. Yet the parent fetcher will fill data in each loop between `latest - lower_bound` and `latest`. Doesn't that mean that there is, or should be no difference between `delay` and `lower_bound`, otherwise one one deletes, the other puts back. Or actually it can't put it back because of the cache cutoff, but it still keeps fetching it.

aakoshh: This dissonance would disappear if you only kept one value, the finality-delay, and in the fetcher you:
1. probed the last height you fetched before
2. if it's the same as before, fetch the next region up to the latest height
3. if it's not the same, find the greatest common ancestor (walk backwards either linearly, or skipping), and then re-fetch up to the latest height, and replace

That way you only fetch what you need, and never anything beyond the finality height (if it was indeed final). Under these conditions the cache cutoffs wouldn't be required, because if the configuration is right, you should never fetch lower than lowest data.

aakoshh: Similarly upon startup, you can initialize the component with the last known finalized block, then check if it's still the same on the parent: if it doesn't exist, the parent is syncing; if it's different, it's a fatal error because the subnet state diverged from the parent. Then, go fetch until the latest height is reached, but be careful that we only start doing this once our own node is caught up with the network, or you might be fetching the entire parent chain (if our own node is going from genesis). 

Once you have fetched all the region between the last known finalized block and the latest, then you can use the finality-delay to choose the next proposal.

aakoshh:
I would propose the following breakdown of PRs:
1. Top-down IPC message
2. An STM enabled component to accumulate the data coming from the parent, one which captures the invariants of what of possible transitions (atomic reorgs) in tests, and exposes the relevant queries (and nothing more). 
3. A parent syncer that feeds this STM component with data, including the initialization requirements (never starting from an empty state, at least we should be aware of genesis), with hook for starting once self-chain-sync finished
4. The smart contract that keeps the data, including the hooks in the `actor_interface` to query it.
5. Hook it into the interpreters, disabled by default because there are no parents in end-to-end tests at the moment, and a Fendermint root should be possible as well.
6. How do we add this to the end-to-end tests? Maybe do Fendermint-only setups in the current CI? 

Open to any other recommendations.

fendermint/ipc/src/message.rs
aakoshh: I added the equivalent of this enum in the crate I mentioned above in https://github.com/consensus-shipyard/fendermint/pull/187 with a placeholder for `TopDown`.

fendermint/ipc/src/pof.rs
cryptoAtwill: I renamed to `validator_set`, should be better.

fendermint/ipc/src/parent.rs
cryptoAtwill: Good point raise, will take note of this one.

fendermint/ipc/src/pof.rs
cryptoAtwill: It is possible to have a purely EVM subnet, just be general here.

fendermint/ipc/src/pof.rs
cryptoAtwill: It's for preparing the data in advance. This could be queried when `process_proposal`, someone else sends a height that is greater than `latest_height - chain_head_delay`, we still need to fetch in advance to get the data so that we can validate.

fendermint/ipc/src/pof.rs
cryptoAtwill: `but not its actual finality`, you mean the height is not the current latest height?

fendermint/ipc/src/message.rs
```diff
@@ -0,0 +1,19 @@
...
+// SPDX-License-Identifier: Apache-2.0, MIT
+
+//! IPC messages
+
+use crate::pof::IPCParentFinality;
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+pub enum IPCMessage {
+    TopDown(IPCParentFinality),
```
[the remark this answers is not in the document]
cryptoAtwill: I think using Cid to replace all the top down messages will be a speed boost. But the issue now, like you mentioned, if someone does not have this `Cid` stored, then there should be a way to retrieve it from other peers. Currently this mechanism is missing. Or I misunderstood.

PR #185, opened 2023-08-11
FM-184: Proposal interpreter

### LINKED ISSUE #184: Proposal interpreters
Create an interpreter abstraction in the spirit of the existing ones to handle proposals. They shouldn't do anything yet, but be there for future specialization.

aakoshh:
Closes #184 

The PR introduces a `ProposalInterpreter` similar to the ones which already exist, so that we can keep the `app.rs` free of much of our application logic, and farm it out into the layers where the different types and representations of messages are handled. 

Currently it does nothing new, accepts everything. The goal was to introduce this necessary boilerplate and agree on the abstractions.

My motivation was to help https://github.com/consensus-shipyard/fendermint/pull/181 focus on the top-down message without having to decode messages in the application root.

### DIFF
--- a/fendermint/abci/src/application.rs
+++ b/fendermint/abci/src/application.rs
@@ -69,7 +69,7 @@ pub trait Application {
         Ok(response::PrepareProposal { txs })
     }
 
-    /// Opporunity for the application to inspect the proposal before voting on it.
+    /// Opportunity for the application to inspect the proposal before voting on it.
     ///
     /// The application should accept the proposal unless there's something wrong with it.
     ///
--- a/fendermint/abci/src/application.rs
+++ b/fendermint/abci/src/application.rs
@@ -11,6 +11,8 @@ use tendermint::abci::{request, response, Request, Response};
 use tower::Service;
 use tower_abci::BoxError;
 
+use crate::util::take_until_max_size;
+
 /// Allow returning a result from the methods, so the [`Application`]
 /// implementation doesn't have to be full of `.expect("failed...")`
 /// or `.unwrap()` calls. It is still good practice to use for example
@@ -62,16 +64,8 @@ pub trait Application {
         &self,
         request: request::PrepareProposal,
     ) -> AbciResult<response::PrepareProposal> {
-        let max_tx_bytes: usize = request.max_tx_bytes.try_into().unwrap();
-        let mut size: usize = 0;
-        let mut txs = Vec::new();
-        for tx in request.txs {
-            if size.saturating_add(tx.len()) > max_tx_bytes {
-                break;
-            }
-            size += tx.len();
-            txs.push(tx);
-        }
+        let txs = take_until_max_size(request.txs, request.max_tx_bytes.try_into().unwrap());
+
         Ok(response::PrepareProposal { txs })
     }
 
--- a/fendermint/app/src/app.rs
+++ b/fendermint/app/src/app.rs
@@ -7,6 +7,7 @@ use std::sync::{Arc, Mutex};
 use anyhow::{anyhow, Context, Result};
 use async_trait::async_trait;
 use cid::Cid;
+use fendermint_abci::util::take_until_max_size;
 use fendermint_abci::{AbciResult, Application};
 use fendermint_storage::{
     Codec, Encode, KVCollection, KVRead, KVReadable, KVStore, KVWritable, KVWrite,
@@ -492,6 +493,45 @@ where
         };
 
         Ok(response)
+    }
+
+    /// Amend which transactions to put into the next block proposal.
+    async fn prepare_proposal(
+        &self,
+        request: request::PrepareProposal,
+    ) -> AbciResult<response::PrepareProposal> {
+        let txs = request.txs.into_iter().map(|tx| tx.to_vec()).collect();
+
+        let txs = self
+            .interpreter
+            .prepare((), txs)
+            .await
+            .context("failed to prepare proposal")?;
+
+        let txs = txs.into_iter().map(bytes::Bytes::from).collect();
+        let txs = take_until_max_size(txs, request.max_tx_bytes.try_into().unwrap());
+
+        Ok(response::PrepareProposal { txs })
+    }
+
+    /// Inspect a proposal and decide whether to vote on it.
+    async fn process_proposal(
+        &self,
+        request: request::ProcessProposal,
+    ) -> AbciResult<response::ProcessProposal> {
+        let txs = request.txs.into_iter().map(|tx| tx.to_vec()).collect();
+
+        let accept = self
+            .interpreter
+            .process((), txs)
+            .await
+            .context("failed to process proposal")?;
+
+        if accept {
+            Ok(response::ProcessProposal::Accept)
+        } else {
+            Ok(response::ProcessProposal::Reject)
+        }
     }
 
     /// Signals the beginning of a new block, prior to any `DeliverTx` calls.
--- a/fendermint/app/src/app.rs
+++ b/fendermint/app/src/app.rs
@@ -22,7 +22,7 @@ use fendermint_vm_interpreter::fvm::state::{
 use fendermint_vm_interpreter::fvm::{FvmApplyRet, FvmGenesisOutput};
 use fendermint_vm_interpreter::signed::InvalidSignature;
 use fendermint_vm_interpreter::{
-    CheckInterpreter, ExecInterpreter, GenesisInterpreter, QueryInterpreter,
+    CheckInterpreter, ExecInterpreter, GenesisInterpreter, ProposalInterpreter, QueryInterpreter,
 };
 use fvm::engine::MultiEngine;
 use fvm_ipld_blockstore::Blockstore;
@@ -309,6 +309,15 @@ where
     S::Namespace: Sync + Send,
     DB: KVWritable<S> + KVReadable<S> + Clone + Send + Sync + 'static,
     SS: Blockstore + Clone + Send + Sync + 'static,
+    I: GenesisInterpreter<
+        State = FvmGenesisState<SS>,
+        Genesis = Vec<u8>,
+        Output = FvmGenesisOutput,
+    >,
+    I: ProposalInterpreter<
+        State = (), // TODO
+        Message = Vec<u8>,
+    >,
     I: ExecInterpreter<
         State = FvmExecState<SS>,
         Message = Vec<u8>,
@@ -325,11 +334,6 @@ where
         State = FvmQueryState<SS>,
         Query = BytesMessageQuery,
         Output = BytesMessageQueryRet,
-    >,
-    I: GenesisInterpreter<
-        State = FvmGenesisState<SS>,
-        Genesis = Vec<u8>,
-        Output = FvmGenesisOutput,
     >,
 {
     /// Provide information about the ABCI application.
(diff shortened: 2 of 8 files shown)


aakoshh:
# Notes 

https://github.com/tendermint/tendermint/blob/v0.37.0-rc2/spec/abci/abci++_app_requirements.md say that `ProcessProposal` needs to be deterministic and depend only on the state, which seem at odds with what we are trying to do with it: 
* only vote on block if the parent state is final according to the view of individual validators
* only vote on block if the checkpoint CID is available to the validator casting the vote

The formal requirements state that the vote should only depend on the block and the previous state, which is obviously not the case for us. Obviously we are aware of this, which is why we discussed that timing is critical, and we should consider:
* using the vote gossiping feature of the _IPLD Resolver_ to only propose top-down checkpoints after a quorum of votes have been collected, at which point we can say that we don't even need the `ProcessProposal` mechanism, but at least we should be able to use it because the honest validators will vote for the block if they have voted for the CID (although we can consider that if +2/3 voted for the CID, at least +1/3 should vote for the block, but 1/3 might be Byzantine and vote against it, while the rest of the 1/3 have not seen it as final)
* always use a _for resolution_ transaction transaction first to make sure that every validator is aware of what they should try to resolve - technically it is in the state then, even though it's availability isn't. We could add their votes to the state too, but the data would still not be in the ledger, only pledged to be available. Again we'd only know that +1/3 of the nodes should be willing to serve the data.

So while it's not deterministic, we have good mitigation strategies to minimise the chance of stalling consensus.

fendermint/app/src/app.rs
```diff
@@ -490,6 +495,45 @@ where
...
+    /// Amend which transactions to put into the next block proposal.
+    async fn prepare_proposal(
+        &self,
+        request: request::PrepareProposal,
+    ) -> AbciResult<response::PrepareProposal> {
+        let txs = request.txs.into_iter().map(|tx| tx.to_vec()).collect();
+
+        let txs = self
+            .interpreter
+            .prepare((), txs)
```
cryptoAtwill, line 507: Why do we need to pass txs as a parameter to `prepare`? I had the impression that we are creating extra transactions but not modifying?

fendermint/vm/interpreter/src/bytes.rs
```diff
@@ -31,6 +31,73 @@ impl<I> BytesMessageInterpreter<I> {
...
+                    tracing::warn!(
+                        error = e.to_string(),
+                        "failed to decode message in mempool as ChainMessage"
+                    );
+                }
+                Ok(msg) => chain_msgs.push(msg),
+            }
+        }
+
+        let chain_msgs = self.inner.prepare(state, chain_msgs).await?;
```
cryptoAtwill, line 68: Not sure for bottom up, but for top down, it's not needed to deserialize the bytes to get `ChainMessage` as no inspection needed. Or it's likely someone will send a `IPCMessage` instead of a `SignedMessage` and we have to inspect here?

fendermint/vm/interpreter/src/bytes.rs
```diff
@@ -31,6 +31,73 @@ impl<I> BytesMessageInterpreter<I> {
...
+    async fn process(&self, state: Self::State, msgs: Vec<Self::Message>) -> anyhow::Result<bool> {
+        let mut chain_msgs = Vec::new();
+
+        for msg in msgs {
+            match fvm_ipld_encoding::from_slice::<ChainMessage>(&msg) {
+                Err(e) => {
+                    // This would indicate a Byzantine validator which includes rubbish in their proposal.
+                    // We could reject the proposal here, or we can accept it and punish the validator during
+                    // block execution, so that their power is reduced.
+                    tracing::debug!(
```
cryptoAtwill, line 88: should we reject the proposal here is we cannot deserialize the txn?

fendermint/vm/interpreter/src/chain.rs
```diff
@@ -34,6 +34,40 @@ impl<I> ChainMessageInterpreter<I> {
...
+        &self,
+        _state: Self::State,
+        msgs: Vec<Self::Message>,
+    ) -> anyhow::Result<Vec<Self::Message>> {
+        // For now this is just a placeholder.
+        Ok(msgs)
+    }
+
+    /// Perform finality checks on top-down transactions and availability checks on bottom-up transactions.
+    async fn process(
```
cryptoAtwill, line 61: maybe we can call it `validate` or something else along the line as process I feel it's consuming or processing the message, but we are actually inspecting.

aakoshh: Yeah I know what you mean, I based all names in all the interfaces on what they are called in ABCI but without their prefix they are a bit confusing, like `begin` and `end`, or this `process` - they make sense in the context of their trait, but when the interpreter implements all interfaces, they are not very good. 

Should we just use the full names, like `begin_block` and `process_proposal`?

aakoshh: `validate` would be similar to `process` in that it would be clashing with `check`, which lives in the `CheckInterpreter`.

aakoshh: Created https://github.com/consensus-shipyard/fendermint/issues/207

fendermint/vm/interpreter/src/bytes.rs
aakoshh: Added the flag in https://github.com/consensus-shipyard/fendermint/pull/185/commits/a36603e957a58853dbe10af1053b309e8d92664a

fendermint/vm/interpreter/src/bytes.rs
aakoshh: Added a mode in https://github.com/consensus-shipyard/fendermint/pull/185/commits/22e3719c92bb05bc5598abfb2294155909160324

cryptoAtwill (approved):

fendermint/abci/src/application.rs
```diff
@@ -62,20 +64,12 @@ pub trait Application {
...
-        let mut size: usize = 0;
-        let mut txs = Vec::new();
-        for tx in request.txs {
-            if size.saturating_add(tx.len()) > max_tx_bytes {
-                break;
-            }
-            size += tx.len();
-            txs.push(tx);
-        }
+        let txs = take_until_max_size(request.txs, request.max_tx_bytes.try_into().unwrap());
```
adlrocha, line 67: We only check here if we reach the size limit of the Tendermint block, right? Should we add also here a `TODO` to remember not to exhaust the gas limit of a block?

fendermint/app/src/app.rs
```diff
@@ -490,6 +495,45 @@ where
...
+            .await
+            .context("failed to prepare proposal")?;
+
+        let txs = txs.into_iter().map(bytes::Bytes::from).collect();
+        let txs = take_until_max_size(txs, request.max_tx_bytes.try_into().unwrap());
```
adlrocha, line 512: Same here for the gas limit check or should we only check as part of `process_proposal`?

fendermint/vm/interpreter/src/bytes.rs
```diff
@@ -19,15 +19,124 @@ pub type BytesMessageQueryRet = Result<FvmQueryRet, fvm_ipld_encoding::Error>;
 /// Close to what the ABCI sends: (Path, Bytes).
 pub type BytesMessageQuery = (String, Vec<u8>);
 
+/// Behavour of proposal preparation. It's an otimisation to cut down needless serialization
```
adlrocha, line 22:
```suggestion
/// Behavour of proposal preparation. It's an optimisation to cut down needless serialization
```

fendermint/app/src/cmd/run.rs
```diff
@@ -27,7 +29,8 @@ async fn run(settings: Settings) -> anyhow::Result<()> {
     );
     let interpreter = SignedMessageInterpreter::new(interpreter);
     let interpreter = ChainMessageInterpreter::new(interpreter);
-    let interpreter = BytesMessageInterpreter::new(interpreter);
+    let interpreter =
+        BytesMessageInterpreter::new(interpreter, ProposalPrepareMode::AppendOnly, false);
```
adlrocha, line 33: What is the rationale for not rejecting a malformed proposal here (if any, as maybe is just a placeholder value for now)?

adlrocha (approved):
This is gold. Thank you for this @aakoshh, it also helped me understand the role of interpreters better :pray:

fendermint/vm/interpreter/src/bytes.rs
aakoshh: One more problem with rejection is that we have to be careful not to reject something we accidentally admitted into our own mempool, otherwise it will never be removed from it.

fendermint/abci/src/application.rs
aakoshh: The gas aspect is definitely ignored at the moment, thanks for the reminder. I created an issue to track it: https://github.com/consensus-shipyard/fendermint/issues/208

The check would not be here because at this size we don't know what messages we are dealing with, but once they are parsed into `ChainMessage` we can look at the gas limits.

fendermint/app/src/app.rs
aakoshh: You are right that it must be part of both _prepare_ and _process_ otherwise you risk freaking out CometBFT by rejecting your own proposal, which is a big no-no. 

Again it would not be here as this is purely based on sizes and cannot access gas. This logic is here because it is mandatory, not something application specific like the gas, which must be delegated to interpreters.

fendermint/app/src/cmd/run.rs
aakoshh: I included the [considerations](https://github.com/consensus-shipyard/fendermint/pull/185/files#diff-5eac90311bc4bcf59a295948bfe9b9c82b725532cff3ce4036d92a8654052b85R118-R126) in the code as comments where this flag is used. 

The reason I thought we might not want to reject them for now is because 
1. They should not happen as our nodes are honest. 
2. If they happen, we have a bug with a submitted transaction, and this way the error will show up in the transaction results which the RPC client will get and print, so the test will fail with an error, rather than a timeout. Whereas if we reject the proposal, CometBFT itself will most likely die as a result because our honest nodes will have proposed something that they subsequently reject themselves.  And with `AppendOnly` we are not going to inspect the bytes before adding them to the proposal, as it was pointed out as wasted effort by @cryptoAtwill . That will have to change if we want to limit gas, though. 
3. If we admitted a malformed transaction to our mempool, inspected it (by using `PassThrough` instead of `AppendOnly`), and not included it in our proposal, it will still stay in the mempool, resulting in a memory leak in CometBFT if we keep piling up such transactions. This would be another bug that is potentially hard to figure out, apart from flooding the logs with warnings by trying to propose the same transaction every 2 seconds. 

If all our nodes are not honest and some are proposing faulty transactions then this would be a stupid way to attack instead of just not proposing any block to cause a timeout. If it happened, our only chance to weed out such actors is to admit the transaction and penalize the miner. If that happened due to a bug, we'd risk consensus failure, but at least we'd see what's happening, and by coding the penalty mechanism we'd discourage anyone from intentionally doing this, and to implement these checks in `check_tx`. 

So, overall, I thought that a malformed TX here is most likely a bug that is easiest to detect by letting it through.

adlrocha: That makes sense! Assuming our implementation honest saving ourselves from nasty bugs is the way to go, thanks!

PR #198, opened 2023-08-16
FM-190: Check relayed bottom-up checkpoints

### LINKED ISSUE #190: Implement CheckInterpreter for Ipc::BottomUpResolve
Implement the `check` handler for the `Ipc::BottomUpResolve` variant of `ChainMessage`. 

It should behave similar to the `SignedMessageInterpreter`: 
1. check the signature of the relayer, to protect against DoS attacks
2. check the existence and nonce of the relayer, to protect against replay attacks

We don't need to check the quorum certificate in the relayed checkpoint itself. If it's invalid, the relayer will have to pay the price of the execution of the transaction, which should make this attack moot.

aakoshh:
Closes #190 
Depends on #187 

Introduces a `VerifiableMessage` and `SynteticMessage` which are used to expand the messages handled by the `SignedMessageInterpreter` to include one that is artificially constructed by the interpreter to make use of existing services. 

In particular, the signed relayed bottom up checkpoint is turned into an FVM message, with the relayer signature assumed to be over the CID of the relay-message itself. The signature of this synthetic message is verified by the `SignedMessageInterpreter` before it passes it into the inner `FvmMessageInterpreter` which checks the existence of the relayer account and the nonce.

The PR adds missing gas fields to the relayed message.


fendermint/vm/interpreter/src/signed.rs
```diff
@@ -17,6 +20,67 @@ pub struct InvalidSignature(pub String);
...
+    }
+}
+
+pub struct SyntheticMessage {
+    /// The artifical message.
+    message: FvmMessage,
+    /// The CID of the original message (assuming here that that's what was signed).
+    orig_cid: cid::Cid,
+    /// The signature over the original CID.
+    signature: Signature,
```
adlrocha, line 57: Just to double-check, this is the original signature of the message, right? Should we maybe call it `WrappedMessage` instead of `SyntheticMessage`? (weak opinion weakly held, but synthetic seemed like "artificial" to me :) )

fendermint/vm/interpreter/src/signed.rs
```diff
@@ -17,6 +20,67 @@ pub struct InvalidSignature(pub String);
...
+        Ok(Self {
+            message,
+            orig_cid,
+            signature,
+        })
+    }
+
+    pub fn verify(&self, chain_id: &ChainID) -> Result<(), SignedMessageError> {
+        let mut data = self.orig_cid.to_bytes();
+        data.extend(chain_id_bytes(chain_id).iter());
```
adlrocha, line 76: I thought the chainID was a prefix instead of a suffix of the message cid. Out of a curiosity, is this Fendermint-specific or Filecoin also does this?

fendermint/vm/message/src/ipc.rs
```diff
@@ -39,6 +41,10 @@ pub struct RelayedMessage<T> {
     pub relayer: Address,
     /// The nonce of the relayer in the current subnet.
     pub sequence: u64,
+    /// The gas the relayer is willing to spend on the verification of the relayed message.
```
adlrocha, line 44: Nice one!!! I completely missed it in the previous PR.

adlrocha (approved):
A few minor questions, but LGTM.

fendermint/vm/interpreter/src/signed.rs
aakoshh: It _is_ an artificial message, that's exactly what I was trying to convey, that it's just a transient technical message to piggy-back on the signature checks and FVM message execution. The construction of the FVM message from the original is synthesising an artificial message that nobody sent.

It's indeed wrapping something but that wouldn't say why; so does the other one after all. 

You are correct that `signature` is the signature over the original message, in particular the CID of the original message. 

We could get rid of this if we used FVM messages, in which case instead of synthesising a message in the `ChainMessageInterpreter` from an IPC specific one, we would parse an FVM message into an IPC message to look for CIDs to resolve.

fendermint/vm/interpreter/src/signed.rs
aakoshh: I have no idea what Filecoin does, I did this in https://github.com/consensus-shipyard/fendermint/pull/115

Ethereum seems to play with the signature itself, which seemed so complicated I didn't think we need to replicate it in Fendermint. The `ethers` library takes care of it for us, but for regular messages I do the suffix. Is prefix safer?

fendermint/vm/interpreter/src/signed.rs
adlrocha: I think this is fine for now, and now I understand the rationale behind the name. Thanks!

PR #199, opened 2023-08-16
FM-186: Fix rust-toolchain action version

jsoares:
@aakoshh Apologies but I missed this while reviewing #188. When specifying the `toolchain`, we should use the `master` branch of the action rather than `stable` (per https://github.com/dtolnay/rust-toolchain#inputs). I don't think it breaks anything right now but better avoid unsupported combinations.

### DIFF
--- a/.github/actions/install-tools/action.yaml
+++ b/.github/actions/install-tools/action.yaml
@@ -13,7 +13,7 @@ runs:
 
   steps:
     - name: Install Rust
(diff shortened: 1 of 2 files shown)


aakoshh:
@jsoares good catch, thanks! I was rather confused about how to interpret the tag in the `uses`, and missed this comment.

aakoshh (approved):
Awesome, thank you!

PR #201, opened 2023-08-17
FM-191: Execute BottomUpResolve

### LINKED ISSUE #191: Implement ExecInterpreter for Ipc::BottomUpResolve
Implement the `deliver` handler for the `Ipc::BottomUpResolve` variant of `ChainMessage`.

This is similar to the handling of `SignedMessage`:
1. Check the signature of the relayer. 
2. We can transform the message into an FVM message and send it to the Gateway actor, so that it dispatches it to the right SubnetActor, which should perform the check of the checkpoint and the quorum certificate. By doing so, we leverage the FVM to i) check and increment the relayer nonce and ii) charge gas for the execution of the checkpoint validation.

See https://github.com/consensus-shipyard/fendermint/issues/192 for the validation of the checkpoint by the actor.

If the transaction is successfully executed, the interpreter adds the checkpoint to a pool for resolution.

aakoshh:
Closes #191 
Depends on #198

_The PR looks bigger than it is because it merged `main`, and because after running `cargo update` there are lots of changes to `Cargo.lock` and `ethers` also generated different code for Solidity actors. We can rebase once the dependencies are merged._

- [x] Transform the `BottomUpResolve` to a message we can send to the FVM (does not include cross-messages, just for quorum certificate validation and reward tracking)
- [x] Pass the FVM message to the inner interpreters
- [x] Create an STM data structure for tracking background resolutions
- [x] Add the cross-messages' root CID to the background resolution pool

### DIFF
--- a/fendermint/vm/interpreter/src/chain.rs
+++ b/fendermint/vm/interpreter/src/chain.rs
@@ -12,6 +12,7 @@ use fendermint_vm_message::{
     chain::ChainMessage,
     ipc::{BottomUpCheckpoint, CertifiedMessage, IpcMessage, SignedRelayedMessage},
 };
+use fendermint_vm_resolver::pool::ResolvePool;
 use fvm_ipld_encoding::RawBytes;
 use fvm_shared::econ::TokenAmount;
 use num_traits::Zero;
@@ -79,7 +80,11 @@ impl<I> ExecInterpreter for ChainMessageInterpreter<I>
 where
     I: ExecInterpreter<Message = VerifiableMessage, DeliverOutput = SignedMessageApplyRet>,
 {
-    type State = I::State;
+    // The state consists of the resolver pool, which this interpreter needs, and the rest of the
+    // state which the inner interpreter uses. This is a technical solution because the pool doesn't
+    // fit with the state we use for execution messages further down the stack, which depend on block
+    // height and are used in queries as well.
+    type State = (ResolvePool, I::State);
     type Message = ChainMessage;
     type BeginOutput = I::BeginOutput;
     type DeliverOutput = ChainMessageApplyRet;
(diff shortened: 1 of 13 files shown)


fendermint/vm/interpreter/src/chain.rs
```diff
@@ -56,28 +84,63 @@ where
...
-                // I expect for now that we don't run with untrusted validators, so it's okay to quit.
-                todo!("#191: implement execution handling for IPC")
+                Ok(((pool, state), ChainMessageApplyRet::Signed(ret)))
             }
+            ChainMessage::Ipc(msg) => match msg {
+                IpcMessage::BottomUpResolve(msg) => {
+                    let smsg = relayed_bottom_up_ckpt_to_fvm(&msg)
+                        .context("failed to syntesize FVM message")?;
+
+                    // Let the FVM validate the checkpoint quorum certificate and take not of the relayer for rewards.
```
adlrocha, line 102: typo?
```suggestion
                    // Let the FVM validate the checkpoint quorum certificate and take note of the relayer for rewards.
```

adlrocha, line 102: Let me see if I understand this. The idea is that: 
- There is a `Resolve` message committing the checkpoint which includes the `Cid` of the list of bottom-up messages to be executed. 
- If the commitment happens, the `Cid` of the messages are queue for resolution and a posterior proposal an execution?

I still think we may need to remove the resolution stage and directly include messages as part of the checkpoint in a first implementation. Message resolution may include additional complexity in the L2->root interaction :(  

In any case, I feel this will make more sense as I move towards the implementation of the actual execution.

adlrocha (approved):
The high-level idea LGTM and I like the abstractions we are using. My only concern is the same as in the previous PRs, while the protocol would work end-to-end if all peers where Fendermint-based, we may need to simplify it a bit to support mainnet interactions.

fendermint/vm/interpreter/src/chain.rs
aakoshh: Yes, you got that right, and as per our sync we will implement both behaviours depending on what kind of parent we'll be dealing with. For Fendermint<-Fendermint relationships we will use this, for Lotus<-Fendermint we won't.

PR #202, opened 2023-08-18
FM-196: Implement ProposalInterpreter::process for BottomUp

### LINKED ISSUE #196: Check data availability of checkpoints in ProposalInterpreter::process
Assuming https://github.com/consensus-shipyard/fendermint/issues/195 added the `BottomUpExec` message to the block proposal. 

Implement `ProposalInterpreter::process` by checking if the data is marked available in the pool. Don't have to check the blockstore, because https://github.com/consensus-shipyard/fendermint/issues/193 should have already loaded the pending ones into the pool, where the background resolver should mark them as available if they are.

We also don't have to validate the checkpoint here. That can be done during the execution: if it's invalid, penalize the validator who proposed it.

aakoshh:
Closes #196 
Depends on #201 and #185

Implements `ProposalInterpreter::process` method for `ChainMessageInterpreter` by looking up proposed checkpoints for execution in the `ResolvePool` and checking that the item is present and resolved, otherwise we don't vote for the proposal.

### DIFF
--- a/fendermint/app/src/app.rs
+++ b/fendermint/app/src/app.rs
@@ -330,10 +330,7 @@ where
         Genesis = Vec<u8>,
         Output = FvmGenesisOutput,
     >,
-    I: ProposalInterpreter<
-        State = (), // TODO
-        Message = Vec<u8>,
-    >,
+    I: ProposalInterpreter<State = CheckpointPool, Message = Vec<u8>>,
     I: ExecInterpreter<
         State = (CheckpointPool, FvmExecState<SS>),
         Message = Vec<u8>,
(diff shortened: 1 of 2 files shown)


adlrocha (approved):
LGTM. One quick question (and I think I saw a message for this). The checkpoint pool will be garbage collected after a checkpoint execution, right? At this point we know it won't have to be re-proposed anymore.

aakoshh:
> The checkpoint pool will be garbage collected after a checkpoint execution, right? At this point we know it won't have to be re-proposed anymore.

That's right, when the checkpoint is executed it has to be removed from the resolver pool. Thanks for mentioning this because I have not included this in any of the issue description, just commented on it in the code. I created https://github.com/consensus-shipyard/fendermint/issues/209 to track it now.

PR #206, opened 2023-08-21
Top Down Checkpoint - Parent Sync

cryptoAtwill:
Constantly syncs with the parent using a polling mechanism. The ipc-agent pending integration.


fendermint/vm/topdown/src/sync.rs
```diff
@@ -0,0 +1,147 @@
...
+use ipc_sdk::ValidatorSet;
+use std::cmp::max;
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::Arc;
+use std::time::Duration;
+
+/// Constantly syncing with parent through polling
+pub struct PollingParentSyncer<T> {
+    config: Config,
+    started: Arc<AtomicBool>,
```
aakoshh, line 18: Why does this have to be in an `Arc`? The struct itself is not even `Clone`, so who would it share it with?

fendermint/vm/topdown/src/sync.rs
```diff
@@ -0,0 +1,147 @@
...
+impl<T: ParentFinalityProvider + Send + Sync + 'static> PollingParentSyncer<T> {
+    /// Start the proof of finality listener in the background
+    pub fn start(&mut self) -> anyhow::Result<()> {
+        match self
+            .started
+            .compare_exchange(false, true, Ordering::SeqCst, Ordering::SeqCst)
+        {
+            Ok(_) => {}
+            Err(_) => return Err(anyhow!("already started")),
+        }
```
aakoshh, line 43:
```suggestion
    pub fn start(self) {    
```

By consuming `self` this check can be completely avoided. You set up the component, then run it till the end of the program, optionally returning some handle to stop it. We communicate with the rest of the world through the parent view anyway, so there doesn't seem to be a need to call this once it's started.

fendermint/vm/topdown/src/sync.rs
```diff
@@ -0,0 +1,147 @@
...
+        let starting_height = if let Some(h) = provider.latest_height().await {
+            h + 1
+        } else {
+            // if latest_recorded_height is None, it means we have not started fetching any heights
+            // we just use the latest height minus a lower bound as the parent view
+            max(
+                1,
+                latest_height.saturating_sub(config.chain_head_lower_bound),
+            )
+        };
```
aakoshh, line 93: This still looks wrong: the parent view even contains a "last committed finality" component, so that's where we should be resuming from at the beginning. And this is an example of where returning `StmResult` would be helpful: you can inspect `latest_height` and if it's `None` then also look at `last_committed_finality` in the same block. 

By only considering `latest_height`, which only depends on what you fetched since `start`, if I shut down my node longer than `chain_head_lower_bound` I will have a gap. 

And as I said last week, this should be combined with a check to whether CometBFT is syncing with its own chain, because while we are syncing we should not be asking the agent for historical data, since we'll most likely get it from our peers. If that's the reason you only consider the last N blocks, it's not clear from the comments.

But still, if all nodes were down, or I'm an only-node subnet for whatever reason, then who is going to fill in the gap?

fendermint/vm/topdown/src/sync.rs
```diff
@@ -0,0 +1,147 @@
...
+                .get_block_hash(h)
+                .await
+                .context("cannot fetch block hash")?;
+            let validator_set = agent_proxy
+                .get_validator_set(h)
+                .await
+                .context("cannot fetch validator set")?;
+
+            match provider
+                .new_parent_view(Some((h, block_hash, validator_set)), vec![])
```
aakoshh, line 106: It would be helpful to add a comment as to why the top-down messages are empty.

fendermint/vm/topdown/src/sync.rs
```diff
@@ -0,0 +1,147 @@
...
+                .await
+            {
+                Ok(_) => {}
+                Err(e) => match e {
+                    Error::ParentReorgDetected(_) => {
```
aakoshh, line 111: I don't fully grasp how this error will come about. Say the `latest_height` is `N`, which means you will try to fetch from `N+1`. From the previous PR I [only see](https://github.com/consensus-shipyard/fendermint/pull/205/files#diff-8edd9923a9eb262950fd919a858ece4449d8a41abf401bd3c1e4ac1d63190454R88) that any error inserting into the height based cache is considered a reorg, which is if the height you are trying to insert isn't what's expected, ie. lower or higher than `N+1`. 

But since you are fetching from `N+1`, you won't see `N` or `N+2`. The only way to detect a reorg would be to make the cache more like a chain and check that the parent hash of the newly inserted head isn't what you had last, however this wouldn't even be possible because the cache doesn't store each height, only ones divisible by the "checkpointing" interval.

fendermint/vm/topdown/src/sync.rs
```diff
@@ -0,0 +1,147 @@
...
+            .await
+            .context("cannot fetch parent chain head")?;
+
+        if latest_height < config.chain_head_delay {
+            tracing::debug!("latest height not more than the chain head delay");
+            continue;
+        }
+
+        let starting_height = if let Some(h) = provider.latest_height().await {
+            h + 1
```
aakoshh, line 85: Shouldn't `+ 1` instead be the same increment as the value configured for the storage cache? Otherwise it will just return an error with unexpected height.

fendermint/vm/topdown/src/sync.rs
```diff
@@ -0,0 +1,147 @@
...
+                1,
+                latest_height.saturating_sub(config.chain_head_lower_bound),
+            )
+        };
+
+        for h in starting_height..=latest_height {
+            let block_hash = agent_proxy
+                .get_block_hash(h)
+                .await
+                .context("cannot fetch block hash")?;
```
aakoshh, line 99: What if there is no block at that height yet?

fendermint/vm/topdown/src/sync.rs
```diff
@@ -0,0 +1,147 @@
...
+                        todo!()
+                    }
+                    _ => unreachable!(),
+                },
+            }
+        }
+
+        // now get the top down messages
+        let nonce = provider.latest_nonce().await.unwrap_or(0);
+        let top_down_msgs = agent_proxy.get_top_down_msgs(latest_height, nonce).await?;
```
aakoshh, line 123: I see that this only depends on [what has been fetched](https://github.com/consensus-shipyard/fendermint/pull/205/files#diff-8edd9923a9eb262950fd919a858ece4449d8a41abf401bd3c1e4ac1d63190454R63-R64) in this session, not what has been committed before. 

Does that mean that after every restart we'll ask the agent to fetch all messages since genesis, with nonce = 0? I don't see how this could be what we want, how could all cross-messages fit into memory?

aakoshh (commented):
I have similar comments about restart behaviour as last week, and I again feel like there's something missing about how cross message nonces are not related to block heights. 

In my mind the nonces are committed together with all the other changes, and one should be talking about cross messages up to and including the nonce in the last finalized parent block. There would be no need to fetch beyond that, and no need to handle them as a separate thing to fetch from the heights: give me the cross messages at each height for which I'm fetching the block hash or the validator set.

cryptoAtwill:
@aakoshh The current way of using `nonce` is because the solidity contract is using `nonce`, the contract does not record the messages submitted at certain height. If tracking using height is easier, we have to update the contract first, cc @adlrocha.

fendermint/vm/topdown/src/sync.rs
cryptoAtwill: I will remove the `started` and in favor of taking `self` as parameter to `start`

fendermint/vm/topdown/src/sync.rs
```diff
@@ -0,0 +1,151 @@
...
+            config,
+            started: Arc::new(AtomicBool::new(false)),
+            parent_view_provider,
+            agent,
+        }
+    }
+}
+
+impl<T: ParentFinalityProvider + Send + Sync + 'static> PollingParentSyncer<T> {
+    /// Start the proof of finality listener in the background
```
adlrocha, line 35: Is this a typo? 
```suggestion
    /// Start the parent finality listener in the background
```

fendermint/vm/topdown/src/sync.rs
adlrocha: ^+1

fendermint/vm/topdown/src/sync.rs
```diff
@@ -0,0 +1,151 @@
...
+        let provider = self.parent_view_provider.clone();
+        let agent = self.agent.clone();
+
+        tokio::spawn(async move {
+            loop {
+                if let Err(e) = sync_with_parent(&config, &agent, &provider).await {
+                    tracing::info!("sync with parent encountered error: {e}");
+                }
+
+                // TODO: add termination channel?
```
adlrocha, line 55: I think it is useful to expose some way to stop the process and make it up to the user to use it or not.

fendermint/vm/topdown/src/sync.rs
adlrocha: ^+1

adlrocha (commented):
I think Akosh's reviews mostly covers all my comments, let me know if you'd like to discuss further sync

fendermint/vm/topdown/src/sync.rs
cryptoAtwill: right, it should be fetching from the last committed finality, similar to the block height. Not sure why I placed 0.

fendermint/vm/topdown/src/sync.rs
cryptoAtwill: it would raise an error and retry.

fendermint/vm/topdown/src/sync.rs
cryptoAtwill: Based on the way we are fetching, `N+2` should not have happened. Come to think of it, a reorg should already be detected if `starting_height > latest_height`.

fendermint/vm/topdown/src/sync.rs
cryptoAtwill: I have splited the api into two separate calls for now. If we have solidity contract updated to track top down msgs by height, then we can merge the two apis again.

fendermint/vm/topdown/src/sync.rs
cryptoAtwill: I currently removed this `todo` as I'm not so sure how/when should be stop this process. I assume this should run as long as `fendermint` is up? I will add the necessary code when it starts to get clearer.

cryptoAtwill:
@aakoshh @adlrocha After querying top down messages with block height, most of the comments should resolve themselves. 

The part missing is when block height goes backwards, it is a potential chain reorg or node error. Anyways, I put what I thought could be an initial implementation. Please help take a look and let me know what you think.

adlrocha:
> The part missing is when block height goes backwards, it is a potential chain reorg or node error. Anyways, I put what I thought could be an initial implementation. Please help take a look and let me know what you think.

I will have a look right now, but just a note. In order to initially simplify the implementation, we could just set the delay of the finality proposed to a number of epochs where the probability or re-orgs is really low. IIRC, @guy-goren was double-checking that the practical value may be around 100 epochs (which is still a really long time, and we should definitely make it better). This is just in case reorgs complicate our lives a lot at this point.

fendermint/vm/topdown/src/sync.rs
```diff
@@ -0,0 +1,205 @@
...
+    let r = atomically_or_err(|| {
+        Ok(if let Some(h) = provider.latest_height()? {
+            h + 1
+        } else {
+            let last_committed_finality = provider.last_committed_finality()?;
+            last_committed_finality.height + 1
+        })
+    })
+    .await;
+    let starting_height = downcast_err!(r)?;
```
adlrocha, line 134: How is the starting height retrieved? IIRC from the previous PR the `ParentFinalityProvider` doesn't fallback on-chain to see what was the latest height committed. If this is the case, what happens if the node is syncing from scratch a the finality for height `n >> 0` is already committed. Will this start fetching blocks from `0` to the latest committed?

I think we should maybe include the fallback here

fendermint/vm/topdown/src/sync.rs
```diff
@@ -0,0 +1,205 @@
...
+        interval.tick().await;
+
+        let (starting_height, latest_height) =
+            get_query_block_heights(agent_proxy, provider).await?;
+        if latest_height < config.chain_head_delay {
+            tracing::debug!("latest height not more than the chain head delay");
+            continue;
+        }
+
+        // we are going backwards in terms of block height, the latest block height is lower
```
adlrocha, line 90: While this makes sense from the syncing/caching perspective, what may be a bit more complicated is how to handle all the on-chain side-effects that may have happened before the reorg. Finality commitments and cross-net messages may need to be reverted. This may be hard to handle and not that obvious :(

fendermint/vm/topdown/src/sync.rs
```diff
@@ -0,0 +1,205 @@
...
+            Ok(v) => Ok(v),
+            Err(e) => match e.downcast_ref::<Error>() {
+                None => unreachable!(),
+                Some(e) => Err(e.clone()),
+            },
+        }
+    };
+}
+
+async fn sync_with_parent<T: ParentFinalityProvider + Send + Sync + 'static>(
```
adlrocha, line 71: Would you mind adding a comment on what is the end-to-end process so it is clear:
1. Get the heights to be populated in the parent view. 
2. Populate the finality information for each of these heights.

adlrocha (commented):
We are getting closer. My main concern now is with nodes syncing from scratch when there's some finality already committed. We should avoid syncing from zero and query on-chain what is the latest finality committed.

fendermint/vm/topdown/src/sync.rs
cryptoAtwill: As discussed over slack, when it first start, it will fetch from the chain, but the storage on chain has yet to be implemented, so I left the part on fetching from storage unimplemented.

fendermint/vm/topdown/src/finality.rs
```diff
@@ -0,0 +1,476 @@
...
+        top_down_msgs: Vec<CrossMsg>,
+    ) -> StmResult<(), Error> {
+        if !top_down_msgs.is_empty() {
+            // make sure incoming top down messages are ordered by nonce sequentially
+            ensure_sequential_by_nonce(&top_down_msgs)?;
+        };
+
+        let r = self.parent_view_data.height_data.modify(|mut cache| {
+            let r = cache
+                .append(height, (block_hash, validator_set, top_down_msgs))
```
aakoshh, line 98: Based on what I have seen so far of the cache, this will fail unless `height` follows the previous by `increment`, which we know is 1 in the syncer, but could be anything here.

Whose responsibility is it to aggregate into the expected intervals? This or the caller?

aakoshh: Oh I see it is indeed hardcoded for single blocks [here](https://github.com/consensus-shipyard/fendermint/pull/206/files#diff-8edd9923a9eb262950fd919a858ece4449d8a41abf401bd3c1e4ac1d63190454R171).

Maybe worth a comment here to say this is the assumption in this method.

fendermint/vm/topdown/src/finality.rs
```diff
@@ -0,0 +1,476 @@
...
+                proposal: proposal.height,
+                parent: latest_height,
+            });
+        }
+
+        let last_committed_finality = self.last_committed_finality.read()?;
+        if proposal.height <= last_committed_finality.height {
+            return abort(Error::HeightAlreadyCommitted(proposal.height));
+        }
+        Ok(())
```
aakoshh, line 199: This doesn't seem to be the whole picture. [Here](https://github.com/consensus-shipyard/fendermint/pull/206/files#diff-8edd9923a9eb262950fd919a858ece4449d8a41abf401bd3c1e4ac1d63190454R129) you consider final only something that has certain embedding, yet in this method we accept any height such that `last_committed_finality.height < height <= latest_height`. So it doesn't check that this height is indeed final.

aakoshh (commented):
Here's another round of comments, I'm sorry I seem to have missed this PR last week.

fendermint/vm/topdown/src/sync.rs
```diff
@@ -0,0 +1,206 @@
...
+
+        // we are going backwards in terms of block height, the latest block height is lower
+        // than our previously fetched head. It could be a chain reorg. We clear all the cache
+        // in `provider` and start from scratch
+        if starting_height > latest_height {
+            todo!()
+        }
+
+        let new_parent_views =
+            get_new_parent_views(agent_proxy, starting_height, latest_height).await?;
```
aakoshh, line 85: Okay so we say we go form the latest height we know of + 1, and fetch until the latest height we see, then feed the blocks in this new range into our cache. Then, when we are asked what is final, we look into that cache and subtract 10 and call that block final.

The problem is that there _can_ be reorgs at the tip which this algorithm will fail to detect because of the +1 thing.

aakoshh: Like I mentioned before I saw two options here:

#### Only fetch finalized blocks
1. Get the height of the parent chain
2. Subtract N to get the finalized height
3. Fetch blocks between our last fetch and the new finalized height. 

#### Deal with reorgs
1. Get the height of the parent chain
2. Fetch _again_ the block hash at the last height we fetched before
3. If it's the same, keep fetching until the latest height and append the new ones to the ache
4. If it's not the same, keep going back until we find a height where we still agree, and replace the ones in the cache before appending new ones.

Since we generally don't have to do anything with pending block, we might as well not fetch them, and have a dedicated `TVar` for the latest height, rather than get it as the height of the highest block we fetched.

fendermint/vm/topdown/examples/run_polling.rs
```diff
@@ -0,0 +1,97 @@
...
+    let chain_head = agent_proxy.get_chain_head_height().await.unwrap();
+
+    let config = Config {
+        chain_head_delay: 10,
+        polling_interval_secs: 5,
+    };
+    let provider = InMemoryFinalityProvider::new(
+        config.clone(),
+        IPCParentFinality {
+            height: chain_head - 20,
```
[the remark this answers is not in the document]
cryptoAtwill: I should add more comments to this. This is just a mocked value since we dont really have a contract to store the actual finality.

fendermint/vm/topdown/src/lib.rs
```diff
@@ -0,0 +1,72 @@
...
+
+pub trait ParentViewProvider {
+    /// Get the latest height of the parent recorded
+    fn latest_height(&self) -> StmResult<Option<BlockHeight>, Error>;
+    /// Get the block hash at height
+    fn block_hash(&self, height: BlockHeight) -> StmResult<Option<BlockHash>, Error>;
+    /// Get the validator set at height
+    fn validator_set(&self, height: BlockHeight) -> StmResult<Option<ValidatorSet>, Error>;
+    /// Get the top down messages at height
+    fn top_down_msgs(&self, height: BlockHeight) -> StmResult<Vec<CrossMsg>, Error>;
```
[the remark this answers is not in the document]
cryptoAtwill: Yep, this is the idea. There is an issue in solidity contracts that returns all the data in one query. We should be able to group `validator_set` and `top_down_msgs` into one method. This way we dont have to deal with so many `None`.

fendermint/vm/topdown/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+serde_json = { workspace = true }
+cid = { workspace = true }
+fvm_ipld_encoding = { workspace = true }
+num-traits = { workspace = true }
+async-stm = { workspace = true }
+thiserror = { workspace = true }
+ipc-agent-sdk = { workspace = true }
+fvm_shared = { workspace = true }
+
+tracing-subscriber = "0.3.17"
```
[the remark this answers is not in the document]
cryptoAtwill: Yep, `dev-dependencies` works and shifted.

fendermint/vm/topdown/examples/run_polling.rs
```diff
@@ -0,0 +1,97 @@
...
+    set_network_from_env();
+
+    tracing_subscriber::fmt()
+        .with_max_level(tracing::Level::INFO)
+        .with_target(false)
+        .init();
+
+    let url = std::env::var("URL").unwrap_or_else(|_| "http://0.0.0.0:3030/json_rpc".to_string());
+    let raw_target_subnet_id = std::env::var("TARGET").unwrap_or_else(|_| "/r31415926".to_string());
+    let subnet = SubnetID::from_str(&raw_target_subnet_id).unwrap();
```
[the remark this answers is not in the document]
cryptoAtwill: updated

fendermint/vm/topdown/examples/run_polling.rs
```diff
@@ -0,0 +1,97 @@
...
+            println!("proposal: {proposal:?}");
+
+            if let Some(p) = proposal {
+                let msgs = provider.top_down_msgs(p.height)?;
+                println!("topdown messages: {:?}", msgs);
+            }
+
+            Ok(())
+        })
+        .await
```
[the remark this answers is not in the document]
cryptoAtwill: with the async-stm 0.4, there is no need to do the matching anymore, updated already.

fendermint/vm/topdown/src/lib.rs
```diff
@@ -0,0 +1,72 @@
...
+    /// The latest chain height
+    pub height: BlockHeight,
+    /// The block hash. For FVM, it is a Cid. For Evm, it is bytes32 as one can now potentially
+    /// deploy a subnet on EVM.
+    pub block_hash: BlockHash,
+}
+
+pub trait ParentViewProvider {
+    /// Get the latest height of the parent recorded
+    fn latest_height(&self) -> StmResult<Option<BlockHeight>, Error>;
```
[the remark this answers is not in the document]
cryptoAtwill: I was deciding on completely using `Result` or just `Option`, but then I reckoned that `Result` is better because there are different scenarios that it could result in `Error` and upstream might want to handle the different scenarios. Just returning `None` might not be able to differentiate the scenarios of `None`. In the end `StmResult<Option<BlockHeight>, Error>` is chosen because if the parent has stopped producing block in someway and parent view has caughted up, so there wont be any new `height`, I don't consider this as an error, but a `None`. That's why it's `StmResult<Option<BlockHeight>, Error>`.

fendermint/vm/topdown/src/finality.rs
cryptoAtwill: comment added

fendermint/vm/topdown/src/sync.rs
cryptoAtwill: Nice one!

I have updated the ending height from `latest` chain head to `latest` chain head - N. But addressing the reorgs, reorg detection + error handling, I feel we might have to open another issue to address it.

I think we have two cases for reorgs, one is for the parent, one is for the child, i.e. fendermint.

For dealing with the cache in both scanarios, we can just clear cache and restart. The way we are garbage collection cache is when finality is committed, I reckon the overall data stored in cache should be small. Clearing the cache should be a `brutal` and simple way to handle this. I think the way you described can be used to detect the parent block height where the reorg happens, as the current check might be too simple. Also we might have to consider the case where the polling is just starting and the cache is empty, but parent reorg has already happened.

But I'm afraid the more complex issue is revert the execution of top down messages and revert the committed validator set.

That's why I feel we should start another issue to address the reorg while this PR just ensures the happy path works first.

fendermint/vm/topdown/src/finality.rs
cryptoAtwill: I guess this one is about how strict we want the checking to be. According to the design doc, this is the check, as it is more tolerant of the block height difference and relying on the proposals as a final check.

fendermint/vm/topdown/src/cache.rs
```diff
@@ -0,0 +1,282 @@
...
+        }
+    }
+
+    /// Create a cache with key increment 1
+    pub fn sequential() -> Self {
+        Self {
+            increment: K::one(),
+            data: Default::default(),
+        }
+    }
```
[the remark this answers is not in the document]
cryptoAtwill: Default implementation added, but I think it's still good to keep this method as it's more informative than default.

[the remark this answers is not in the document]
aakoshh: Never mind, `sequential` is fine, but I reckon it a `Default` instance wouldn't be too bad either.

fendermint/vm/topdown/examples/run_polling.rs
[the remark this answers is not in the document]
aakoshh: Cool. By the way if within STM you want to do A, get out of it for some reason and do B instead, then the combinators you want to use are `retry` and `or.

Here's an [example](https://github.com/aakoshh/async-stm-rs/blob/v0.4.0/src/test.rs#L209-L216), which is a copy of a test scenario else implemented in a different system, and I wanted to show that it works here. It's a bit confusing because I use `abort` with an error normally, and here it's just an alias for `retry`. But there you go, with retry it backs out, `or` discards the changes and goes into the other branch.

fendermint/vm/topdown/src/lib.rs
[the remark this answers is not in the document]
aakoshh: But this is the provider, it's there to provide a view to the parent. It was absolutely not clear anywhere that if I ask "what's the latest height on the parent that you know of" it might return `None` even after it has returned `Some` before because it considers the parent stopped producing blocks. Why complicate it like this? How long must the parent be dead? Why do we consider this scenario at all? Why not let the caller decide that `latest_height` has not moved for a long time?

fendermint/vm/topdown/src/finality.rs
aakoshh: @adlrocha can you chime in here. If this is true the design doc seems wrong to me.

aakoshh: Once again what I am seeing is that the proposal check _does not finality check whatsoever_.

fendermint/vm/topdown/src/sync.rs
aakoshh: > But I'm afraid the more complex issue is revert the execution of top down messages and revert the committed validator set.

Well, we should never have to revert the execution of top down messages and committed validator sets, because we should not execute them until they are final. I know you added this default delay of 10 blocks, but to be correct, before something like the Granite work is added to Filecoin, we must actually use 900 blocks if the parent is Lotus. 

> I think we have two cases for reorgs, one is for the parent, one is for the child, i.e. fendermint.

If you mean what if Fendermint reorgs, you don't have to worry, it never does.

fendermint/vm/topdown/examples/run_polling.rs
```diff
@@ -0,0 +1,117 @@
...
+use ipc_sdk::subnet_id::SubnetID;
+use num_traits::FromPrimitive;
+use std::str::FromStr;
+use std::sync::Arc;
+use std::time::Duration;
+use tracing::Level;
+
+#[derive(Parser, Debug)]
+pub struct Options {
+    /// The URL of the ipc agent's RPC endpoint.
```
aakoshh, line 21: Ah, okay, I figured later this would be the agent. 

You know what would be super useful is to add a README.md to the `topdown` crate with a section about how to run this example, which should include instructions for starting both Lotus and the agent, or at least links to where this is explained, because I don't think readers will know, which limits the usefulness of this example.

fendermint/vm/topdown/examples/run_polling.rs
```diff
@@ -0,0 +1,117 @@
...
+
+    /// Enable DEBUG logs.
+    #[arg(long, short)]
+    pub verbose: bool,
+
+    /// The subnet id expressed a string
+    #[arg(long, short)]
+    pub subnet_id: String,
+
+    /// The subnet id expressed a string
```
aakoshh, line 38: Wrong docs. If this is between mainnet and testnet, maybe it could say which value corresponds to what.

fendermint/vm/topdown/src/sync.rs
```diff
@@ -0,0 +1,211 @@
...
+
+        // we are going backwards in terms of block height, the latest block height is lower
+        // than our previously fetched head. It could be a chain reorg. We clear all the cache
+        // in `provider` and start from scratch
+        if starting_height > ending_height {
+            todo!()
+        }
+
+        let new_parent_views =
+            get_new_parent_views(agent_proxy, starting_height, ending_height).await?;
```
aakoshh, line 95: Need to be somewhat careful here that if a node hasn't seen anything previously then it will I think start from the Genesis of Lotus and load everything into memory. 

The IPC genesis of Fendermint should have a non-zero starting height (the height at which this subnet has been created), and as we discussed earlier this routine must not be started while the node is syncing with others, so new joiners don't fetch the entire history of the subnet.

fendermint/vm/topdown/src/sync.rs
aakoshh: Okay so you said you want to deal with reorgs in a different issue. Can you link that here, please?

fendermint/vm/topdown/src/sync.rs
```diff
@@ -0,0 +1,211 @@
...
+    let mut block_height_to_update = vec![];
+    for h in start_height..=end_height {
+        let block_hash = agent_proxy
+            .get_block_hash(h)
+            .await
+            .context("cannot fetch block hash")?;
+        let validator_set = agent_proxy
+            .get_validator_set(h)
+            .await
+            .context("cannot fetch validator set")?;
```
aakoshh, line 144: It would be worth creating an issue to optimise this so that if there's no change then it only records the configuration ID. Like a method that says `get_validator_set_if_changed(h, current_id)` and it returns `None` if it's the same as the `current_id`, so we don't store a copy of the entire power table at every height in memory if nothing happened. cc @adlrocha

fendermint/vm/topdown/src/lib.rs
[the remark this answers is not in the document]
aakoshh: Yes, why don't we do that, seeing that in the implementation these are actually stored together and breaking them up is artificial.

fendermint/vm/topdown/src/finality.rs
aakoshh: Can't see where but I believe you.

fendermint/vm/topdown/src/sync.rs
aakoshh: Still think it would be great if there was something to verify that the block hash at the latest_height is still what we thought it was before, and panic if not because it means our delay settings is bad, but at least we wouldn't carry on building a non-chain.

fendermint/vm/topdown/src/finality.rs
```diff
@@ -0,0 +1,479 @@
...
+        self.check_height(proposal)?;
+        self.check_block_hash(proposal)
+    }
+
+    fn on_finality_committed(&self, finality: &IPCParentFinality) -> StmResult<(), Error> {
+        // the height to clear
+        let height = finality.height;
+
+        self.parent_view_data.height_data.update(|mut cache| {
+            cache.remove_key_below(height + 1);
```
aakoshh, line 162: I wonder if this might be working on the previous assumption where the cache contained the blocks all the way to the tip of the chain, rather than just the finalized block. There, if you had the chain cached up to 102, you proposed for finality height 92, and when it was executed you removed everything below 93. The cache still remembered what the highest height was. 

Now, it only contains entries up to 92, so by removing basically it clears out potentially everything. Is that okay?

aakoshh: I think it is because we write to `last_committed_finality`, just wanted to make sure there's no problem because you rely so much on insertions range checks.

fendermint/vm/topdown/src/finality.rs
```diff
@@ -0,0 +1,479 @@
...
+        } else {
+            return abort(Error::HeightNotReady);
+        };
+
+        // latest height has not reached, we should wait or abort
+        if latest_height < self.config.chain_head_delay {
+            return abort(Error::HeightThresholdNotReached);
+        }
+
+        let height = latest_height - self.config.chain_head_delay;
```
aakoshh, line 132: This seems to be not needed any more, because the syncer only adds data that is final (or should only do so anyway, like I said you should add docstrings to communicate this).

fendermint/vm/topdown/src/lib.rs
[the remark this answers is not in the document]
aakoshh: Now that I see that the parent view only has finalized blocks, and it also removes them after they have been executed, I feel this whole interface is rather misleading. `latest_height` is really `latest_finalized_height` _but_ it's `None` after we execute, so it's like the `next_finalized_height`, but that's exactly what the `next_proposal` (forgot the name) is after, so maybe it would be easier to just hide all these details and reduce them to the three methods that the interpreter actually needs to call (`next`, `check`) and the `new` which is called to extend it. Even the syncer can use `next` to figure out where to resume fetching from.

fendermint/vm/topdown/src/finality.rs
```diff
@@ -0,0 +1,479 @@
...
+        Ok(())
+    }
+}
+
+impl ParentFinalityProvider for InMemoryFinalityProvider {
+    fn last_committed_finality(&self) -> StmResult<IPCParentFinality, Error> {
+        let finality = self
+            .last_committed_finality
+            .read_clone()?
+            .ok_or(StmError::Abort(Error::CommittedFinalityNotReady))?;
```
aakoshh, line 116: By the way I'm surprised that this is possible. There must always be something that we can consider final on the parent: genesis; either that of the entire parent chain, or where the subnet was conceived.

fendermint/vm/topdown/src/finality.rs
```diff
@@ -0,0 +1,479 @@
...
+    fn on_finality_committed(&self, finality: &IPCParentFinality) -> StmResult<(), Error> {
+        // the height to clear
+        let height = finality.height;
+
+        self.parent_view_data.height_data.update(|mut cache| {
+            cache.remove_key_below(height + 1);
+            cache
+        })?;
+
+        self.last_committed_finality.write(Some(finality.clone()))?;
```
aakoshh, line 166: If this is the only place where this is written, how are you supposed to get the cache started?

The syncer looks like it will get errors as long as  `last_committed_finality` is `None`, and not fetch anything, and therefore it will never propose anything. :chicken: :egg: 

Would be nicer if the `InMemoryFinalityProvider::new` demanded the current finality to be provided.

fendermint/vm/topdown/src/finality.rs
cryptoAtwill: Really nice catch. I'm doing the integration testing, and with all the updates, I think the initial logic might have to update a little.

fendermint/vm/topdown/src/sync.rs
cryptoAtwill: There is an issue that tracks `node is syncing with others`. 
Regarding the genesis, I think `fendermint` alone might not be able to handle this, but instead should be handled by the gateway contract. If there are no previous committed finality, the contract should return the `genesis` finality with the `epoch` that the subnet was created and an initial block hash, like `vec![0; 32]`.

fendermint/vm/topdown/src/sync.rs
cryptoAtwill: I think for all configuration id related issues, we should handle them altogether

fendermint/vm/topdown/src/sync.rs
```diff
@@ -0,0 +1,211 @@
...
+        let ending_height = latest_height - config.chain_head_delay;
+
+        tracing::debug!("starting height: {starting_height}, ending_height: {ending_height}");
+
+        // we are going backwards in terms of block height, the latest block height is lower
+        // than our previously fetched head. It could be a chain reorg. We clear all the cache
+        // in `provider` and start from scratch
+        if starting_height > ending_height {
+            todo!()
+        }
```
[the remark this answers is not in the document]
cryptoAtwill: RIght, I removed the `todo` and made the checks more explicit:
```rust
        let last_recorded_height = last_recorded_height(provider).await?;
        let ending_height = ... // some logic omitted for the current discussion;

        tracing::debug!(
            "last recorded height: {}, parent chain head: {}, ending_height: {}",
            last_recorded_height, parent_chain_head_height, ending_height
        );

        if last_recorded_height == ending_height {
            tracing::debug!("the parent has yet to produce a new block, stops at height: {last_recorded_height}");
            continue;
        }

        // we are going backwards in terms of block height, the latest block height is lower
        // than our previously fetched head. It could be a chain reorg. We clear all the cache
        // in `provider` and start from scratch
        if last_recorded_height > ending_height {
            todo!()
        }
```
I still think we have to resync with parent when `last_recorded_height > ending_height` as we still rely on caching for `validating`. So in ` The syncer would consider 91 final now, and attempt to fetch from 93 to 91, hitting this condition and fail, while in fact this is all fine, as long as 92 isn't rolled back.`, even if `92` is all fine, but 91 might have invalid block hash and `process_proposal` might fail. Maybe we can just reset and restart caching then we should be ok.

fendermint/vm/topdown/src/finality.rs
cryptoAtwill: This is a compromise for the circular dependency between provider and App. Maybe making it Optional here is confusing.

fendermint/vm/topdown/src/finality.rs
cryptoAtwill: From `sync` point of view, this should be ok. This implies that the parent is producing block slowly or completely stopped and it means there should not be any parent view related proposal.

fendermint/app/config/default.toml
```diff
@@ -56,3 +56,17 @@ host = "127.0.0.1"
...
+
+# IPC Parent finality related configuration
+[ipc]
+# Subnet id assigned to this fendermint node. Update this to the correct subnet id.
+# TODO: fetch this value from genisis if possible.
+subnet_id = "/r0"
+
+[ipc.config]
+# The number of blocks behind chain head to propose parent finality.
+chain_head_delay = 10
```
aakoshh, line 68:
```suggestion
chain_head_delay = 900
```
It's Filecoin, and this is the safe default.

fendermint/app/src/cmd/run.rs
```diff
@@ -21,6 +23,11 @@ cmd! {
   }
 }
 
+async fn create_parent_finality(settings: &Settings) -> anyhow::Result<InMemoryFinalityProvider> {
```
aakoshh, line 26: Just curious why this has to be `async`.

fendermint/vm/interpreter/src/chain.rs
```diff
@@ -89,32 +97,65 @@ where
...
+        }
+
         // Append at the end - if we run out of block space, these are going to be reproposed in the next block.
         msgs.extend(ckpts);
         Ok(msgs)
     }
 
     /// Perform finality checks on top-down transactions and availability checks on bottom-up transactions.
-    async fn process(&self, pool: Self::State, msgs: Vec<Self::Message>) -> anyhow::Result<bool> {
+    async fn process(&self, state: Self::State, msgs: Vec<Self::Message>) -> anyhow::Result<bool> {
```
aakoshh, line 121:
```suggestion
    async fn process(&self, (pool, provider): Self::State, msgs: Vec<Self::Message>) -> anyhow::Result<bool> {
```
I think this works and would make the rest of the code more readable. `state.0.get_status` is a head scracher.

fendermint/vm/interpreter/src/chain.rs
```diff
@@ -89,32 +97,65 @@ where
...
+                    height,
+                    block_hash,
+                })) => {
+                    let prop = IPCParentFinality {
+                        height: height as u64,
+                        block_hash,
+                    };
+                    if atomically_or_err(|| state.1.check_proposal(&prop))
+                        .await
+                        .is_err()
```
aakoshh, line 152: `check_proposal` should return `Stm<bool>` :pray:

fendermint/vm/topdown/examples/run_polling.rs
aakoshh: :point_up:

fendermint/vm/topdown/src/lib.rs
```diff
@@ -0,0 +1,74 @@
...
+pub trait ParentFinalityProvider: ParentViewProvider {
+    /// Obtains the last committed finality
+    fn last_committed_finality(&self) -> StmResult<IPCParentFinality, Error>;
+    /// Latest proposal for parent finality
+    fn next_proposal(&self) -> StmResult<Option<IPCParentFinality>, Error>;
+    /// Check if the target proposal is valid
+    fn check_proposal(&self, proposal: &IPCParentFinality) -> StmResult<(), Error>;
+    /// Called when finality is committed
+    fn on_finality_committed(&self, finality: &IPCParentFinality) -> StmResult<(), Error>;
+}
```
aakoshh, line 74:
```suggestion
pub trait ParentFinalityProvider: ParentViewProvider {
    /// Obtains the last committed finality
    fn last_committed_finality(&self) -> Stm<IPCParentFinality>;
    /// Latest proposal for parent finality
    fn next_proposal(&self) -> Stm<Option<IPCParentFinality>>;
    /// Check if the target proposal is valid
    fn check_proposal(&self, proposal: &IPCParentFinality) -> Stm<bool>;
    /// Called when finality is committed
    fn on_finality_committed(&self, finality: &IPCParentFinality) -> Stm<()>;
}
```

fendermint/vm/topdown/src/lib.rs
```diff
@@ -0,0 +1,74 @@
...
+    /// The block hash. For FVM, it is a Cid. For Evm, it is bytes32 as one can now potentially
+    /// deploy a subnet on EVM.
+    pub block_hash: BlockHash,
+}
+
+pub trait ParentViewProvider {
+    /// Get the latest height of the parent recorded
+    fn latest_height(&self) -> StmResult<Option<BlockHeight>, Error>;
+    /// Get the block hash at height
+    fn block_hash(&self, height: BlockHeight) -> StmResult<Option<BlockHash>, Error>;
```
aakoshh, line 50:
```suggestion
    fn latest_height(&self) -> Stm<Option<BlockHeight>>;
    /// Get the block hash at height
    fn block_hash(&self, height: BlockHeight) -> Stm<Option<BlockHash>>;
```

fendermint/vm/topdown/src/lib.rs
```diff
@@ -0,0 +1,74 @@
...
+    /// Get the validator set at height
+    fn validator_set(&self, height: BlockHeight) -> StmResult<Option<ValidatorSet>, Error>;
```
aakoshh, line 52:
```suggestion
```

fendermint/vm/topdown/src/finality.rs
aakoshh: This is what we discussed turning into a panic in the new phased initialisation: 
1. Internally the `last_committed_finality` is a `TVar<Option<...>>` and starts empty because we don't want to peak into the genesis file, and we might be restoring from a snapshot anyway that CometBFT needs to fetch for us.
2. Before the genesis is applied or the snapshot is loaded, CometBFT will not give us blocks to process, so the `App` should not call this method. 
3. If it did, that would be a programming error, and we can just `panic!` if the `last_committed_finality` is `None`. 
4. During genesis or loading a snapshot, the interpreter should call `set_committed_finality` to cause this to go become `Some`. 
5. After that it's safe to query and return `Stm<IPCParentFinality>`

fendermint/vm/topdown/src/finality.rs
aakoshh: Hm, I thought this would change.

aakoshh:
So the fixes to the comments in this PR are supposed to be in https://github.com/consensus-shipyard/fendermint/pull/236

PR #203, opened 2023-08-18
FM-195: Implement ProposalInterpreter::prepare for BottomUp

aakoshh:
Closes #195 
Depends on #202 

Added a `ResolvePool::collect_resolved` method to gather all resolved items, which are appended to the messages as checkpoints proposed for execution by the `ChainMessageInterpreter` in the `ProposalInterpreter::prepare` method implementation.

PR #210, opened 2023-08-22
FM-193: Resolve CIDs in the background

### LINKED ISSUE #193: Resolve the CID in the bottom-up checkpoint
Implement a background task that picks up bottom-up checkpoints added to the resolution pool in https://github.com/consensus-shipyard/fendermint/issues/191 and uses the https://github.com/consensus-shipyard/ipc-ipld-resolver to fetch the data from the subnet. Once it's available, mark the checkpoint as available in the pool.

When the component starts, ask the IPC Gateway actor for any outstanding checkpoints, ones which have been relayed, but not executed yet (so, rewards are pending), and resume their resolution. This way if the node is restarted, we don't get stuck never proposing relayed checkpoints.

aakoshh:
Closes #193 
Depends on #203 

Created an `IpldResolver` component which reads tasks from the `ResolvePool` introduced in #201 and uses the IPLD Resolver to fetch data from peers in the subnet. If the resolution fails, it is re-queued after some delay. 

- [x] Execute multiple lookups in the background - this would make sense so slow subnets don't get in the way of others
- [x] Indicate when a resolution can happen from our own subnet, rather than just the child subnet. I think this can be done if the item is added to the pool with immediate effect, in which case it should have a flag inside which can be flipped, and at the next retry it should result in the current subnet attempted as well. This is another place where background resolutions would be better.
- [x] Construct and start an IPLD Resolver service


fendermint/app/src/cmd/run.rs
```diff
@@ -36,11 +43,44 @@ async fn run(settings: Settings) -> anyhow::Result<()> {
...
 
     let resolve_pool = CheckpointPool::new();
 
+    if settings.resolver_enabled() {
+        let service =
+            make_resolver_service(&settings, db.clone(), state_store.clone(), ns.bit_store)?;
+
+        let client = service.client();
+
+        let own_subnet_id = settings.resolver.subnet_id.clone();
```
adlrocha, line 58: I agree with your TODO from above, it'll be great if we can reconcile all the `subnet_id`  config by picking up the value from an IPC setting or, ideally, from genesis.

fendermint/app/src/settings/resolver.rs
```diff
@@ -0,0 +1,112 @@
...
+    pub content: ContentSettings,
+}
+
+#[derive(Clone, Debug, Deserialize)]
+pub struct NetworkSettings {
+    /// Cryptographic key used to sign messages.
+    ///
+    /// This is the name of a Secp256k1 private key file,
+    /// relative to the `home_dir`.
+    local_key: PathBuf,
```
adlrocha, line 40: This is the private key used for the libp2p host identity, i.e. the one that determines the PeerID?

fendermint/app/src/store.rs
```diff
@@ -37,3 +42,75 @@ where
         fvm_ipld_encoding::from_slice(repr).map_err(|e| KVError::Codec(Box::new(e)))
     }
 }
+
+/// A `Blockstore` and `BitswapStore` implementation we can pass to the IPLD Resolver.
+pub struct BitswapBlockstore {
+    /// The `Blockstore` implementation where we the FVM actors store their data.
```
adlrocha, line 48:
```suggestion
    /// The `Blockstore` implementation where the FVM actors store their data.
```

fendermint/app/src/store.rs
```diff
@@ -37,3 +42,75 @@ where
...
+    ///
+    /// This must not be written to by Bitswap operations, because that could result
+    /// in some nodes having some data that others don't, which would lead to a
+    /// consensu failure. We can use read data from it, but not write to it.
+    state_store: NamespaceBlockstore,
+    /// The `Blockstore` implementation where Bitswap operations can write to.
+    bit_store: NamespaceBlockstore,
```
adlrocha, line 55: I really like this dual blockstore approach where we have the `bit_store` as cache and we expose the `state_store`. Really cool!

adlrocha (approved):
Really cool PR, thanks! I think in the future we may want to use the IPLD resolver for a lot of things. It looks really powerful. 

The only thing I feel we should fix ASAP is the way we want to handle the `SubnetID` for Fendermint networks :) It is starting to appear everywhere also for me :)

fendermint/vm/resolver/src/pool.rs
```diff
@@ -32,11 +34,12 @@ impl<T> ResolveStatus<T>
 where
     T: Clone + Hash + Eq + PartialEq + Sync + Send + 'static,
 {
-    pub fn new(item: T) -> Self {
+    pub fn new(item: T, use_own_subnet: bool) -> Self {
```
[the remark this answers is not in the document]
aakoshh: Great question, thanks. I am not sure I added proper docs about this anywhere other than the PR readme (checkbox #2). 

So this isn't about `own` and `parent`, it's primarily about `own` and the _children_. 

There are two situations:
1. Someone is relaying a checkpoint to the validators on the parent subnet from one of the many child subnets in the form of a transaction. Say the signature checks out and the relayer has gas, then we are left with a CID. Where do we resolve it from? The providers of the child subnet it came from. Why not ask everyone? Because only the child subnet providers can be expected to have them. Why not ask anyway? Because in Bitswap we send messages that we need a CID, and we receive it back, or a timeout or maybe a Nack. To limit the amount of network use and waste, I only send it to a limited number of peers at a time. The more peers we send it to who _don't_ have it, the longer it will take to find someone who _does_. 
2. The parent validators (_own subnet_ in this context) finalized a block with a transaction to _execute_ a checkpoint. This required the majority of validators to vote that they have it, however, a minority may not have it yet. In this case they want to resolve the CID _now_ otherwise they cannot execute the block, so they should ask those other validators in their own (parent) subnet who voted for it, as well as the child subnet providers - someone should be able to serve it.

fendermint/app/src/settings/resolver.rs
aakoshh: That's right, see: 
* https://github.com/consensus-shipyard/ipc-ipld-resolver/blob/main/src/behaviour/mod.rs#L30-L35
* https://github.com/consensus-shipyard/ipc-ipld-resolver/blob/main/src/service.rs#L150-L151
* https://github.com/consensus-shipyard/ipc-ipld-resolver/blob/main/src/service.rs#L169

fendermint/app/src/store.rs
aakoshh: Yeah we have to be careful that we don't put stuff into the `state_store` that isn't a result of actor execution. 

In order to make cross-message data available to Bitswap I think we'll have to shovel it form Solidity to some IPLD-native data, and we can store that in the `state_store` because it happens automatically as part of execution. We could store it in `bit_store` too. 

What we receive must go into the `bit_store` though, and then when we want to execute, we have to turn them back to Solidity by pulling from `bit_store` or `state_store`. 

I imagine using namespaces makes stuff slower but safety first :police_officer:

fendermint/vm/resolver/src/ipld.rs
```diff
@@ -0,0 +1,131 @@
...
+            retry_delay,
+            own_subnet_id,
+        }
+    }
+
+    /// Start taking tasks from the resolver pool and resolving them using the IPLD Resolver.
+    pub async fn run(self) {
+        loop {
+            let (task, use_own_subnet) = atomically(|| {
+                let task = self.queue.read()?;
```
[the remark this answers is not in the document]
aakoshh: Good question, yes and no. The original version of STM would block this thread, but the async version doesn't. It will just add itself to the wait queue of the `TVar` and `await` for some change. Because it's `async` and using `tokio`, it will not block, but it also won't return until something appears in the queue.

fendermint/vm/resolver/src/ipld.rs
```diff
@@ -0,0 +1,131 @@
...
+
+        let err = match (theirs, own) {
+            (Err(e), _) => {
+                tracing::error!(error = e.to_string(), "failed to submit resolution task");
+                // The service is no longer listening, we might as well stop taking new tasks from the queue.
+                // By not quitting we should see this error every time there is a new task, which is at least is a constant reminder.
+                return;
+            }
+            (Ok(Ok(())), _) | (_, Some(Ok(Ok(())))) => None,
+            (Ok(Err(e)), _) => Some(e),
```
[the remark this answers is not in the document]
aakoshh: Thanks for asking, it's definitely not easy to see from this why this is correct. It was actually the compiler telling me that if I added more cases they would never be matched after this. 

Okay, so to answer your question: no, we don't need `from_theirs` and `from_own` to succeed at the same time, we just want one of them to go through, we just want the data ASAP. I could use `FuturesUnordered` and react as soon as one is ready, but then I wouldn't know if I should reschedule or wait for the other to succeed. It may be easier overall to do that, I'm not sure.

The way this Bitswap library works is that it deposits the date into the `BitswapStore`, so all we get back is either a `()` to indicate success, or an `Error` if it exhausted all options. So I'm thinking as long as I just see success from one of them, we are fine, the data arrived. 

If one is `Err`, the other is most likely `Err` as well, the `Service` died, so we are only interested in `Ok`:
* The first arm matches if `from_theirs` was successful, or if we tried `from_own` and _that_ was a success. 
* The second arm checks if `from_theirs` failed, and ignores the result of `from_own` (`None` or `Some` doesn't matter), because if `from_theirs` was successful, it would have matched the first arm already. 

So this is a not as fast as possible in terms of not retrying as quickly as it could, but it sends as few requests as possible by waiting until it knows that both failed before retrying, and it only retries once.

fendermint/vm/resolver/src/pool.rs
```diff
@@ -46,6 +49,38 @@ where
     }
 }
 
+/// Tasks emitted by the pool for background resolution.
+#[derive(Clone)]
+pub struct ResolveTask {
+    /// Content to resolve.
+    key: ResolveKey,
+    /// Flag to flip when the task is done.
+    is_resolved: TVar<bool>,
```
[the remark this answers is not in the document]
aakoshh: No, there is only up to one resolve task working on a `ResolveTask`, because of the way `own` and `theirs` are both awaited before the whole task is rescheduled for later.

PR #212, opened 2023-08-24
FM-211: eth address in logs and filters

### LINKED ISSUE #211: Use Ethereum address in events and filters
```[tasklist]
### Tasks
- [x] Event emitter turned into Ethereum address, not masked ID
- [x] Filter subscriptions turn Ethereum address back into ID for the filter expression
- [x] Return Ethereum address from contract creation
```

aakoshh:
Closes #211

### DIFF
--- a/fendermint/eth/api/src/conv/from_tm.rs
+++ b/fendermint/eth/api/src/conv/from_tm.rs
@@ -7,15 +7,18 @@ use std::str::FromStr;
 
 use anyhow::{anyhow, Context};
 use ethers_core::types::{self as et};
+use fendermint_rpc::client::FendermintClient;
+use fendermint_rpc::query::QueryClient;
 use fendermint_vm_actor_interface::eam::EthAddress;
 use fendermint_vm_message::{chain::ChainMessage, signed::SignedMessage};
+use fvm_shared::address::Address;
 use fvm_shared::chainid::ChainID;
 use fvm_shared::{bigint::BigInt, econ::TokenAmount};
 use lazy_static::lazy_static;
 use tendermint::abci::response::DeliverTx;
 use tendermint::abci::{self, EventAttribute};
 use tendermint::crypto::sha256::Sha256;
-use tendermint_rpc::endpoint;
+use tendermint_rpc::{endpoint, Client};
 
 use super::from_fvm::{to_eth_address, to_eth_signature, to_eth_tokens};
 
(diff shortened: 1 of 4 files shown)


fendermint/eth/api/src/conv/from_tm.rs
```diff
@@ -342,33 +351,60 @@ fn maybe_contract_address(deliver_tx: &DeliverTx) -> Option<EthAddress> {
...
         .map(|cr| {
-            // We can return `cr.eth_address` here and that's an address usable for calling the contract.
-            // However, the events are reported using the actor ID, so unless we want to translate those
-            // to ETH addresses on the fly, we can return a masked address here and see if that works.
-            EthAddress::from_id(cr.actor_id)
+            // We can return either `cr.actor_id` as a masked address,
+            // or `cr.eth_address`. Both addresses are usable for calling the contract.
+            // However, the masked ID doesn't work with some of the Ethereum tooling which check some hash properties.
+            // We also have to make sure to use the same kind of address that we do in the filtering and event logs,
+            // otherwise the two doesn't align and it makes the API difficult to use. It's impossible(?) to find out
```
adlrocha, line 358: It is impossible without access to the state tree, as far as I know

fendermint/eth/api/src/conv/from_tm.rs
```diff
@@ -342,33 +351,60 @@ fn maybe_contract_address(deliver_tx: &DeliverTx) -> Option<EthAddress> {
...
             .attributes
             .iter()
             .find(|a| a.key == "emitter")
             .and_then(|a| a.value.parse::<u64>().ok())
-            .map(EthAddress::from_id)
-            .map(|a| et::H160::from_slice(&a.0))
-            .unwrap_or_default();
+            .ok_or_else(|| anyhow!("cannot find the 'emitter' key"))?;
+
+        // TODO: Cache the emitters to save API roundtrips.
```
adlrocha, line 391: This can potentially be removed once we fix the return inconsistency between Eth and Fendermint transactions returned, right?

adlrocha (approved):
Thank you for jumping right into this :pray:

fendermint/eth/api/src/conv/from_tm.rs
aakoshh: No, it can't, because the `emitter` in the FVM [StampedEvent](https://github.com/filecoin-project/ref-fvm/blob/master/shared/src/event/mod.rs#L15) is an `ActorID`, and we can't possibly know what to turn it into. Ostensibly we could emit a "delegated emitter" too, and then we could filter on that one, or we do what we do now to translate on the fly for Ethereum.

aakoshh: The fix for the transaction hash is similar in that it will rely on adding further indexed fields, so we can explore this idea as well.

PR #214, opened 2023-08-24
FM-213: Cache actor ID and address

### LINKED ISSUE #213: Use caching when looking up actor IDs
Follow up https://github.com/consensus-shipyard/fendermint/pull/212 with some sort of caching solution.

Perhaps the simplest would be to add a cache to the `FendermintClient` instance, and two new methods to lookup the delegated address of an ID or the ID of an address. But currently the client is stateless, so we could also do a cache on top of the client, passing it to the things that need it, making it part of the `JsonRpcState`. That's cleaner but requires more parameter threading.

aakoshh:
Closes #213 
Depends on #212

Added an `AddressCache` to cache `ActorID <-> DelegatedAddress` relations.

### DIFF
--- a/fendermint/eth/api/src/cache.rs
+++ b/fendermint/eth/api/src/cache.rs
@@ -1,0 +1,99 @@
+// Copyright 2022-2023 Protocol Labs
+// SPDX-License-Identifier: Apache-2.0, MIT
+
+use std::{
+    collections::HashMap,
+    sync::{Arc, RwLock},
+};
+
+use anyhow::Context;
+use fendermint_rpc::client::FendermintClient;
+use fendermint_rpc::query::QueryClient;
+use fvm_shared::{address::Address, ActorID};
+use tendermint_rpc::Client;
+
+/// Facilitate Ethereum address <-> Actor ID lookups.
+#[derive(Clone)]
+pub struct AddressCache<C> {
+    client: FendermintClient<C>,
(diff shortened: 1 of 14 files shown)


fendermint/eth/api/src/cache.rs
```diff
@@ -0,0 +1,105 @@
...
+        let res = self
+            .client
+            .actor_state(addr, None)
+            .await
+            .context("failed to lookup actor state")?;
+
+        match res.value {
+            Some((id, _)) => {
+                self.set_id(*addr, id);
+                if let Payload::Delegated(_) = addr.payload() {
```
adlrocha, line 54: Why do you need to explicitly check if it is a delegated address to index the id by address, and you don't check to index by address by id?

adlrocha (approved):
LGTM. Thanks!

fendermint/eth/api/src/cache.rs
aakoshh: When I look up the `Address` by ID, all I can get is the optional `delegated_address` field in the `ActorState`. However if you look up an ID by `Address`, the system can handle any type of address, so it's not symmetric. I only want to insert into the id-to-address if it's the same kind of address as what I'd do if I did the opposite lookup. I don't want to fill in the space with something that isn't a delegated address. It's contrived, I know.

PR #222, opened 2023-08-28
FM-150: Subnet deployment scripts

### LINKED ISSUE #150: IPC: Child subnet deployment scripts
In Eudico, we include a [set of convenient scripts for users to be able to deploy](https://github.com/consensus-shipyard/lotus/tree/78bcc4706fcfc146c60dbef0c5de72580dbd1ab2/scripts/ipc) the required infrastructure to run their own child subnet validator without a lot of operational overhead. We should have something equivalent with Fendermint and integrate it in the agent's `install-infra` make rule.

adlrocha:
This PR adds a makefile to run two local test networks: one for one node and the second for four nodes.

Closes #150 

> WIP

The idea is to leverage [cargo-make](https://sagiegurari.github.io/cargo-make/) to configure the deployment and trigger the end-to-end.


aakoshh:
I think these can live in just a top level `infra` directory, next to `docker` and `scripts`, and leave the `fendermint` directory for Rust code.

docs/localnet.md
```diff
@@ -0,0 +1,54 @@
+# Local Testnets
+
+Prerequisites:
+```bash
+make build
```
aakoshh, line 5:
```suggestion
make build docker-build
```

I see below that it's about docker, so..

docs/localnet.md
```diff
@@ -0,0 +1,54 @@
...
+
+```
+
+To stop the network:
+```bash
+cargo make --makefile ./infra/Makefile.toml testnet-down
+```
+
+The testnet contains four logical nodes. Each node consists of cometbft, fendermint, and ethapi containers.
+The testnet network is `192.167.10.0/24`.
```
aakoshh, line 36: How does this IP work?

docs/localnet.md
```diff
@@ -0,0 +1,54 @@
...
+
+ETH-API is accessible on the following interfaces:
+- `192.167.10.10:8545`
+- `192.167.10.11:8546`
+- `192.167.10.12:8547`
+- `192.167.10.13:8548`
```
aakoshh, line 42: Why does both the IP _and_ the port change? If it's possible to assign different IPs to different nodes, yhy not keep the port be the same on all of them?

docs/localnet.md
```diff
@@ -0,0 +1,54 @@
...
+
+The deployment process is as follows:
+- Remove all docker containers, files, networks, etc. from the previous deployment
+- Create all necessary directories
+- Initialize CometBFT testnet by creating `config` and `data` directories using `cometbft` tools
+- Read cometbft nodes private keys,derive node IDs and store in `config.toml` for each node
+- Create the `genesis` file for Fendermint
+- Share the genesis among all Fendermint nodes
+- Run Fendermint application in 4 containers
+- Run Eth API in 4 containers
```
aakoshh, line 54:
```suggestion
- Run Fendermint application in 4 containers
- Run CometBFT in 4 containers
- Run Eth API in 4 containers
```
At least I assume that's what's happening.

docs/localnet.md
```diff
@@ -0,0 +1,54 @@
...
+The testnet contains four logical nodes. Each node consists of cometbft, fendermint, and ethapi containers.
+The testnet network is `192.167.10.0/24`.
+
+ETH-API is accessible on the following interfaces:
+- `192.167.10.10:8545`
+- `192.167.10.11:8546`
+- `192.167.10.12:8547`
+- `192.167.10.13:8548`
+
+## Development
```
aakoshh, line 44: Is this the subtitle you wanted? The process below doesn't say anything about development.

fendermint/app/src/cmd/key.rs
```diff
@@ -66,6 +70,27 @@ cmd! {
   }
 }
 
+cmd! {
+  AddPeer(self) {
+    let node_key = NodeKey::load_json_file(&self.node_key_file).unwrap();
```
aakoshh, line 75: The [exec](https://github.com/consensus-shipyard/fendermint/blob/60ac4f47331f5c95a38b00e2d0aecace0a903caf/fendermint/app/src/cmd/mod.rs#L43) method returns `anyhow::Result`, which is [handled](https://github.com/consensus-shipyard/fendermint/blob/60ac4f47331f5c95a38b00e2d0aecace0a903caf/fendermint/app/src/main.rs#L29-L31) in `main` by logging and exiting. 

So let's not use `.unwrap()` inside but the `?` operator [like all the other commands](https://github.com/consensus-shipyard/fendermint/blob/60ac4f47331f5c95a38b00e2d0aecace0a903caf/fendermint/app/src/cmd/key.rs#L46), perhaps with [context](https://docs.rs/anyhow/latest/anyhow/trait.Context.html) to attach some meaningful error message.

fendermint/app/src/cmd/key.rs
```diff
@@ -66,6 +70,27 @@ cmd! {
...
+    let peer_id : String = node_key.node_id().to_owned().to_string()+"@"+self.network_addr.as_str();
```
aakoshh, line 76:
```suggestion
    let peer_id = format!("{}@{}", node_key.node_id(), self.network_addr);
```

fendermint/app/src/cmd/key.rs
```diff
@@ -66,6 +70,27 @@ cmd! {
...
+        Ok(mut v) => {
+            if v.is_empty()  {
+                std::fs::write(&self.local_peers_file, peer_id)?;
+            } else {
+                v = v.to_owned().to_string()+ "," + peer_id.as_str();
+                std::fs::write(&self.local_peers_file, v)?;
+            }
+        }
+        Err(_) => {
+                std::fs::write(&self.local_peers_file, peer_id)?;}
```
aakoshh, line 88: What's that `}` at the end? 

It looks like you haven't configured your editor to use `rustfmt`. CI will reject unformatted code.

fendermint/app/src/cmd/key.rs
```diff
@@ -66,6 +70,27 @@ cmd! {
...
 
+cmd! {
+  AddPeer(self) {
+    let node_key = NodeKey::load_json_file(&self.node_key_file).unwrap();
+    let peer_id : String = node_key.node_id().to_owned().to_string()+"@"+self.network_addr.as_str();
+    let peers = std::fs::read_to_string(&self.local_peers_file).context("failed to read peer file");
+    match peers {
+        Ok(mut v) => {
+            if v.is_empty()  {
+                std::fs::write(&self.local_peers_file, peer_id)?;
```
aakoshh, line 81: It would be easier to follow what's going on if the file was only written to in one place:
1. Read the file into a string, or get the default one if it doesn't exist.
2. Modify the string by appending the new peer, with or without a comma (do you care about duplicates?)
3. Write the string to the file.

fendermint/app/src/options/key.rs
```diff
@@ -13,6 +13,8 @@ pub enum KeyCommands {
     IntoTendermint(KeyIntoTendermintArgs),
     /// Convert a public key file from base64 into an f1 Address format an print it to STDOUT.
     Address(KeyAddressArgs),
+    /// TBD
```
aakoshh, line 16: Missing docs

fendermint/app/src/options/key.rs
```diff
@@ -21,6 +23,19 @@ pub struct KeyArgs {
     pub command: KeyCommands,
 }
 
+#[derive(Args, Debug)]
+pub struct AddPeer {
+    /// The path to node key file.
```
aakoshh, line 28:
```suggestion
    /// The path to a CometBFT node key file.
```
Let's give an example of its default location, so people have an idea where to look to understand what this is about.

fendermint/app/src/options/key.rs
```diff
@@ -21,6 +23,19 @@ pub struct KeyArgs {
...
+    #[arg(long, short)]
+    pub node_key_file: PathBuf,
+    /// The path to a local file with persistent peers.
+    #[arg(long, short)]
+    pub local_peers_file: PathBuf,
+    /// The path to a local file with persistent peers.
```
aakoshh, line 34: Wrong docs.

fendermint/app/src/options/key.rs
```diff
@@ -21,6 +23,19 @@ pub struct KeyArgs {
...
+    #[arg(long, short)]
```
aakoshh, line 35: You will have to say what the `short` should be because both `node_key_file` and `network_addr` start with `n`. See for example [here](https://github.com/consensus-shipyard/fendermint/blob/60ac4f47331f5c95a38b00e2d0aecace0a903caf/fendermint/app/src/options/rpc.rs#L153-L158)

fendermint/app/src/options/key.rs
```diff
@@ -21,6 +23,19 @@ pub struct KeyArgs {
     pub command: KeyCommands,
 }
 
+#[derive(Args, Debug)]
+pub struct AddPeer {
+    /// The path to node key file.
+    #[arg(long, short)]
+    pub node_key_file: PathBuf,
+    /// The path to a local file with persistent peers.
```
aakoshh, line 31: It would be good to explain a little bit more whether this is some standard CometBFT file that people should find, or an arbitrary path where the peer IDs will be added (I think this is the case), so that they can just put down whatever they want. Maybe say it will be created if it doesn't exist.

fendermint/app/src/options/key.rs
aakoshh: If this is a standard CometBFT file, let's mention its default location, so people can find it easier. At the moment I had to check if it's an input or an output.

aakoshh: Ah, sorry, this isn't even a file. So let's document the format, of it, mention which port to put in.

infra/docker-compose.yml
```diff
@@ -0,0 +1,241 @@
+version: '3'
+
+services:
+  cometbft-node0:
+    container_name: cometbft-node0
+    image: "cometbft/cometbft:v0.37.x"
```
aakoshh, line 6: You can use a `.env` file to put things like these versions and not have to repeat them across all the containers, see [here](https://docs.docker.com/compose/environment-variables/set-environment-variables/).

infra/docker-compose.yml
```diff
@@ -0,0 +1,241 @@
...
+    healthcheck:
+      test: curl --fail http://localhost:26657 || exit 1
+      interval: 8s
+      timeout: 10s
+      retries: 20
+    networks:
+      localnet:
+        ipv4_address: 192.167.10.2
+
+  cometbft-node1:
```
aakoshh, line 26: I wonder if [extends](https://docs.docker.com/compose/multiple-compose-files/extends/) would help cut down the repetition. 

In the past what I did was to put everything that can change into the `.env` file and then keep a single `docker-compose` for the set of services I want to start (just a single set of 3 containers in this case) and tell them which `.env` file to use when I up or down services, and then I can say how many containers I want in the setup script. 

I can see how this is easier to understand, but also more cumbersome to keep in sync.

infra/docker-compose.yml
```diff
@@ -0,0 +1,241 @@
...
+      retries: 20
+    networks:
+      localnet:
+        ipv4_address: 192.167.10.5
+
+  fendermint-node0:
+    container_name: fendermint-node0
+    image: "fendermint:latest"
+    build:
+      context: .
```
aakoshh, line 95: How does this `build` work? Is it producing a different image than `make docker-build`?

infra/docker-compose.yml
```diff
@@ -0,0 +1,241 @@
...
+        ipv4_address: 192.167.10.6
+
+  fendermint-node1:
+    container_name: fendermint-node1
+    image: "fendermint:latest"
+    build:
+      context: .
+    environment:
+      - FM_DATA_DIR=/data/fendermint/data
+      - FM_CHAIN_NAME=root
```
aakoshh, line 113: env var?

infra/Makefile.toml
```diff
@@ -0,0 +1,597 @@
...
+echo - Keys directory: ${KEYS_DIR}
+echo - Genesis file: ${GENESIS_FILE}
+echo - Private key: ${PRIV_KEY_PATH}
+echo - Network: ${NETWORK_NAME}
+echo - CometBFT container: ${CMT_CONTAINER_NAME}
+echo - Fendermint container: ${FM_CONTAINER_NAME}
+echo
+echo
+echo 4 nodes testnet layout:
+echo - IPC directory: ${BASE_TESTNET_DIR}
```
aakoshh, line 84: NB the docker-compose file doesn't respect this value.

infra/Makefile.toml
```diff
@@ -0,0 +1,597 @@
...
+"""
+
+[tasks.fendermint-newaccount]
+cwd = "./target/release"
+script="""
+./fendermint genesis --genesis-file ${GENESIS_FILE} add-account --public-key ${PUB_KEY_PATH} --balance ${BALANCE}
+"""
+
+
+[tasks.fendermint-newgateway]
```
aakoshh, line 198: This task does a lot, it could use some docs. Why is gateway creation, which is just putting a section into the gensis file, coupled with adding a validator and exporting its key to CometBFT?

infra/Makefile.toml
```diff
@@ -0,0 +1,597 @@
...
+  "testnet-config-newaccount",
+  "testnet-config-newgateway",
+  "testnet-share-genesis"
+]
+
+[tasks.testnet-config-newgenesis]
+cwd = "./target/release"
+script = """
+./fendermint genesis --genesis-file $BASE_TESTNET_DIR/genesis.json \
+      new --chain-name testnet --base-fee ${BASE_FEE} --timestamp ${TIMESTAMP}
```
aakoshh, line 342: `--chain-name` here is different than in the `docker-compose` file.

infra/Makefile.toml
```diff
@@ -0,0 +1,597 @@
...
+script="""
+mkdir -p $BASE_TESTNET_DIR/node0/fendermint/keys
+mkdir -p $BASE_TESTNET_DIR/node1/fendermint/keys
+mkdir -p $BASE_TESTNET_DIR/node2/fendermint/keys
+mkdir -p $BASE_TESTNET_DIR/node3/fendermint/keys
+
+./fendermint key gen --out-dir $BASE_TESTNET_DIR/node0/fendermint/keys --name validator0_key
+./fendermint key gen --out-dir $BASE_TESTNET_DIR/node1/fendermint/keys --name validator1_key
+./fendermint key gen --out-dir $BASE_TESTNET_DIR/node2/fendermint/keys --name validator2_key
+./fendermint key gen --out-dir $BASE_TESTNET_DIR/node3/fendermint/keys --name validator3_key
```
aakoshh, line 358: Loop?

infra/Makefile.toml
```diff
@@ -0,0 +1,597 @@
...
+cat $BASE_TESTNET_DIR/genesis.json
+"""
+
+[tasks.testnet-config-peers]
+cwd = "./target/release"
+script="""
+./fendermint key add-peer --node-key-file $BASE_TESTNET_DIR/node0/cometbft/config/node_key.json --network-addr 192.167.10.2:26656 --local-peers-file $BASE_TESTNET_DIR/peers
+./fendermint key add-peer --node-key-file $BASE_TESTNET_DIR/node1/cometbft/config/node_key.json --network-addr 192.167.10.3:26656 --local-peers-file $BASE_TESTNET_DIR/peers
+./fendermint key add-peer --node-key-file $BASE_TESTNET_DIR/node2/cometbft/config/node_key.json --network-addr 192.167.10.4:26656 --local-peers-file $BASE_TESTNET_DIR/peers
+./fendermint key add-peer --node-key-file $BASE_TESTNET_DIR/node3/cometbft/config/node_key.json --network-addr 192.167.10.5:26656 --local-peers-file $BASE_TESTNET_DIR/peers
```
aakoshh, line 386: Loop?

I am surprised a bit that all the p2p addresses use the same port. So these IPs work internally as well as externally? Then why are the p2p ports even mapped to the host?

infra/Makefile.toml
```diff
@@ -0,0 +1,597 @@
...
+
+unset CMT_PERSISTENT_PEERS
+export CMT_PERSISTENT_PEERS=`cat $BASE_TESTNET_DIR/peers`
+echo Persistent peers: $CMT_PERSISTENT_PEERS
+
+sed -i'bak' "s/persistent_peers = \\"\\"/persistent_peers = \\"$CMT_PERSISTENT_PEERS\\"/" $BASE_TESTNET_DIR/node0/cometbft/config/config.toml
+sed -i'bak' "s/persistent_peers = \\"\\"/persistent_peers = \\"$CMT_PERSISTENT_PEERS\\"/" $BASE_TESTNET_DIR/node1/cometbft/config/config.toml
+sed -i'bak' "s/persistent_peers = \\"\\"/persistent_peers = \\"$CMT_PERSISTENT_PEERS\\"/" $BASE_TESTNET_DIR/node2/cometbft/config/config.toml
+sed -i'bak' "s/persistent_peers = \\"\\"/persistent_peers = \\"$CMT_PERSISTENT_PEERS\\"/" $BASE_TESTNET_DIR/node3/cometbft/config/config.toml
```
aakoshh, line 395: Loop?

infra/Makefile.toml
```diff
@@ -0,0 +1,597 @@
...
+    --public-key $BASE_TESTNET_DIR/node0/fendermint/keys/validator0_key.pk --power 1
+
+./fendermint genesis --genesis-file $BASE_TESTNET_DIR/genesis.json add-validator \
+    --public-key $BASE_TESTNET_DIR/node1/fendermint/keys/validator1_key.pk --power 1
+
+./fendermint genesis --genesis-file $BASE_TESTNET_DIR/genesis.json add-validator \
+    --public-key $BASE_TESTNET_DIR/node2/fendermint/keys/validator2_key.pk --power 1
+
+./fendermint genesis --genesis-file $BASE_TESTNET_DIR/genesis.json add-validator \
+    --public-key $BASE_TESTNET_DIR/node3/fendermint/keys/validator3_key.pk --power 1
```
aakoshh, line 411: :loop: ?

aakoshh (commented):
Thank you for putting this together and figuring out how to configure a local CometBFT testnet. 

I am a little confused as to how much in the `Makefile.toml` is a copy of what I have in the smoke tests that is left in by accident. Do you want to exercise the same smoke and ethapi tests?

I'm also not sure `cargo make` is the best tool to use here. Because there is only a single one with a mix of single node, local testnet and apps to test, and it's flat and quite granular, it becomes confusing: "which setup am I looking at? I have seen genesis validators being added before... or was that the single node and this is the testnet?"

In general it's rather repetitive because of the lack of using loops, in favour of repeating the same lines 4 times. If it was me I would move the scripts out of the `Makefile.toml` into their own files separated into a single and the multi-node setup, which I think would make it much easier to follow top-to-bottom how to set up a testnet of any size. I used a separate script [here](https://github.com/consensus-shipyard/fendermint/blob/main/fendermint/testing/smoke-test/scripts/init.sh) as well. 

I would only use `cargo make` to drive the high level processes that you are trying to do, like bring up a testnet, run tests, tear it down. Using dependencies for doing one thing after another that could all go in a single script is obscruing what's going on in my experience.

I am not sure if it's possible to use Docker Compose like this:
```
docker comple -f ./fendermint/infa/docker-compose.yml \
  --env-file ./fendermint/infa/node-0.env up -d
  
docker comple -f ./fendermint/infa/docker-compose.yml \
  --env-file ./fendermint/infa/node-1.env up -d
```
That is, to keep the services separate from each other by using env vars in the service names themselves, or will Compose complain a lot about orphans. Maybe it needs some other parameter to differentiate, like `--project-name`. It would help make the script more flexible and maintainable IMO.

docs/localnet.md
dnkolegov: This network is from comet-bft docker-compose file. I decided to use the same network since the docker-compose file. See https://docs.cometbft.com/v0.37/networks/docker-compose

docs/localnet.md
dnkolegov: It is a misprint. In the docker-compose, the ports are the same for containers but different for the host machine. Fixed

docs/localnet.md
dnkolegov: You are right.

fendermint/app/src/cmd/key.rs
dnkolegov: Fixed.

fendermint/app/src/options/key.rs
dnkolegov: Fixed

fendermint/app/src/options/key.rs
dnkolegov: Fixed

fendermint/app/src/options/key.rs
dnkolegov: fixed

fendermint/app/src/options/key.rs
dnkolegov: Fixed.

fendermint/app/src/cmd/key.rs
dnkolegov: Fixed

infra/docker-compose.yml
dnkolegov: Do you mean to set FM_CHAIN_NAME to env variable $CHAIN_NAME ?

dnkolegov: Anyway, I passed that variable

infra/docker-compose.yml
```diff
@@ -0,0 +1,241 @@
...
+    container_name: fendermint-node3
+    image: "fendermint:latest"
+    build:
+      context: .
+    environment:
+      - FM_DATA_DIR=/data/fendermint/data
+      - FM_CHAIN_NAME=root
+      - LOG_LEVEL=info
+    volumes:
+      - ${HOME}/.ipc-testnet/node3:/data
```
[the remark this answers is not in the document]
dnkolegov: Fixed

infra/docker-compose.yml
dnkolegov: Fixed

infra/docker-compose.yml
dnkolegov: Fixed. That was a legacy from the cometbft docker-compose

infra/Makefile.toml
dnkolegov: Fixed

infra/Makefile.toml
```diff
@@ -0,0 +1,597 @@
...
+
+mkdir -p $BASE_TESTNET_DIR/node0/fendermint
+mkdir -p $BASE_TESTNET_DIR/node1/fendermint
+mkdir -p $BASE_TESTNET_DIR/node2/fendermint
+mkdir -p $BASE_TESTNET_DIR/node3/fendermint
+
+mkdir -p $BASE_TESTNET_DIR/node0/cometbft
+mkdir -p $BASE_TESTNET_DIR/node1/cometbft
+mkdir -p $BASE_TESTNET_DIR/node2/cometbft
+mkdir -p $BASE_TESTNET_DIR/node3/cometbft
```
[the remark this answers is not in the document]
dnkolegov: Fixed

infra/Makefile.toml
dnkolegov: Fixed

infra/Makefile.toml
dnkolegov: Fixed

infra/Makefile.toml
```diff
@@ -0,0 +1,597 @@
...
+        --public-key $BASE_TESTNET_DIR/node0/fendermint/keys/validator0_key.pk --balance ${BALANCE}
+
+./fendermint genesis --genesis-file $BASE_TESTNET_DIR/genesis.json add-account \
+        --public-key $BASE_TESTNET_DIR/node1/fendermint/keys/validator1_key.pk --balance ${BALANCE}
+
+./fendermint genesis --genesis-file $BASE_TESTNET_DIR/genesis.json add-account \
+        --public-key $BASE_TESTNET_DIR/node2/fendermint/keys/validator2_key.pk --balance ${BALANCE}
+
+./fendermint genesis --genesis-file $BASE_TESTNET_DIR/genesis.json add-account \
+        --public-key $BASE_TESTNET_DIR/node3/fendermint/keys/validator3_key.pk --balance ${BALANCE}
```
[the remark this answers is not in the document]
dnkolegov: Fixed

infra/Makefile.toml
dnkolegov: Fixed

infra/Makefile.toml
dnkolegov: Fixed

infra/Makefile.toml
dnkolegov: This is the same approach the cometbft testnet uses. Should we use a different approach?

dnkolegov:
> Thank you for putting this together and figuring out how to configure a local CometBFT testnet.
> 
> I am a little confused as to how much in the `Makefile.toml` is a copy of what I have in the smoke tests that is left in by accident. Do you want to exercise the same smoke and ethapi tests?
> 
> I'm also not sure `cargo make` is the best tool to use here. Because there is only a single one with a mix of single node, local testnet and apps to test, and it's flat and quite granular, it becomes confusing: "which setup am I looking at? I have seen genesis validators being added before... or was that the single node and this is the testnet?"
> 
> In general it's rather repetitive because of the lack of using loops, in favour of repeating the same lines 4 times. If it was me I would move the scripts out of the `Makefile.toml` into their own files separated into a single and the multi-node setup, which I think would make it much easier to follow top-to-bottom how to set up a testnet of any size. I used a separate script [here](https://github.com/consensus-shipyard/fendermint/blob/main/fendermint/testing/smoke-test/scripts/init.sh) as well.
> 
> I would only use `cargo make` to drive the high level processes that you are trying to do, like bring up a testnet, run tests, tear it down. Using dependencies for doing one thing after another that could all go in a single script is obscruing what's going on in my experience.
> 
> I am not sure if it's possible to use Docker Compose like this:
> 
> ```
> docker comple -f ./fendermint/infa/docker-compose.yml \
>   --env-file ./fendermint/infa/node-0.env up -d
>   
> docker comple -f ./fendermint/infa/docker-compose.yml \
>   --env-file ./fendermint/infa/node-1.env up -d
> ```
> 
> That is, to keep the services separate from each other by using env vars in the service names themselves, or will Compose complain a lot about orphans. Maybe it needs some other parameter to differentiate, like `--project-name`. It would help make the script more flexible and maintainable IMO.

Hi Akosh!
Thank you for thoroughly reviewing.
I believe I have fixed most of the comments. At the same time, this is not the final code because Alfonso has not looked at it. I am removing all unnecessary code from the Makefile when Alfonso approves or says what I should add for the demo.

dnkolegov:
Hey @adlrocha.
I addressed most of Akosh's comments. Please also look at the code.

docs/localnet.md
aakoshh: Well I suppose these addresses are valid on the host, no, and there I assume the ports do have to be different, even though the IPs are also unique?

infra/Makefile.toml
aakoshh: I think the CometBFT approach is good when you want something readable on the web, like in the docs you linked to. They are easy to read and doesn't require any extra knowledge like scripting, you just get the basics so there is no distraction. 

For our own infra scripts where there are not 4 but 12 containers, scripts we supposedly have to maintain and evolve, I would not stick verbatim to the CometBFT examples.

There is value in both, I know scripts can get confusing as well. I chose the scripting approach in the past so that I can create local nets of any number of nodes and for example automatically configure Prometheus to scrape all of them.

It's up to you :shrug:

infra/docker-compose.yml
aakoshh: Hm, sorry, I can't recall what I meant exactly. Probably that there exists an environment elsewhere already.

fendermint/app/src/cmd/key.rs
aakoshh: Ah, I remember, because we are inside `cmd!` macro it doesn't do formatting :bulb: 
That's why I started delegating out, to make it uniform. 

Thanks for fixing.

docs/localnet.md
```diff
@@ -0,0 +1,55 @@
...
+
+To stop the network:
+```bash
+cargo make --makefile ./infra/Makefile.toml testnet-down
+```
+
+The testnet contains four logical nodes. Each node consists of cometbft, fendermint, and ethapi containers.
+The testnet network is `192.167.10.0/24`.
+
+ETH-API is accessible on the following interfaces:
```
aakoshh, line 38: Do these now work from the host? You can use the IP and the internal port, not the exposed one? That would be pretty cool.

aakoshh (commented):
Thank you for the changes @dnkolegov , I think the for loops improved it a lot :pray: 

Let's wait then for @adlrocha to consider the more general comments.

aakoshh:
@adlrocha @dnkolegov I did a very small experiment to see if what I suggested [here](https://github.com/consensus-shipyard/fendermint/pull/222#pullrequestreview-1619616235) would work:


```console
$ tree infra
infra
├── compose.yml
├── down.sh
├── node0.env
├── node1.env
└── up.sh

$ for file in $(ls infra); do printf "\n### $file\n"; cat infra/$file;  done

### compose.yml
services:
  fendermint-node:
    container_name: fendermint-node${NODE_NR}
    image: "fendermint:latest"
    environment:
      - FM_DATA_DIR=/data/fendermint/data
      - LOG_LEVEL=info
    networks:
      - localnet

networks:
  # Network created externally so services across different projects can share it.
  localnet:
    external: true

### down.sh
#!/bin/bash

for i in {0..1}; do
  docker-compose --env-file node${i}.env --project-name node${i} down
done

docker network remove localnet

### node0.env
NODE_NR=0

### node1.env
NODE_NR=1

### up.sh
#!/bin/bash

docker network create localnet || echo "network already exists"

for i in {0..1}; do
  docker-compose --env-file node${i}.env --project-name node${i} up -d
done
``` 

Here's how you run it:

```console
$ cd infra
$ ./up.sh
191c493783905f86b5bc1a74c3a3935fc14ab174c772054ed5c0e81fda9f421a
Creating fendermint-node0 ... done
Creating fendermint-node1 ... done
$ docker ps
CONTAINER ID   IMAGE               COMMAND            CREATED         STATUS         PORTS       NAMES
0f8252a86535   fendermint:latest   "fendermint run"   4 seconds ago   Up 3 seconds   26658/tcp   fendermint-node1
a7e811838a45   fendermint:latest   "fendermint run"   5 seconds ago   Up 4 seconds   26658/tcp   fendermint-node0
$ ./down.sh
Stopping fendermint-node0 ... done
Removing fendermint-node0 ... done
Network localnet is external, skipping
Stopping fendermint-node1 ... done
Removing fendermint-node1 ... done
Network localnet is external, skipping
localnet
```

So with some helper scripts that manage the `--project-name` and `--env-file` parameters for you the repetition can be largely taken away.

infra/Makefile.toml
dnkolegov: Sorry for confusion. I did not mean scripting. I meant this:

```
I am surprised a bit that all the p2p addresses use the same port. So these IPs work internally as well as externally? Then why are the p2p ports even mapped to the host?
```

infra/docker-compose.yml
dnkolegov: Ok. I fixed that one

infra/Makefile.toml
dnkolegov: I would even aggregate them all in one task because I do not see reasons to separate them in this Makefile, but I know what you mean, and I will divide them.
Fixed

docs/localnet.md
dnkolegov: This network is accessible for containers only. You can access it from the host OS.

docs/localnet.md
dnkolegov: Fixed

docs/localnet.md
dnkolegov: Fixed. But we can only use the host interface (127.0.0.1) and the exposed ports. That will be compatible with all operating systems.

docs/localnet.md
aakoshh: Your two sentences seem to contradict each other, or I am misunderstand what you are trying to say :thinking:

infra/Makefile.toml
aakoshh: Okay, so the IPs work in the container, and they can be used with the internal ports. I am not sure why they should be mapped to the host, but why not, and I see CometBFT does the same :+1:

docs/localnet.md
aakoshh: Oh I see, so the IPs are internal too? Huh, I wonder why they bothered with IPs instead of just using the container names as hostnames (maybe it needs setting the `hostname` of the container, can't remember), which is more readable because you don't have to remember which IP range you are looking at (like 0-3 is fendermint, 10-13 is ethereum, etc, just use `ethapi-0`).

docs/localnet.md
dnkolegov: Sorry. Misprint, as always.
```
This network is accessible for containers only. You can not access it from the host OS.
```

docs/localnet.md
dnkolegov: You can do both inside the internal network. You can use `cometbft-node0`,...

dnkolegov: Fixed

adlrocha (commented):
Thank you for this, @dnkolegov. This is a huge leap towards how I was envisioning the infra management for subnets. 

A few high-level comments: 
> I would only use cargo make to drive the high level processes that you are trying to do, like bring up a testnet, run tests, tear it down. Using dependencies for doing one thing after another that could all go in a single script is obscruing what's going on in my experience.

I agree with @aakoshh here. It is quite hard to follow the `Makefile`. If one wants to add a change there it is quite hard to track what and where it should change. I was also thinking of `cargo-make` as the orchestration front-end for all the infra. We can have the high-level recipes in the `Makefile` and delegate the core logic of the scripts on independent files if that is possible. 

As discussed sync, I haven't been able to run it in Linux. I keep seeing errors when starting `cometbft` due to some permission issue, and restarting the infra leads to having to remove certain directories manually as sudo. I will keep trying to debug it myself but would love your input here. 

If we manage to refactor so the recipes are in independent scripts and the cargo-make has the high-level jobs available to deploy infra, I would be comfortably merging this so on top of that we can start having the jobs to deploy subnet nodes for specific subnets. Let me know if this makes sense. Thanks! (also, last note, I originally creates this PR, so I won't be able to approve it to merge :) )

dnkolegov:
> As discussed sync, I haven't been able to run it in Linux. I keep seeing errors when starting `cometbft` due to some permission issue, and restarting the infra leads to having to remove certain directories manually as sudo. I will keep trying to debug it myself but would love your input here.

Fixed.

dnkolegov:
> I was also thinking of `cargo-make` as the orchestration front-end for all the infra. We can have the high-level recipes in the `Makefile` and delegate the core logic of the scripts on independent files if that is possible.

Then I do not understand why we need `cargo-make` at all. Should not we switch to classic `make`? scripts is embedded mechanism of `cargo-make` and it is strange for me to use external scripts and call them from `cargo-make`. That is what `make` can do.

If I understood @aakoshh's comment right he meant the following:
- there are a lot of legacy scripts I have not removed
- `docker-compose` items can be written externally
- `cargo-make is the best tool to use here`: I do not know, maybe. For me, it works fine.

But the PR says 
> The idea is to leverage [cargo-make](https://sagiegurari.github.io/cargo-make/) to configure the deployment and trigger the end-to-end

That is what PR does right now: it configures deployment using `cargo-make`.

I do not mind clearing all the old stuff and improving `cargo-make` but I do not see how creating new scripts can help us especially when we do not have much time. Moreover, this will not affect the functionality and the final result at all.

The main question is still is that tesnet is useful and can be used or not.
@adlrocha Please check whether it works for you after my last fixes

aakoshh:
I agree that it's difficult to say when `make` is better than `cargo-make`, and how much to put into the `Makefile` or `Makefile.toml` versus scripts. 

I can only speak from experience that I put too much in the [Makefile](https://github.com/consensus-shipyard/ipc-agent/blob/main/testing/e2e/Makefile), trying to rely on variables declared at the top to do cute commands as dependencies but they were much harder to understand than a chain of script invocations with explicit parameter passing would have been. In retrospect I would have tried to keep the `Makefile` as a set of memorable commands do very high level stuff and delegate to scripts ASAP. 

The benefit of `cargo-make` to me were things like:
* [profile based overrides](https://github.com/consensus-shipyard/fendermint/blob/main/fendermint/testing/smoke-test/Makefile.toml#L20) for env vars 
* unconditional [cleanup task](https://github.com/consensus-shipyard/fendermint/blob/main/fendermint/testing/smoke-test/Makefile.toml#L42)
* Rust scripts would have been cool, but I did not need them

But I would not have attempted to do the steps I put into the [init.sh](https://github.com/consensus-shipyard/fendermint/blob/main/fendermint/testing/smoke-test/scripts/init.sh) to set up accounts, create keys, etc. That's where the Makefile in this PR started to break down for me: too many little steps that were chained together as dependencies, which reminded me of how I did the e2e tests for the agent. 

I agree with your that you want to get testing ASAP but it's also about maintainability, hopefully we don't have to go in circles forever.

infra/scripts/examples.toml
```diff
@@ -0,0 +1,16 @@
...
+  --secret-key fendermint/testing/smoke-test/test-data/fendermint/keys/alice.sk \
+  ${VERBOSITY}
+"""
+
+[tasks.ethapi-example]
+script = """
+cd ${CARGO_MAKE_WORKSPACE_WORKING_DIRECTORY}
+cargo run -p fendermint_eth_api --release --example ethers -- \
+  --secret-key-from fendermint/testing/smoke-test/test-data/fendermint/keys/emily.sk \
+  --secret-key-to   fendermint/testing/smoke-test/test-data/fendermint/keys/eric.sk
```
aakoshh, line 15: Does this setup crate these keys at all?

infra/scripts/testnet.toml
```diff
@@ -0,0 +1,184 @@
...
+
+[tasks.testnet-config]
+dependencies = [
+    "testnet-script-new-genesis",
+    "testnet-script-add-peers",
+    "testnet-script-new-key",
+    "testnet-script-new-account",
+    "testnet-script-new-gateway",
+    "testnet-script-share-genesis"
+]
```
aakoshh, line 100: I reckon it's these and [the node equivalent](https://github.com/consensus-shipyard/fendermint/pull/222/files#diff-c7acb1a1a82d3b2fa4b0b12e343395aee71560f17d0a888a933fc3f2ecea4c3cR56-R64) steps that benefit least from being in the `Makefile.toml` and chained with dependencies as individual tasks, as opposed to just one script that does it all.

But it's much better now that you have split it out into multiple `.toml` files, thanks for that!

infra/scripts/testnet.toml
```diff
@@ -0,0 +1,184 @@
...
+"""
+
+[tasks.testnet-init]
+dependencies = [
+    "testnet-clear",
+    "testnet-network-create",
+    "cometbft-pull",
+    "testnet-mkdir",
+    "testnet-cometbft-init",
+    "testnet-mkdir",
```
aakoshh, line 48: Seems like a duplicate.

infra/scripts/node.toml
```diff
@@ -0,0 +1,111 @@
...
+"""
+
+[tasks.node-script-new-gateway]
+cwd = "./target/release"
+script = """
+./fendermint genesis --genesis-file ${GENESIS_FILE} ipc gateway --subnet-id /r0 \
+    --top-down-check-period 10 \
+    --bottom-up-check-period 10 \
+    --msg-fee 10 \
+    --majority-percentage 66
```
aakoshh, line 101: Minor but strictly speaking 66% is less than 2/3.

infra/scripts/examples.toml
dnkolegov: No. I was waiting for Alfonso's on example. Removed not to add more questions.

aakoshh:
I don't see any blockers. Have you given any thought to [this suggestion](https://github.com/consensus-shipyard/fendermint/pull/222#issuecomment-1717302364) to remove the duplicated service definitions? It would make it easier to set up networks with arbitrary number of nodes.

infra/scripts/testnet.toml
dnkolegov: fixed

infra/scripts/node.toml
dnkolegov: fixed

dnkolegov:
> ```
> #!/bin/bash
> 
> docker network create localnet || echo "network already exists"
> 
> for i in {0..1}; do
>   docker-compose --env-file node${i}.env --project-name node${i} up -d
> done
> ```

I tried your suggestion. It did not work well for me when I tried it for 12 containers. The main reason is that instances are started sequentially, it took 40 seconds for me to start the testnet since we use a health check for comet-bft nodes.

In the future, we can write a simple generator and use it - https://github.com/cometbft/cometbft/blob/main/cmd/cometbft/commands/testnet.go

adlrocha:
Hey, @dnkolegov, thank you for all the changes, amazing job. Dividing the jobs into different `toml` files makes it way more readable.

> I tried your suggestion. It did not work well for me when I tried it for 12 containers. The main reason is that instances are started sequentially, it took 40 seconds for me to start the testnet since we use a health check for comet-bft nodes.

Regarding this outstanding issue, my take would be to use this approach even if it takes around a minute to bootstrap the network. We can address this in a follow up PR once we have the full end-to-end deployment of subnets fleshed out.
 either through simple code generation or using some kind of simple templating scheme as @aakoshh suggests.

At this early stage, I personally rather have non-performant but maintainable test so we don't introduce a lot of debt too soon (as was the case in the agent). So I would say, let's introduce the approach that used loops and let's get this one merged. Once I've taken it for a spin I will open tickets with potential follow-ups. WDYT?

dnkolegov:
> Hey, @dnkolegov, thank you for all the changes, amazing job. Dividing the jobs into different `toml` files makes it way more readable.
> 
> > I tried your suggestion. It did not work well for me when I tried it for 12 containers. The main reason is that instances are started sequentially, it took 40 seconds for me to start the testnet since we use a health check for comet-bft nodes.
> 
> Regarding this outstanding issue, my take would be to use this approach even if it takes around a minute to bootstrap the network. We can address this in a follow up PR once we have the full end-to-end deployment of subnets fleshed out. either through simple code generation or using some kind of simple templating scheme as @aakoshh suggests.
> 
> At this early stage, I personally rather have non-performant but maintainable test so we don't introduce a lot of debt too soon (as was the case in the agent). So I would say, let's introduce the approach that used loops and let's get this one merged. Once I've taken it for a spin I will open tickets with potential follow-ups. WDYT?

Fixed

infra/up.sh
```diff
@@ -0,0 +1,23 @@
...
+	export CMT_NODE_ADDR=192.167.10.$((${i}+2))
+	export FMT_NODE_ADDR=192.167.10.$((${i}+6))
+	export ETHAPI_NODE_ADDR=192.167.10.$((${i}+10))
+	docker compose -f ./infra/docker-compose.yml -p testnet_node_${i} up -d &
+	PORT1=$((PORT1+3))
+	PORT2=$((PORT2+3))
+	PORT3=$((PORT3+1))
+done
+
+wait $(jobs -p)
```
aakoshh: Never seen this before, very clever :+1:

adlrocha (commented):
@dnkolegov , one last fix before merging, when running it in Linux from scratch I get:  `network testnet declared as external, but could not be found
`

dnkolegov (approved):

PR #223, opened 2023-08-29
FM-216: Index transaction by Ethereum hash (take 2)

### LINKED ISSUE #216: Index transactions by Ethereum hash
Currently we [return](https://github.com/consensus-shipyard/fendermint/blob/main/fendermint/eth/api/src/apis/eth.rs#L511) the CometBFT produced transaction hash `eth_sendRawTransaction`, which is different from what Ethereum would produce as a transaction hash, because 
* for example `ethers-rs` would use the [rlp](https://github.com/gakonst/ethers-rs/blob/ethers-v2.0.9/ethers-core/src/types/transaction/eip2718.rs#L337) function to serialize the data and then [keccak256](https://github.com/gakonst/ethers-rs/blob/ethers-v2.0.9/ethers-core/src/types/transaction/eip2718.rs#L364) to hash it, followed by [rlp_signed](https://github.com/gakonst/ethers-rs/blob/ethers-v2.0.9/ethers-core/src/types/transaction/eip2718.rs#L314) to get the final binary which it sends to `eth_sendRawTransaction`
* this becomes the `data` field in the transaction, which is hashed using [Sha256](https://github.com/consensus-shipyard/fendermint/blob/main/fendermint/eth/api/src/conv/from_tm.rs#L434)

So we hash different things with different hashing algorithms. 

This was fine while working on the [ethers example](https://github.com/consensus-shipyard/fendermint/blob/main/fendermint/eth/api/examples/ethers.rs) because the Rust implementation accepts whatever `eth_sendRawTransaction` returns as the transaction hash. And we returned the Tendermint one, so that later we can use the [/tx](https://docs.tendermint.com/v0.34/rpc/#/Info/tx) RPC endpoint to fetch transaction results from CometBFT. It would not recognise the Ethereum hash. 

The problem is that ethers.js is more fussy and actually asserts that the transaction hash received after sending matches what it expects, which fails, and this assertion cannot be turned off. 

### How did others tackle this?

Lotus has an [indexer](https://github.com/filecoin-project/lotus/blob/v1.23.1-rc2/node/impl/full/eth.go#L425C2-L425C75) to turn Ethereum hashes into CIDs. 

Etheremint also has an [IndexBlock](https://github.com/evmos/ethermint/blob/v0.22.0/indexer/kv_indexer.go#L64) method and an [IndexerService](https://github.com/evmos/ethermint/blob/07cf2bd2b1ce9bdb2e44ec42a39e7239292a14af/server/indexer_service.go#L113) which seems to be running in the background, and then they first [consult this service](https://github.com/evmos/ethermint/blob/main/rpc/backend/tx_info.go#L305-L322) before falling back on the Tendermint API. When they fall back on Tendermint, they utilise the Tendermint [/tx_search](https://docs.tendermint.com/v0.34/rpc/#/Info/tx_search) endpoint which uses the kind of [filter expressions](https://docs.tendermint.com/v0.34/rpc/#/Websocket/subscribe) we use in subscriptions, combined with a custom indexed attribute, like our [emitter](https://github.com/consensus-shipyard/fendermint/blob/main/fendermint/app/src/tmconv.rs#L139-L142) which we can [filter on](https://github.com/consensus-shipyard/fendermint/blob/main/fendermint/eth/api/src/filters.rs#L133). 

Utilising CometBFT [indexing](https://docs.cometbft.com/v0.37/app-dev/indexing-transactions) seems to be the most accessible solution for the problem.

aakoshh:
Fixes #216 

Depends on #217 

I missed the fact that Lotus returned the Keccak256 hash of the full raw transaction including the signature RLP, not the sighash. Which makes sense because in Ethereum the `from` is not reified in the transaction payload, only inferred from the signature, so without the signature two transactions doing the same thing sent from different account (with the same nonce) would not be unique unless the signature was also hashed.

This PR drop the `sig.hash` field introduced in #217 and adds an `eth.hash` field instead based on a new `SignedMessage::domain_hash(&self, chain_id)` method that returns an optional `DomainHash`, with currently only the `DomainHash::Eth` variant defined. In the API we always look for the value in events, to keep it simple - if it's not present it's not an Ethereum transaction.

I opened a new PR instead of messing up the previous one because I thought it was a well rounded change, and this is easier to review as an increment.

It looks like Lotus actually has [API methods](https://github.com/filecoin-project/lotus/blob/v1.23.1-rc2/api/api_full.go#L803C2-L804C114) for CID<->Ethereum conversions, so we can potentially extend the `DomainHash` that way.

### DIFF
--- a/fendermint/eth/api/examples/ethers.rs
+++ b/fendermint/eth/api/examples/ethers.rs
@@ -46,8 +46,8 @@ use ethers_core::{
     k256::ecdsa::SigningKey,
     types::{
         transaction::eip2718::TypedTransaction, Address, BlockId, BlockNumber, Bytes,
-        Eip1559TransactionRequest, Filter, Log, SyncingStatus, TransactionReceipt, H160, H256,
-        U256, U64,
+        Eip1559TransactionRequest, Filter, Log, SyncingStatus, TransactionReceipt, TxHash, H160,
+        H256, U256, U64,
     },
 };
 use fendermint_rpc::message::MessageFactory;
@@ -405,8 +405,18 @@ where
     // However, ethers.js actually asserts this and we cannot disable it, rendering that, or any similar tool, unusable if we rely on
     // the default Tendermint transaction hash, which is a Sha256 hash of the entire payload (which includes the signature),
     // not a Keccak256 of the unsigned RLP.
-    // let expected_hash = todo!("we need to reproduce the signature to get the full RLP");
-    // assert_eq!(tx_hash, expected_hash, "Ethereum hash should match");
+
+    let expected_hash = {
+        let sig = mw
+            .signer()
+            .sign_transaction(&transfer)
+            .await
+            .context("failed to sign transaction")?;
+
+        let rlp = transfer.rlp_signed(&sig);
+        TxHash::from(ethers_core::utils::keccak256(rlp))
+    };
+    assert_eq!(tx_hash, expected_hash, "Ethereum hash should match");
 
     // Get a block with transactions by number.
     let block = request(
--- a/fendermint/eth/api/examples/ethers.rs
+++ b/fendermint/eth/api/examples/ethers.rs
@@ -405,7 +405,8 @@ where
     // However, ethers.js actually asserts this and we cannot disable it, rendering that, or any similar tool, unusable if we rely on
     // the default Tendermint transaction hash, which is a Sha256 hash of the entire payload (which includes the signature),
     // not a Keccak256 of the unsigned RLP.
-    assert_eq!(tx_hash, transfer.sighash(), "Ethereum hash should match");
+    // let expected_hash = todo!("we need to reproduce the signature to get the full RLP");
+    // assert_eq!(tx_hash, expected_hash, "Ethereum hash should match");
 
     // Get a block with transactions by number.
     let block = request(
--- a/fendermint/vm/message/src/signed.rs
+++ b/fendermint/vm/message/src/signed.rs
@@ -34,6 +34,16 @@ pub enum SignedMessageError {
     Ethereum(#[from] anyhow::Error),
 }
 
+/// Domain specific transaction hash.
+///
+/// Some tools like ethers.js refuse to accept Tendermint hashes,
+/// which use a different algorithm than Ethereum.
+///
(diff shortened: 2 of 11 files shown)


fendermint/vm/message/src/signed.rs
```diff
@@ -33,6 +34,16 @@ pub enum SignedMessageError {
     Ethereum(#[from] anyhow::Error),
 }
 
+/// Domain specific transaction hash.
+///
+/// Some tools like ethers.js refuse to accept Tendermint hashes,
+/// which use a different algorithm than Ethereum.
+///
+/// We can potentially extend this list to include CID based indexing.
+pub enum DomainHash {
```
adlrocha, line 43: I really like this abstraction, we may not use it ever, but it may also safe some headaches in the future

fendermint/eth/api/examples/ethers.rs
```diff
@@ -405,7 +405,8 @@ where
     // However, ethers.js actually asserts this and we cannot disable it, rendering that, or any similar tool, unusable if we rely on
     // the default Tendermint transaction hash, which is a Sha256 hash of the entire payload (which includes the signature),
     // not a Keccak256 of the unsigned RLP.
-    assert_eq!(tx_hash, transfer.sighash(), "Ethereum hash should match");
+    // let expected_hash = todo!("we need to reproduce the signature to get the full RLP");
```
adlrocha, line 408: Is this on purpose?

adlrocha (approved):

fendermint/vm/message/src/signed.rs
aakoshh: Yeah, I think even a list might be possible. Like a commented I saw in Lotus it's possible to retrieve a message by CID, so who knows, the need might come.

fendermint/eth/api/examples/ethers.rs
aakoshh: Yeah, I didn't want to go into the trouble of using the signer middleware to produce a signature which I can recombine into a signed RLP, but that was out of laziness, I will check now if it would be simple to do.

aakoshh: Done, thanks for keeping me honest here :pray:

PR #228, opened 2023-08-30
FM-224: implement eth_maxPriorityFeePerGas

### LINKED ISSUE #224: `eth_maxPriorityFeePerGas` not implemented for eth RPC
I was testing some one of the early scripts from Lilypad, and realized that the Python library was leveraging the `eth_maxPriorityFeePerGas` method of the RPC but it wasn't implemented in Fendermint (returning the corresponding `Method Not Allowed`). 

In Lotus here's where the [endpoint is implemented](https://github.com/consensus-shipyard/lotus/blob/156f5556b3ecc042764d76308dca357da3adfb4d/node/impl/full/eth.go#L798) and [this is the underlying implementation](https://github.com/consensus-shipyard/lotus/blob/156f5556b3ecc042764d76308dca357da3adfb4d/node/impl/full/gas.go#L187)

adlrocha:
Closes #224


fendermint/eth/api/Cargo.toml
```diff
@@ -24,6 +24,7 @@ tracing = { workspace = true }
 tendermint = { workspace = true }
 tendermint-rpc = { workspace = true }
 tokio = { workspace = true }
+reqwest = { workspace = true }
```
aakoshh, line 27: Can this just be among `[dev-dependencies]`?

fendermint/eth/api/src/apis/eth.rs
```diff
@@ -43,7 +46,12 @@ use crate::{
     error, JsonRpcData, JsonRpcResult,
 };
 
+// Lotus uses only 2 epochs to compupte the MAX_PRIO_FEE
+// but they compute the median over (on average) 10 blocks,
+// 5 per blocks. Thus, our 10 heights instead of 2.
+const NUM_BLOCKS_MAX_PRIO_FEE: u64 = 10;
 const MAX_FEE_HIST_SIZE: usize = 1024;
+const MIN_GAS_PREMIUM: u64 = 100_000;
```
aakoshh, line 54: These could be moved into the `JsonRpcState` unders some struct, so they can be added to config instead of being hardcoded.

fendermint/eth/api/src/apis/eth.rs
```diff
@@ -81,6 +89,121 @@ where
...
+    median *= BigInt::from(coeff);
+    median.div_ceil(BigInt::from(1 << PRECISION));
+
+    Ok(to_eth_tokens(&median)?)
+}
+
+// finds 55th percntile instead of median to put negative pressure on gas price
+// Rust implementation of:
+// https://github.com/consensus-shipyard/lotus/blob/156f5556b3ecc042764d76308dca357da3adfb4d/node/impl/full/gas.go#L144
+fn median_gas_premium(prices: &mut Vec<(TokenAmount, i64)>, block_gas_target: i64) -> TokenAmount {
```
aakoshh, line 177: Could you move this helper function into the [gas module](https://github.com/consensus-shipyard/fendermint/blob/main/fendermint/eth/api/src/gas/mod.rs) please? 

I wanted to keep the `eth.rs` file free of any helpers, so every method in here corresponds to an API endpoint, so we know those are here, and everything else is not here.

aakoshh (commented):
Thank you for taking on this task, I know you must have felt great fulfilment :medal_military:

fendermint/eth/api/examples/ethers.rs
```diff
@@ -752,3 +772,38 @@ where
 
     Ok(call)
 }
+
+/// Performs an `eth_maxPriorityFeePerGas` call to the ETH RPC
+/// This method is not provided by `ethers`, so we have to perform
+/// it manually.
+async fn test_max_priority_fee_call(opts: &Options) -> anyhow::Result<TokenAmount> {
```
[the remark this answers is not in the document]
adlrocha: Oh!! That's way easier to use. I wasn't aware that method was available in `ethers`, my bad for now reading the code :) Thanks!

fendermint/eth/api/Cargo.toml
adlrocha: Agree. Fortunately we don't even need it per your comment above.

fendermint/eth/api/src/apis/eth.rs
adlrocha: Sure, I didn't see the module there, that makes sense!

fendermint/eth/api/src/apis/eth.rs
adlrocha: Moved!

aakoshh (approved):
Very nice!

PR #229, opened 2023-08-30
FM-225: Query pending actor state

### LINKED ISSUE #225: Take mempool into account when querying pending state
Our `eth_getTransactionCount` method doesn't take transactions in the mempool into account.  The client script isn't using the ethers `NonceManager`; the result is when it sends multiple transactions in a quick succession they arrive with the same nonce.

https://www.quicknode.com/docs/ethereum/eth_getTransactionCount has a `blockNumber` parameter which can be `pending`, indicating that these should be treated.

aakoshh:
Fixes #225 and #226 

Added a `pending: bool` field to `FvmQuery::ActorState` which, if `true`, causes the interpreter to look into the `CheckState` maintained by the `App`, so that Hardhat with ethers.js can continue not using a `NonceManager` and still submit multiple requests in a quick succession. 

To exercise it, a new `QueryClient::pending_state` method has been added, which takes no `height` parameter. The reason it's a new method, even though it's piggybacking on the existing `ActorState` query which is available as `QueryClient::actor_state`, is that it involves locking the check state `Mutex`, which means there can only ever be one query like this running and it actually blocks transaction submission as well, which is not nice. It's not available in the CLI, but it's used by the Ethereum API in `eth_getTranasactionCount` and `eth_getBalance` _iff_ the block ID is `pending`. 

To avoid having to take a lock we'd need to clone the `StateTree` inside `FvmCheckState`, but that's not possible.


adlrocha (approved):
Thank you for this one. Once we merge it I will test if it works with hardhat finally :)

PR #230, opened 2023-08-30
FM-227: Return error message from nonce and balance checks

### LINKED ISSUE #227: Return message with error in `check_tx` and `deliver_tx`
We return `None` instead of returning some message with the error to give additional context in `check_tx` and `deliver_tx`: https://github.com/consensus-shipyard/fendermint/blob/97a202062373986b43f81290030734ca523d609e/fendermint/vm/interpreter/src/fvm/check.rs#L65

It would be good if we can add an error message to make our lives easier while debugging (we sometimes have to check the error code returned and figure out what could be happening).

aakoshh:
Fixes #227


adlrocha (approved):
LGTM. Thanks!

PR #231, opened 2023-08-31
Prepare and process top down checkpoint proposal

cryptoAtwill:
Implementing prepare and process top down checkpoint proposal.


fendermint/app/config/default.toml
```diff
@@ -50,3 +52,12 @@ host = "127.0.0.1"
...
 port = 8545
+
+# IPC Parent finality related configuration
+[parent_finality]
+# The number of blocks behind chain head to propose parent finality.
+chain_head_delay = 10
+# The number of seconds to polling ipc agent for latest parent view
+polling_interval_secs = 10
+# The ipc agent url
+ipc_agent_url = "http://0.0.0.0:3030/json_rpc"
```
aakoshh, line 63:
```suggestion
ipc_agent_url = "http://127.0.0.1:3030/json_rpc"
```

Although both work, I think this is more idiomatic; I only see 0.0.0.0 as a binding address for starting servers.

fendermint/app/src/cmd/run.rs
```diff
@@ -21,6 +23,15 @@ cmd! {
...
 }
 
+async fn create_parent_finality(_settings: &Settings) -> anyhow::Result<InMemoryFinalityProvider> {
+    // let provider = InMemoryFinalityProvider::new(
+    //     settings.parent_finality.clone(),
+    //     unimplemented!(),
+    // );
+    // Ok(provider)
+    todo!()
+}
```
aakoshh, line 33: Maybe you could have a default implementation which does nothing, for the case when the node is the root?

fendermint/app/src/settings.rs
```diff
@@ -68,6 +71,10 @@ pub struct Settings {
     pub db: DbSettings,
     pub eth: EthSettings,
     pub fvm: FvmSettings,
+
+    #[serde(deserialize_with = "deserialize_subnet_id", default)]
+    pub subnet_id: SubnetID,
+    pub parent_finality: fendermint_vm_topdown::Config,
```
aakoshh, line 77: Let's create some `IpcSettings` for these.

fendermint/app/config/default.toml
```diff
@@ -50,3 +52,12 @@ host = "127.0.0.1"
 # The default port where the Ethereum API will listen to
 # JSON-RPC (POST) and WebSockets (GET) requests.
 port = 8545
+
+# IPC Parent finality related configuration
+[parent_finality]
```
aakoshh, line 57: Let's put these under an `[ipc]` section.

fendermint/app/src/settings.rs
```diff
@@ -68,6 +71,10 @@ pub struct Settings {
     pub db: DbSettings,
     pub eth: EthSettings,
     pub fvm: FvmSettings,
+
+    #[serde(deserialize_with = "deserialize_subnet_id", default)]
```
aakoshh, line 75: What does the `default` do? If there is a default here, why do we need one in the `default.toml` file?

fendermint/app/src/settings.rs
```diff
@@ -145,6 +152,29 @@ pub fn expand_tilde<P: AsRef<Path>>(path: P) -> PathBuf {
...
+        }
+
+        fn visit_str<E>(self, v: &str) -> std::result::Result<Self::Value, E>
+        where
+            E: serde::de::Error,
+        {
+            SubnetID::from_str(v).map_err(E::custom)
+        }
+    }
+    deserializer.deserialize_str(SubnetIDVisitor)
```
aakoshh, line 175:
```suggestion
    let s = String::deserialize(deserializer)?;
    s.parse::<SubnetID>().map_err(|e| D::Error::custom(e))
```

Requires `use serde::de::Error;` at the top. 

I don't see the visitor pattern very often, I see it's powerful, but maybe for this case a map over a known type is more concise?

fendermint/vm/interpreter/src/chain.rs
```diff
@@ -89,32 +97,72 @@ where
...
+                    height: proposal.height as ChainEpoch,
+                    block_hash: proposal.block_hash,
+                },
+            ))),
+            Ok(None) => {
+                tracing::debug!("parent finality proposal not ready yet");
+            }
+            Err(e) => {
+                // safe to unwrap as the type is correct
+                let e = e.downcast_ref::<fendermint_vm_topdown::Error>().unwrap();
```
aakoshh, line 113: But it is possible to use a type hint to the compiler to avoid having to cast and unwrap, right? It can be done either by assigning to a variable, or by calling it like `atomically_or_err<_, fendermint_vm_topdown::Error, _, _>(...)`.

fendermint/vm/interpreter/src/chain.rs
```diff
@@ -89,32 +97,72 @@ where
...
+        // Prepare top down proposals
+        match atomically_or_err(|| finality_provider.next_proposal()).await {
+            Ok(Some(proposal)) => msgs.push(ChainMessage::Ipc(IpcMessage::TopDownProposal(
+                ParentFinalityProposal {
+                    height: proposal.height as ChainEpoch,
+                    block_hash: proposal.block_hash,
+                },
+            ))),
+            Ok(None) => {
+                tracing::debug!("parent finality proposal not ready yet");
```
aakoshh, line 109: Or maybe this node is root and it will never be ready, in which case this log would be just noise.

From what I understand the default polling of 10s means that with the default 2s block finality we will see this quite often. Maybe it could live inside the provider, instead of here, as it's quite normal.

fendermint/vm/interpreter/src/chain.rs
```diff
@@ -89,32 +97,72 @@ where
...
+                            block_hash: block_hash.clone(),
+                        })
+                    })
+                    .await)
+                        .is_ok()
+                    {
+                        Ok(true)
+                    } else {
+                        Ok(false)
+                    };
```
aakoshh, line 162:
```suggestion
                    let prop = IPCParentFinality {
                            height: height as u64,
                            block_hash,
                    };
                    if atomically_or_err(|| state.1.check_proposal(&prop)).await.is_err() {
                      return Ok(false);
                    }
```
I think this is better because:
* It's more readable IMO. 
* You can create the prop once, because you only need a reference inside the tranaction, so you don't need to clone. (I think)
* You returned from the loop in both positive and negative outcomes. As you see from the bottom-up case, there can be other messages we need to check, so we can't return `true` just because the top-down is correct.

fendermint/vm/message/src/ipc.rs
```diff
@@ -25,8 +25,9 @@ pub enum IpcMessage {
     /// or we can gossip votes using the _IPLD Resolver_ and attach them as a quorum certificate.
     BottomUpExec(CertifiedMessage<BottomUpCheckpoint>),
 
-    // TODO
-    TopDown,
+    /// A top-down checkpoint parent finality proposal. This proposal should contain the latest parent
+    /// state that to be checked and voted by validators.
+    TopDownProposal(ParentFinalityProposal),
```
aakoshh, line 30:
```suggestion
    TopDownExec(ParentFinalityProposal),
```

To be symmetric with `BottomUpExec`, we are proposing that this thing gets executed because it is final.

fendermint/vm/message/src/ipc.rs
```diff
@@ -93,9 +94,19 @@ pub struct BottomUpCheckpoint {
     pub bottom_up_messages: Cid, // TODO: Use TCid
 }
 
+/// A proposal of the parent view that validators will be voting on.
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq, Hash)]
+pub struct ParentFinalityProposal {
```
aakoshh, line 99:
```suggestion
pub struct ParentFinality {
```

I think we don't need to emphasize here that it's a proposal, after all everything in the block is a proposal at some point. What we want to express is that we view some height and hash as final, and this already says it in a way.

aakoshh (commented):
This looks great, I only have a few minor remarks :)

adlrocha (commented):
I completely forgot to send my comments on Friday. Apologies if they are outdated by now

fendermint/app/src/cmd/run.rs
```diff
@@ -51,6 +62,7 @@ async fn run(settings: Settings) -> anyhow::Result<()> {
         state_store,
         interpreter,
         resolve_pool,
+        parent_finality.clone(),
```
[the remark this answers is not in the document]
cryptoAtwill: yep

fendermint/app/src/settings.rs
cryptoAtwill: This was copy pasted, should have removed the `default`

fendermint/app/src/cmd/run.rs
cryptoAtwill: yep, it's added

fendermint/app/config/default.toml
```diff
@@ -4,6 +4,8 @@ data_dir = "data"
 contracts_dir = "contracts"
 # Builtin actor bundle path relative to the `--home-dir`, unless given as an absolute path.
 builtin_actors_bundle = "bundle.car"
+# Subnet id assigned to this fendermint node
+subnet_id = "/r1024"
```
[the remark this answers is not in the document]
cryptoAtwill: I used the value for `fendermint` chain id

fendermint/app/src/settings.rs
```diff
@@ -144,6 +151,29 @@ pub fn expand_tilde<P: AsRef<Path>>(path: P) -> PathBuf {
         .unwrap_or(p)
 }
 
+/// A serde deserialization method to deserialize a subnet path string into a [`SubnetID`].
+pub(crate) fn deserialize_subnet_id<'de, D>(deserializer: D) -> anyhow::Result<SubnetID, D::Error>
```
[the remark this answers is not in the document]
cryptoAtwill: yeah, but just one method, should be ok for now.

fendermint/vm/interpreter/src/chain.rs
cryptoAtwill: nice nice, thx for the tip, was not aware of this

fendermint/app/src/settings.rs
```diff
@@ -67,6 +70,10 @@ pub struct Settings {
     pub db: DbSettings,
     pub eth: EthSettings,
     pub fvm: FvmSettings,
+
+    #[serde(deserialize_with = "deserialize_subnet_id", default)]
+    pub subnet_id: SubnetID,
```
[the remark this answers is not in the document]
cryptoAtwill: I think we can do that in config toml

fendermint/app/config/default.toml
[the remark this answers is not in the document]
aakoshh: What do you mean, where is this default value defined? I might have just forgotten about it, but it looks arbitrary.

fendermint/vm/interpreter/src/chain.rs
aakoshh: Sorry, I mentioned it in the description of  https://github.com/aakoshh/async-stm-rs/pull/1 which I sent to you, and then there is an example of the above at https://github.com/aakoshh/async-stm-rs/pull/1/files#diff-b613865b220392b32225c6578f96bfe2418a6d4e8d8c4c63ee664dceb7e60719R102

fendermint/app/config/default.toml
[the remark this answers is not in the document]
aakoshh: @adlrocha anything you pass in the config can be passed as an environment variable, such as `FM_IPC__SUBNET_ID`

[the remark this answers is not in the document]
aakoshh: Even `SubnetID::UNDEF` has 0, which would make it easy to detect that this setting has not been defined.

fendermint/vm/interpreter/src/chain.rs
cryptoAtwill: I removed this `debug`.

fendermint/app/config/default.toml
```diff
@@ -56,3 +58,15 @@ host = "127.0.0.1"
 # The default port where the Ethereum API will listen to
 # JSON-RPC (POST) and WebSockets (GET) requests.
 port = 8545
+
+# IPC Parent finality related configuration
+[ipc]
+subnet_id = "/r1942764459484029"
```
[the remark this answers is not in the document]
aakoshh: And if we _were_ to include such magic numbers, there should be a comment about in the config file explaining how we got it and why it's guaranteed that the user will need to overwrite this. 

Like I said in my [initial comment](https://github.com/consensus-shipyard/fendermint/pull/231/files#r1312943016) aimed at @adlrocha  I think this should dynamically come from the `chain_name` which we use in Genesis, combined with the parent subnet ID. The `chain_name` is also an input to every RPC method on the CLI, so we don't have to remember numbers like this.

fendermint/vm/interpreter/src/chain.rs
```diff
@@ -89,32 +97,65 @@ where
...
+                    if !is_resolved {
+                        return Ok(false);
+                    }
                 }
-            }
+                ChainMessage::Ipc(IpcMessage::TopDownExec(ParentFinality {
+                    height,
+                    block_hash,
+                })) => {
+                    let prop = IPCParentFinality {
```
aakoshh, line 146: Do we need both `ParentFinality` and `IPCParentFinality` to exist?

cryptoAtwill: I think it's just for crate completeness. Initially, `fendermint_vm_topdown` defined `IPCParentFinality` which should be imported into `fendermint_vm_message`. I think you mentioned once `fendermint_vm_message` should be kept thin and avoid import heavy crates such as `tokio` in `fendermint_vm_topdown`. But `fendermint_vm_topdown` import `message` seems strange to me. Since the `IPCParentFinality` is pretty light, so I just created a duplicate as a compromise.

fendermint/app/config/default.toml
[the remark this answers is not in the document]
aakoshh: Please let's use 0, the ID of calibration is meaningless to us, it's a Lotus network, and unless I'm mistaken this is the ID of _this_ subnet, not the parent.

fendermint/app/config/default.toml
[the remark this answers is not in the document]
cryptoAtwill: I think @adlrocha also brought this up too, a better way would be reading from genesis to avoid wrongly configured parameters.

fendermint/app/config/default.toml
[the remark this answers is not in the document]
aakoshh: Okay so as we agreed in chat:
* This can be zero (a.k.a. UNDEF) here 
* Signature checks are based on `chain_id`, which is a hash of `chain_name`, which comes from Genesis and while it's no surprise, it's currently not duplicated elsewhere in config; it also becomes the Tendermint Chain ID. 
* The subnet ID is assigned based on the subnet actor ID and won't be known until the subnet is created in the parent.
* In the Eudico Proof-of-Concepts we set `chain_name := subnet_id`. 
* We can put that into the `FM_CHAIN_NAME` env var to make it accessible to the CLI commands which use `--chain-name`, but save people the trouble of having to type it in.

fendermint/vm/interpreter/src/chain.rs
aakoshh: I have imported the `fendermint_vm_message` crate everywhere, but okay, if you prefer an almost identically named copy. Who knows it might have its own life.

aakoshh (approved):
Thank you for following through with all the comments :v:

fendermint/app/config/default.toml
```diff
@@ -56,3 +56,17 @@ host = "127.0.0.1"
...
+[ipc]
+# Subnet id assigned to this fendermint node. Update this to the correct subnet id.
+# TODO: fetch this value from genisis if possible.
+subnet_id = "/r0"
+
+[ipc.config]
+# The number of blocks behind chain head to propose parent finality.
+chain_head_delay = 10
+# The number of seconds to polling ipc agent for latest parent view
+polling_interval_secs = 10
```
aakoshh, line 70: FYI I am using `#[serde_as(as = "DurationSeconds<u64>")]` and `Duration`, although I'm not sure which one is better for people using the config file.

fendermint/vm/interpreter/src/chain.rs
```diff
@@ -89,32 +97,65 @@ where
             CheckpointPoolItem::BottomUp(ckpt) => ChainMessage::Ipc(IpcMessage::BottomUpExec(ckpt)),
         });
 
+        // Prepare top down proposals
+        match atomically_or_err::<_, fendermint_vm_topdown::Error, _>(|| {
+            finality_provider.next_proposal()
+        })
+        .await?
```
aakoshh, line 104: If this fails this will crash the node.

cryptoAtwill: Yep, I encountered this during integration testing. I have updated the logic slightly. It will absorb the error due to:
- can always retry in the next block
- other nodes might propose and process successfully

PR #232, opened 2023-08-31
fix: getBlock null bloom

adlrocha:
This PR: 
- Fixes an error for which `logs_bloom` was being set to `null` when serializing the response for `eth_getBlockBy...` which made some tooling to crash in the deserialization of the response because it expects a mandatory `logs_bloom` field.
- Improves logging for debugging purposes.


aakoshh (approved):
:+1:

PR #233, opened 2023-08-31
fix: additional overestimation to estimateGas

adlrocha:
We found a bug for which when several subsequent messages are being sent, sometimes the gas estimation returned by the gas estimation is exactly the `gas_used` producing transactions to fail. I am not proud of this, but an easy way to fix it is to add an additional overestimation after the `gas_search`. This is not critical, the only impact that it has is that less transactions will fit a block (as the block limit will be reached earlier), and users will have to put more funds upfront for the execution. However, once the message is executed, the FVM will return the funds for the outstanding gas that wasn't used in the execution of the message.

This should be a temporary fix until we come up with a better strategy to estimate gas more accurately in a way it that doesn't return `gas_search == gas_used` making transactions to potentially run out of gas. 

Additionally, subnet operators can always reduce as much as they want the overestimation used in the config as this parameter is configurable. 

> TODO: Add follow-up issue.

### DIFF
--- a/fendermint/vm/interpreter/src/fvm/query.rs
+++ b/fendermint/vm/interpreter/src/fvm/query.rs
@@ -78,14 +78,20 @@ where
                 // Populate gas message parameters.
                 let est = match self.estimate_gassed_msg(&state, &mut msg)? {
                     Some(ret) => {
-                        // return immediately if there is something is returned,
+                        // return immediately if something is returned,
                         // it means that the message failed to execute so there's
                         // no point on estimating the gas.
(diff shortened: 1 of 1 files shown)


adlrocha:
@aakoshh I feel all of this debugging is taking a toll on us and we should make some progress on the cross-net primitives and the next milestone. How comfortable would you be to leave this temporary fix and track an issue to explore better strategies in the future?

aakoshh (approved):
I'm okay with this, whatever works!

PR #234, opened 2023-08-31
Fixes to make the Go example work

aakoshh:
Trying to get the Go example at https://github.com/binocarlos/web3-events-example to work

Using the following steps to exercise it:

```bash
cd go

SMOKE_DIR=$PWD/../../fendermint/fendermint/testing/smoke-test
cd $SMOKE_DIR && cargo make setup && cd -

export PRIVATE_KEY=$(cat $SMOKE_DIR/test-data/fendermint/keys/eric.sk | base64 -d | xxd -p | tr -d '\n')
export RPC_URL=ws://localhost:8545
export CHAIN_ID=$(docker exec -it smoke-fendermint sh -c 'fendermint rpc --url http://smoke-cometbft:26657 query state-params' | jq .response.value.chain_id)

go run events.go
```

Changes:
- [x] The script did not wait for the deployment to finalize before subscribing to events, it used the predictable ethereum address to subscribe. The call failed when I tried to convert this into an actor ID, because the transaction that would have allocated an ID has not run yet. Changed the whole thing so we index an `emitter.deleg` field with the delegated address, and the `emitter.id` for the actor address. This way the API does not have to do any conversions.
- [x] Optimisations to reduce the number of queries going to the API by not looking up headers just to get the latest height before queries. 
- [x] The test script wants to call the deployed contract and for that it looks like an `eth_getCode` request is sent _before_ the contract is deployed, resulting in an error because I think the ref-fvm creates a placeholder actor when it sees that the receiver is an f410 address, but it can't handle the `Bytecode` method obviously. The result is that a completely empty result is returned, instead of what would be the IPLD encoding of an empty byte vector, which then fails to deserialize. I'm not sure what to about that; waiting a few seconds before starting with sending transactions works, but need to check what happens in Lotus. One thing we could do here is to use the full `FvmExecState` instead of `FvmCheckState` in he `check` method, so that we have a complete pending state, rather than just adjusted nonces and balances. I [see from the Go code](https://github.com/ethereum/go-ethereum/blob/92b8f28df3255c6cef9605063850d77b46146763/accounts/abi/bind/base.go#L144) that some backends support this, others don't. The fact that we execute transactions for gas estimation suggests we might as well execute them during check, with the sad caveat that it creates even more of a bottleneck on this locked state.
- [x] The script does not report events. I can see the events arriving from CometBFT if I run `docker logs smoke-ethapi 2>&1 | grep "tx.height"` and then sent to the websocket by running `docker logs smoke-ethapi 2>&1 | grep "eth_sub"`, yet there is nothing in the Go tester. The reason was that Go expected the method to be `eth_subscription`, not `eth_subscribe`. 

I saw this on a positive test run:
```
$ export PRIVATE_KEY=$(cat $SMOKE_DIR/test-data/fendermint/keys/eric.sk | base64 -d | xxd -p | tr -d '\n'); go run events.go
Contract deloyed: 0x307861613743613834384631433439393534326163664437364665624532303161386562303544413942
tx: 0x810f55711da06537161f2d8b9a2f32ef15fd41e65f2ea4290c4716ae822f0f50
saw event [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 1]
tx: 0x29453f84149543e13e7147c954bbe8f7626feb7570eec319bab86a1def48ba92
saw event [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 2]
tx: 0x838179d2940adeb947405e25ed8b498f54aaa4e3f1e7209c45ea299e161afb81
saw event [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 3]
tx: 0x6341eb3fd4c8ab663f1cb839adb69dc30f85fb117c994fb9943597d26f46839f
saw event [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 4]
tx: 0x058f0e680bb2653340aaafccbe4acfb5a36cdf85ffadf1dc8adcaf46e00b0e34
saw event [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 5]
tx: 0x82c2b9cde12d333a327445e11a7a343758602fbaa564d787f431d03cfb1e770e
saw event [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 6]
^Csignal: interrupt
```


adlrocha (approved):
Overall LGTM, and after testing it seems to work fine. Amazing job, thanks!

Issue #173, opened 2023-08-02
IPC: Bottom-up checkpoints and message execution


adlrocha:
As discussed in the CoLo, Fendermint gives us way more flexibility than Lotus to build bottom-up checkpoints properly. The proposal is to leverage Tendermint checkpoints for the implementation of bottom-up checkpoints in order to already include the multisig required for their propagation and verification in the parent.

The high-level operation would be the following (writing it as a task list so each can be extracted --if needed-- as an issue to track the implementation).

- [ ] Every validator will construct a new checkpoint every `bottomup_check_period` (this value is in the subnet actor for the subnet in the parent). The checkpoint is built from the information available for the validator in the ledger at `end_block` of the corresponding height for the checkpoint (or `propose_block` of `height + 1). The checkpoint should have (at least) the following structure:
```
pub struct BottomUpCheckpoint {
   pub height: u64
   pub proof: Vec<u8> //whatever state info we want to propagate
   pub cross_msg_fee: TokenAmount  // cross-msg fees being distributed
   pub config: MembershipSet // info about the next membership?
   pub cross_msgs: Vec<CrossMsg>  // cross-messages that can be executed
}
```

- [ ] As part of the checkpoint generation logic, all nodes in the system at the corresponding height insert the checkpoint into the ledger  (Cron-like behavior, we are all going to do it so if we are honest we won't reach an inconsistent state).

- [ ] In order to build the multisig required for the submission in the parent, validators will broadcast a transaction in the next block voting for that checkpoint. These votes will be included in a block (or a set of blocks). When a majority of votes  are conveniently executed, a new even is emitted signalling that the multisig has been successfully finished, and that it can be submitted in the parent. This requires changes in the solidity contracts for:
   - [ ] Collecting the votes from the validators to include the multisig in the child subnet ledger and emit the corresponding event.
   - [ ] Insert the checkpoint into the ledger.

- [ ] Once the event is emitted, IPC agents will be listening to this event to submit the multisig with the checkpoint to the parent subnet. This will also require changes to the contract to:
   - [ ]  Accept a checkpoint with its corresponding multisig and trigger its commitment in the parent, the execution of the messages, and the reward to the IPC agent relaying the contract, and potentially the validators involved in the propagation

PR #236, opened 2023-09-04
Commit parent finality

cryptoAtwill:
Commit top down parent finality. The key idea is:
- launching a polling parent view syncer that runs when the app starts. The polling parent view syncer always pulls parent view from ipc agent in the background and stores the data in memory. 
- `deliver_tx` detects a `TopDownProposal` IPC message and executes the `commitParentFinality` in the gateway contract through implicit message call.

The key changes are:
- Update `Setting` to include parent finality related configurations.
- Update `ipc_actor` to include latest `gateway_getter_facet` and `gateway_router_facet` contracts.
- Update `ChainMessageInterpreter` to include execution of `IPCMessage::TopDownProposal`, changing the `ChainMessageInterpreter::State` to include `InMemoryParentFinality`.
- Introduce `ParentFinalityQuery` to read the latest committed parent finality from state.
- Update fendermint `run` cmd to launch parent finality related background tasks.


rust-toolchain.toml
```diff
@@ -1,4 +1,4 @@
 [toolchain]
-channel = "stable"
+channel = "nightly"
```
cryptoAtwill, line 2: somehow I need to update to `nightly` so that one of the dependencies can be built. Not sure if this happens to others?

fendermint/app/src/ipc.rs
```diff
@@ -0,0 +1,149 @@
...
+        )
+        .context("error creating execution state")?;
+
+        let msg = get_parent_finality_to_fvm(block_height)?;
+        let data = exec_state
+            .execute_implicit(msg)?
+            .msg_receipt
+            .return_data
+            .to_vec();
+        let decoded = decode_fevm_invoke(data)?;
```
cryptoAtwill, line 89: I feel we can create some common util crate like this so that we dont duplicate code

rust-toolchain.toml
aakoshh: Which dependency is that? Can we find another one, or use another version of it?

fendermint/app/src/ipc.rs
aakoshh: It's in the RPC crate and it's public, you definitely should not have to duplicate this. I am using these methods i[n the Ethereum API](https://github.com/consensus-shipyard/fendermint/blob/main/fendermint/eth/api/src/apis/eth.rs#L639) too.

aakoshh: In general, everything in the RPC crate is public and it's there to help deal with these message formats.

fendermint/app/src/cmd/run.rs
```diff
@@ -50,7 +61,26 @@ async fn run(settings: Settings) -> anyhow::Result<()> {
...
+        ns.app.clone(),
+    );
+    let parent_finality = Arc::new(create_parent_finality(&settings, &parent_finality_getter)?);
+    let ipc_agent_proxy = create_ipc_agent_proxy(&settings)?;
+    let polling_parent_syncer = PollingParentSyncer::new(
+        settings.parent_finality.clone(),
+        parent_finality.clone(),
+        Arc::new(ipc_agent_proxy),
+    );
+    polling_parent_syncer.start()?;
```
adlrocha, line 82: Can you remind me how are errors handled by this routine? 

Also, we should be careful once we introduce the ability to sync a node from scratch, as we may start fetching parent finality that is outdated while our peer is still syncing with the chain.

fendermint/vm/interpreter/src/chain.rs
```diff
@@ -221,32 +238,51 @@ where
...
+                    let (state, ret) = self
+                        .inner
+                        .deliver(state, VerifiableMessage::NotVerify(msg))
+                        .await?;
+
+                    downcast_err!(
+                        atomically_or_err(|| provider.on_finality_committed(&finality)).await
+                    )?;
+
+                    Ok(((pool, provider, state), ChainMessageApplyRet::Signed(ret)))
```
adlrocha, line 266: Can you point me where we send the message to commit the parent finality?

adlrocha (commented):
I think I would feel more comfortable if @aakoshh leads this review. Conceptually makes sense, I am just missing the place where the finality commitment is actually executed (along with the message execution, etc. as we discussed on Friday). 

There is also a type conversion dance, we should consider how to handle these so we don't have a lot of duplicate code in the agent, here, etc.

fendermint/app/src/app.rs
```diff
@@ -64,11 +64,11 @@ pub enum AppError {
 #[derive(Serialize, Deserialize)]
 pub struct AppState {
     /// Last committed block height.
-    block_height: BlockHeight,
+    pub(crate) block_height: BlockHeight,
```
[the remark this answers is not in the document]
cryptoAtwill: I can create `into_inner` that returns the inner types. With `getter` I have to clone, which I want to avoid.

fendermint/app/src/cmd/run.rs
cryptoAtwill: For now, it would just catch the error, and retry execution.

fendermint/vm/topdown/src/convert.rs
```diff
@@ -0,0 +1,172 @@
...
+use std::marker::PhantomData;
+use std::str::FromStr;
+
+const COMMIT_PARENT_FINALITY_FUNC_NAME: &str = "commitParentFinality";
+const GET_PARENT_FINALITY_FUNC_NAME: &str = "getParentFinality";
+
+impl From<IPCParentFinality> for gateway_router_facet::ParentFinality {
+    fn from(value: IPCParentFinality) -> Self {
+        let mut array = [0u8; 32];
+        array.copy_from_slice(&value.block_hash[0..32]);
```
aakoshh, line 23: Would it make sense to use `TryFrom` and fail if the hash isn't 32 bytes?

fendermint/vm/topdown/src/convert.rs
```diff
@@ -0,0 +1,172 @@
...
+        Payload::Delegated(d) => {
+            let addr = d.subaddress();
+            let b = ethers::abi::encode(&[Token::Tuple(vec![
+                Token::Uint(U256::from(d.namespace())),
+                Token::Uint(U256::from(addr.len())),
+                Token::Bytes(addr.to_vec()),
+            ])]);
+            ethers::types::Bytes::from(b)
+        }
+        _ => unimplemented!(),
```
aakoshh, line 55:
```suggestion
        _ => unimplemented!("unexpected payload type"),
```
Just in case?

fendermint/vm/topdown/src/convert.rs
```diff
@@ -0,0 +1,172 @@
...
+        }
+
+        let data = ethers::contract::encode_function_data(
+            commit_function,
+            gateway_router_facet::CommitParentFinalityCall {
+                finality: gateway_router_facet::ParentFinality::try_from(finality)?,
+                validators: addresses,
+                weights,
+            },
+        )?;
```
aakoshh, line 116: I can't see the configuration ID going into the contract.

fendermint/vm/topdown/src/convert.rs
```diff
@@ -0,0 +1,172 @@
...
+
+/// The decode function call return value in solidity
+pub struct DecodeFunctionReturn<T> {
+    _phantom: PhantomData<T>,
+}
+
+/// The equivalent of abi.encodeWithSignature in solidity
+pub struct EncodeWithSignature<T> {
+    _phantom: PhantomData<T>,
+}
```
aakoshh, line 74: This might make sense to someone familiar with Solidity, but I find the appearance of "Signature" confusing.

```
    let params = EncodeWithSignature::<
        gateway_router_facet::CommitParentFinalityCall,
    >::encode(finality, validator_set)?;
```

given the complexity of these facets, I'd think this crate should try to hide it and make it simpler to use, rather than make it required knowledge for something like the ChainMessageInterpreter to come up with the right combination of a confusing trait name generic over an ever shifting land of facets and their method groupings. 

A simple method like `encode_commit_parent_finality_call` would provide less flare but be way easier to discover. It doesn't look like there's much commonality between these traits anyway apart from being implemented on calls, but like I said since they are very difficult to discover, you have to know which one lives in which Solidity contract, I would try to hide them at all costs.

aakoshh (commented):
Big push :muscle:

fendermint/app/src/app.rs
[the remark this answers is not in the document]
aakoshh: That's a very micro optimisation concern, worrying about cloning a u64

fendermint/vm/interpreter/src/chain.rs
```diff
@@ -168,6 +173,18 @@ where
...
+    ($r:expr) => {
+        match $r {
+            Ok(v) => Ok(v),
+            Err(e) => match e.downcast_ref::<fendermint_vm_topdown::Error>() {
+                None => unreachable!(),
+                Some(e) => Err(e.clone()),
+            },
+        }
+    };
+}
```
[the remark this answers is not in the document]
cryptoAtwill: This can be removed. as `main` now has `async-stm` 0.4.

fendermint/vm/interpreter/src/chain.rs
```diff
@@ -221,32 +238,51 @@ where
...
+                    let validator_set = downcast_err!(
+                        atomically_or_err(|| provider.validator_set(p.height as u64)).await
+                    )?
+                    .ok_or_else(|| anyhow!("cannot find validator set for block: {}", p.height))?;
+
+                    let finality = IPCParentFinality {
+                        height: p.height as u64,
+                        block_hash: p.block_hash,
+                    };
+                    let msg = parent_finality_to_fvm(finality.clone(), validator_set)?;
```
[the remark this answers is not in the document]
cryptoAtwill: cross messages are executed after finality is committed, aka in the next PR.

fendermint/app/src/cmd/run.rs
```diff
@@ -23,13 +24,23 @@ cmd! {
...
-    //     settings.parent_finality.clone(),
-    //     unimplemented!(),
-    // );
-    // Ok(provider)
-    todo!()
+fn create_parent_finality(
+    settings: &Settings,
+    query: &ParentFinalityQuery<RocksDb, NamespaceBlockstore, AppStore>,
+) -> anyhow::Result<InMemoryFinalityProvider> {
+    let last_committed_finality = query.get_committed_finality()?;
```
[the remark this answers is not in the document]
cryptoAtwill: The other place where this could be called would be when a chain reorg is detected in the parent. In this case, we have to restart, that way this query needs to be called again. But how to do that, at this point, is not 100% clear yet. I think having the struct might be easier for future PRs.

rust-toolchain.toml
cryptoAtwill: seems in github action, it's ok. Maybe it's my local. I will revert this change.

fendermint/vm/interpreter/src/chain.rs
cryptoAtwill: discussed this over slack.

fendermint/app/src/ipc.rs
```diff
@@ -0,0 +1,149 @@
...
+
+/// Decode fvm return base64 encoded string bytes
+fn decode_data(data: Vec<u8>) -> anyhow::Result<Vec<u8>> {
+    let b64 =
+        String::from_utf8(data).context("error parsing data as base64 string in ipc finality")?;
+    let data = base64::engine::general_purpose::STANDARD
+        .decode(b64)
+        .context("error parsing base64 to bytes in ipc finality")?;
+    Ok(data)
+}
```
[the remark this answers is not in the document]
cryptoAtwill: Added, I also added a function to handle `bytes` directly as well.

fendermint/app/src/ipc.rs
```diff
@@ -0,0 +1,149 @@
...
+        let block_height = app_state.block_height as ChainEpoch;
+        let state_params = app_state.state_params;
+
+        let mut exec_state = FvmExecState::new(
+            self.state_store.clone(),
+            self.multi_engine.as_ref(),
+            block_height,
+            state_params,
+        )
+        .context("error creating execution state")?;
```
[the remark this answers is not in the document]
cryptoAtwill: Updated

fendermint/app/src/ipc.rs
```diff
@@ -0,0 +1,149 @@
...
+        let decoded = decode_fevm_invoke(data)?;
+        DecodeFunctionReturn::<IPCParentFinality>::decode(decoded)
+    }
+
+    /// Get the last committed state, if exists.
+    fn get_committed_state(&self) -> anyhow::Result<Option<AppState>> {
+        let tx = self.db.read();
+        tx.get(&self.namespace, &AppStoreKey::State)
+            .context("get failed")
+    }
```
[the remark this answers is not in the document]
cryptoAtwill: Updated

fendermint/app/src/app.rs
```diff
@@ -64,11 +64,11 @@ pub enum AppError {
...
     /// Oldest state hash height.
     oldest_state_height: BlockHeight,
     /// Last committed version of the evolving state of the FVM.
-    state_params: FvmStateParams,
+    pub(crate) state_params: FvmStateParams,
```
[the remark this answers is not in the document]
cryptoAtwill: Updated

aakoshh:
I am not familiar with the methods in the Solidity actor, so I expected that this PR will handle cross messages, but I see from the description that it doesn't. Which issue is the one for cross message execution?

fendermint/vm/interpreter/src/fvm/exec.rs
```diff
@@ -90,7 +90,11 @@ where
         let method_num = msg.method_num;
         let gas_limit = msg.gas_limit;
 
-        let apply_ret = state.execute_explicit(msg)?;
+        let apply_ret = if from == system::SYSTEM_ACTOR_ADDR {
```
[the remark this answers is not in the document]
aakoshh: Ah it's okay, I do the same in the query interpreter because otherwise it threw an error, I just saw my comment there. Maybe a comment wouldn't hurt here either.

fendermint/vm/interpreter/src/chain.rs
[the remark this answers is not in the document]
aakoshh: But it's going to be here, right? Because this is the committing of the finality.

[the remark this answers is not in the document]
aakoshh: I found it useful to leave comments like `// TODO #123: Execute Cross Messages` where `123` is the Github issue number. It makes it easier to see what to expect, and also forces us to create issues with descriptions about what we're going to implement, like in https://github.com/consensus-shipyard/fendermint/issues/173

fendermint/app/src/ipc.rs
cryptoAtwill: imported

fendermint/app/src/app.rs
```diff
@@ -280,15 +283,45 @@ where
...
 
+    /// Get a read only fvm execution state. This is useful to perform query commands targeting
+    /// the latest state.
+    pub fn new_read_only_exec_state(
+        &self,
+    ) -> Result<Option<FvmExecState<ReadOnlyBlockstore<Arc<SS>>>>> {
+        let maybe_app_state = self.get_committed_state()?;
+
+        Ok(if let Some(app_state) = maybe_app_state {
+            let block_height = app_state.block_height;
```
aakoshh, line 308: I think this should also return `None` if the `block_height` is `None`. You won't be able to execute queries in that state. This is checked in the `query` method, because the state exists (put there by `init_committed_state` for the `info` method for example, but we won't be able to create an `exec_state` until after `genesis`.

cryptoAtwill: oh, you mean `block_height == 0`? I will add this check.

fendermint/app/src/ipc.rs
```diff
@@ -0,0 +1,149 @@
...
+        .context("error creating execution state")?;
+
+        let msg = get_parent_finality_to_fvm(block_height)?;
+        let data = exec_state
+            .execute_implicit(msg)?
+            .msg_receipt
+            .return_data
+            .to_vec();
+        let decoded = decode_fevm_invoke(data)?;
+        DecodeFunctionReturn::<IPCParentFinality>::decode(decoded)
```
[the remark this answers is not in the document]
cryptoAtwill: it will be default value set in the contract. I assume after genesis, the query function should return whatever the contract returns. If that's the case, it would return `{height: 0, hash: vec![0; 32]}`

fendermint/vm/interpreter/src/chain.rs
```diff
@@ -362,3 +398,30 @@ fn relayed_bottom_up_ckpt_to_fvm(
...
 }
+
+/// Convert a parent finality to fvm message
+fn parent_finality_to_fvm(
+    finality: IPCParentFinality,
+    validator_set: ValidatorSet,
+) -> anyhow::Result<FvmMessage> {
+    let params = RawBytes::new(EncodeWithSignature::<
+        gateway_router_facet::CommitParentFinalityCall,
+    >::encode(finality, validator_set)?);
```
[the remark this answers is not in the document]
cryptoAtwill: I updated to function call, should be easier to read now.

fendermint/vm/interpreter/src/chain.rs
```diff
@@ -221,32 +238,51 @@ where
...
                 IpcMessage::BottomUpExec(_) => {
                     todo!("#197: implement BottomUp checkpoint execution")
                 }
-                IpcMessage::TopDownProposal(_) => {
-                    todo!("implement TopDown handling; this is just a placeholder")
+                IpcMessage::TopDownProposal(p) => {
+                    let validator_set = downcast_err!(
+                        atomically_or_err(|| provider.validator_set(p.height as u64)).await
+                    )?
+                    .ok_or_else(|| anyhow!("cannot find validator set for block: {}", p.height))?;
```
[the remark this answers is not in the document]
cryptoAtwill: I think this one can be tricky as well. `ipc-agent` is an external service, even when we query it, it could be down or its data is somehow lagging. Then we still facing the same problem now. Do we continue execution or we stop the node to check what happened? I assumed (which I could be wrong) it's part of the service level agreement that parent finality provider should contain the data for execution, if it does not, something is worth checking.

fendermint/app/config/default.toml
```diff
@@ -63,7 +63,7 @@ port = 8545
 # TODO: fetch this value from genisis if possible.
 subnet_id = "/r0"
 
-[ipc.config]
+[ipc.topdown]
 # The number of blocks behind chain head to propose parent finality.
 chain_head_delay = 10
```
aakoshh, line 68:
```suggestion
chain_head_delay = 900
```
Let's be brutal here and assume the worst.

fendermint/app/src/cmd/run.rs
```diff
@@ -23,9 +26,15 @@ cmd! {
...
 }
 
-async fn create_parent_finality(settings: &Settings) -> anyhow::Result<InMemoryFinalityProvider> {
-    let provider = InMemoryFinalityProvider::new(settings.ipc.config.clone(), None);
-    Ok(provider)
+fn create_ipc_agent_proxy(
+    settings: &fendermint_vm_topdown::Config,
+    subnet_id: SubnetID,
+) -> anyhow::Result<IPCAgentProxy> {
+    let url = settings.ipc_agent_url.parse()?;
```
aakoshh, line 34:
```suggestion
    let url = settings.ipc_agent_url.parse().context("invalid agent URL")?;
```

fendermint/app/src/app.rs
```diff
@@ -194,6 +196,7 @@ where
         + Codec<FvmStateParams>,
     DB: KVWritable<S> + KVReadable<S> + 'static + Clone,
     SS: Blockstore + 'static + Clone,
+    P: ParentFinalityProvider,
```
aakoshh, line 199: By the way just wondering if we really need to make this generic if we are using STM, are there going to be any other implementations? Like the `ResolverPool` is passed directly, but in itself it's just a communication area, so any change of behaviour would be to whoever is feeding it on the other side.

fendermint/app/src/cmd/run.rs
```diff
@@ -47,8 +56,19 @@ async fn run(settings: Settings) -> anyhow::Result<()> {
...
+    let parent_finality_provider = if settings.ipc.is_topdown_enabled() {
+        info!("topdown finality enabled");
+        let config = settings.ipc.top_down_config()?.clone();
+        Arc::new(MaybeDisabledProvider::enabled(
+            InMemoryFinalityProvider::uninitialized(config),
+        ))
+    } else {
+        info!("topdown finality disabled");
+        Arc::new(MaybeDisabledProvider::disabled())
+    };
```
aakoshh, line 71: This seems very similar to the [Toggle](https://docs.rs/libp2p/latest/libp2p/swarm/behaviour/toggle/struct.Toggle.html) behaviour in libp2p, maybe you can use that name too.

fendermint/app/src/cmd/run.rs
```diff
@@ -59,9 +79,21 @@ async fn run(settings: Settings) -> anyhow::Result<()> {
...
-        parent_finality,
+        parent_finality_provider.clone(),
     )?;
 
+    if settings.ipc.is_topdown_enabled() {
+        let topdown_config = settings.ipc.top_down_config()?;
+        let app_parent_finality_query = AppParentFinalityQuery::new(app.clone());
+        launch_polling_syncer(
+            &app_parent_finality_query,
+            topdown_config.clone(),
```
aakoshh, line 92:
```suggestion
            topdown_config,
```

fendermint/app/src/cmd/run.rs
```diff
@@ -59,9 +79,21 @@ async fn run(settings: Settings) -> anyhow::Result<()> {
...
+            parent_finality_provider,
+            create_ipc_agent_proxy(topdown_config, settings.ipc.subnet_id.clone())?,
```
aakoshh, line 94:
```suggestion
            create_ipc_agent_proxy(topdown_config, settings.ipc.subnet_id.clone()).context("failed to create IPC Agent Proxy")?,
```

fendermint/app/src/cmd/run.rs
```diff
@@ -59,9 +79,21 @@ async fn run(settings: Settings) -> anyhow::Result<()> {
         state_store,
         interpreter,
         resolve_pool,
-        parent_finality,
+        parent_finality_provider.clone(),
     )?;
 
+    if settings.ipc.is_topdown_enabled() {
+        let topdown_config = settings.ipc.top_down_config()?;
```
aakoshh, line 88:
```suggestion
        let topdown_config = settings.ipc.top_down_config().context("failed to parse IPC config")?;
```

Although it would also be nice to find a way to combine this with the previous call to be honest, where the same `is_topdown_enabled()` was followed by this. You could turn it into an optional configuration, and then use that to drive the decision:

```
let topdown_config = if settings.ipc.is_topdown_enabled() { 
  Some(settings.ipc.top_down_config().context("failed to parse topdown config")?)
else { 
  None
};

if let Some(topdown_config) = topdown_config { <create-provider> }
...
if let Some(topdown_config) = topdown_config { <start-syncer> }
```

fendermint/app/src/ipc.rs
```diff
@@ -0,0 +1,95 @@
...
+        + Codec<AppState>
+        + Encode<AppStoreKey>
+        + Encode<BlockHeight>
+        + Codec<FvmStateParams>,
+    DB: KVWritable<S> + KVReadable<S> + 'static + Clone,
+    SS: Blockstore + 'static + Clone,
+    P: ParentFinalityProvider + Send + Sync,
+{
+    fn get_latest_committed_finality(&self) -> anyhow::Result<Option<IPCParentFinality>> {
+        let maybe_exec_state = self.app.new_read_only_exec_state()?;
```
aakoshh, line 61: You have to wonder if we could instantiate this `AppParentFinalityQuery` with a closure, something like 

```rust
pub struct AppParentFinalityQuery<F, DB>
where
    F: Fn() -> anyhow::Result<Option<FvmExecState<DB>>>
{
    get_exec_state: F
}

AppParentFinalityQuery::new(|| app.new_read_only_exec_state())
```

Way less code needs to be duplicated in the generic constraints.

fendermint/app/src/ipc.rs
```diff
@@ -0,0 +1,95 @@
...
+
+        let finality = if let Some(mut exec_state) = maybe_exec_state {
+            let msg = implicit_gateway_message(encode_get_latest_parent_finality()?);
+            tracing::debug!("query gateway parent finality message: {msg:?}");
+
+            let (apply_ret, _) = exec_state.execute_implicit(msg)?;
+
+            let data = apply_ret.msg_receipt.return_data.to_vec();
+            let decoded = decode_fevm_invoke_bytes(&Bytes::from(data))?;
```
aakoshh, line 70: Actually this won't work, sorry, I was confused last time because you copied all these `decode_fevm_invoke` related methods, and didn't notice. 

It wont' work because `decode_fevm_invoke` is specifically made to parse the data inside `DeliverTx` _coming from Tendermint_ which seems to add an extra layer of Base64 encoding on top. This is explained [here](https://github.com/consensus-shipyard/fendermint/blob/main/fendermint/rpc/src/response.rs#L10-L13) and it was a result of trial and erroring and painful debugging, I am not even sure how I figured it out, by accident.

In your case, I think you do have to copy [some magic](https://github.com/consensus-shipyard/fendermint/blob/main/fendermint/rpc/src/response.rs#L53-L55), which is this one:

```rust
    fvm_ipld_encoding::from_slice::<BytesDe>(&data)
        .map(|bz| bz.0)
        .map_err(|e| anyhow!("failed to deserialize bytes returned by FEVM: {e}"))
```

This one apparently is `RawBytes` encoded as IPLD. I will never be able to remember this and I'm not sure it would be healthy to do so. I originally got this from Vyzo's testing framework I think.

fendermint/app/src/ipc.rs
```diff
@@ -0,0 +1,95 @@
...
+
+#[inline]
+fn implicit_gateway_message(params: Vec<u8>) -> FvmMessage {
+    FvmMessage {
+        version: 0,
+        from: system::SYSTEM_ACTOR_ADDR,
+        to: ipc::GATEWAY_ACTOR_ADDR,
+        value: TokenAmount::zero(),
+        method_num: ipc::gateway::METHOD_INVOKE_CONTRACT,
+        params: RawBytes::new(params),
```
aakoshh, line 88: Based on my experience, this won't work :( 

The incantation is [here](https://github.com/consensus-shipyard/fendermint/blob/main/fendermint/rpc/src/message.rs#L132) :+1: 

```rust     
        let calldata = RawBytes::serialize(BytesSer(&params))?;
```

aakoshh: These could be added somewhere, I would suggest `fendermint_vm_actor_interface::evm`

aakoshh: The tell-tale of why `RawBytes::new` would not work is that it expects CBOR encoded(?) IPLD bytes. Otherwise if you have something else you need to call `RawBytes::serialize`, which works fine if we have something like `evm::ConstructorParams` like in the genesis process, but when it's just bytes it's confusing. And then you have this `BytesSer` on top because it has some extra hints for `serde`, and the `BytesDe` you saw previously is the opposite of that :man_shrugging:

fendermint/app/src/settings.rs
```diff
@@ -58,7 +59,21 @@ pub struct FvmSettings {
...
+    /// any top down checkpoint related operations
+    pub topdown: Option<fendermint_vm_topdown::Config>,
+}
+
+impl IPCSettings {
+    pub fn is_topdown_enabled(&self) -> bool {
+        !self.subnet_id.is_root() && self.topdown.is_some()
+    }
+
+    pub fn top_down_config(&self) -> anyhow::Result<&fendermint_vm_topdown::Config> {
```
aakoshh, line 74:
```suggestion
    pub fn topdown_config(&self) -> anyhow::Result<&fendermint_vm_topdown::Config> {
```

Either use `top_down` everywhere or `topdown` but let's not mix them.

fendermint/app/src/settings.rs
```diff
@@ -58,7 +59,21 @@ pub struct FvmSettings {
 pub struct IPCSettings {
     #[serde(deserialize_with = "deserialize_subnet_id")]
     pub subnet_id: SubnetID,
-    pub config: fendermint_vm_topdown::Config,
+    /// The config for top down checkpoint. It's None if subnet id is root or not activating
+    /// any top down checkpoint related operations
+    pub topdown: Option<fendermint_vm_topdown::Config>,
```
aakoshh, line 66: I reckon it's never going to be `None` because it's in the `default.toml` file.

aakoshh: FYI what I started doing in https://github.com/consensus-shipyard/fendermint/pull/210 is to put empty values into fields that have no reasonable defaults, and put some test values into `test.toml` so I can make sure the deserialization works. 

But I think in this case we can leave the defaults like delays etc. in there if you want. 

Another option is that there is a `mode` parameter on the CLI which can be used to overlay config files on top of the `default.toml` file, so you could have an `lotus-parent.toml` variant that sets the delay to 900, but I think this is such a blunt tool we should not use it like this and rely on env vars and custom config files instead that we put in there at the time of deployment.

fendermint/vm/interpreter/src/chain.rs
```diff
@@ -58,22 +61,27 @@ pub type ChainMessageCheckRes = Result<SignedMessageCheckRes, IllegalMessage>;
...
 
 #[async_trait]
-impl<I> ProposalInterpreter for ChainMessageInterpreter<I>
+impl<I, P> ProposalInterpreter for ChainMessageInterpreter<I, P>
 where
     I: Sync + Send,
+    P: ParentFinalityProvider + Send + Sync,
 {
-    type State = (CheckpointPool, Arc<InMemoryFinalityProvider>);
+    type State = (CheckpointPool, Arc<P>);
```
aakoshh, line 84: Hm, this is getting a bit infectious. I reckon a `Toggle<InMemoryFinalityProvider>` we could avoid the extra `P` and the phantomdata and save a lot of extra `where P: ParentFinalityProvider + Send + Sync` along the way, and then it would only affect the traits where its' part of the state.

fendermint/vm/interpreter/src/chain.rs
```diff
@@ -214,39 +223,67 @@ where
...
+                    let msg = parent_finality_to_fvm(finality.clone(), validator_set)?;
+                    let (state, ret) = self
+                        .inner
+                        .deliver(state, VerifiableMessage::NotVerify(msg))
+                        .await?;
+
+                    atomically_or_err::<_, fendermint_vm_topdown::Error, _>(|| {
+                        provider.set_new_finality(finality.clone())
+                    })
+                    .await?;
```
aakoshh, line 254: It would be really confidence inspiring to root out all the cases where an operation on the provider _cannot fail_ and remove the `Result`. Why can this fail so bad that it can crash Fendermint?

fendermint/vm/interpreter/src/chain.rs
```diff
@@ -214,39 +223,67 @@ where
...
+
+                    // TODO: execute top down messages,
+                    // TODO: see https://github.com/consensus-shipyard/fendermint/issues/241
```
aakoshh, line 257: Thank you for the link. Minor but it belongs further up _before_ `set_new_finality`, because by this point the cross messages would have disappeared form the cache. And you also need the previous finality for it, which should also be in the provider before it's overwritten. 

And here I see the value in the range query available for cross messages, although I still think that it would help if the block_hash and validator_set could not individually return `None`.

aakoshh: An example of why the current design is not ideal:
* `provider.validator_set(p.height as u64)` would return `None` if it's not found, which you handle by crashing Fendermint by returning ` anyhow!("cannot find validator set for block: {}", p.height)`. 
* `provider.cross_messages(from, to)` return `Vec<CrossMsg>` so it would be empty, if you implemented this execution where the TODO is now, you would never see an error, it would look like it's all dandy, just there were no messages. 

While in reality both should fail because of the same reason.

fendermint/vm/interpreter/src/chain.rs
```diff
@@ -355,3 +394,26 @@ fn relayed_bottom_up_ckpt_to_fvm(
 
     Ok(msg)
 }
+
+/// Convert a parent finality to fvm message
+fn parent_finality_to_fvm(
+    finality: IPCParentFinality,
+    validator_set: ValidatorSet,
+) -> anyhow::Result<FvmMessage> {
+    let params = RawBytes::new(encode_commit_parent_finality_call(finality, validator_set)?);
```
aakoshh, line 403:
```suggestion
    let params = RawBytes::serialize(encode_commit_parent_finality_call(finality, validator_set)?);
```

I _think_.

fendermint/vm/interpreter/src/fvm/exec.rs
```diff
@@ -95,7 +95,12 @@ where
         let method_num = msg.method_num;
         let gas_limit = msg.gas_limit;
 
-        let (apply_ret, emitters) = state.execute_explicit(msg)?;
+        // system actor executes implicitly
```
aakoshh, line 98:
```suggestion
        // Explicit execution requires `from` to be an account kind, which the System Actor is not, 
        // so it can only be executed implicitly.
```

fendermint/vm/topdown/src/convert.rs
```diff
@@ -0,0 +1,159 @@
...
+        IPCParentFinality {
+            height: value.height.as_u64(),
+            block_hash: value.block_hash.to_vec(),
+        }
+    }
+}
+
+/// Converts a Rust type FVM address into its underlying payload
+/// so it can be represented internally in a Solidity contract.
+fn addr_payload_to_bytes(payload: Payload) -> Bytes {
```
aakoshh, line 47: Since this is only used in `encode_commit_parent_finality_call`, it could return `anyhow::Result` and avoid a nasty crash if someone sends the wrong kind of address.

fendermint/vm/topdown/src/convert.rs
```diff
@@ -0,0 +1,159 @@
...
+        weights.push(U256::from_dec_str(&validator.weight)?);
+    }
+
+    let data = ethers::contract::encode_function_data(
+        commit_function,
+        gateway_router_facet::CommitParentFinalityCall {
+            finality: gateway_router_facet::ParentFinality::try_from(finality)?,
+            validators: addresses,
+            weights,
+        },
```
aakoshh, line 108: @adlrocha what's the outcome of the discussion about the configuration number? Is there an issue we can link here to add it to the contract?

fendermint/vm/topdown/src/convert.rs
```diff
@@ -0,0 +1,159 @@
...
+                GET_LATEST_PARENT_FINALITY_FUNC_NAME
+            )
+        })?
+        .get(0)
+        .ok_or_else(|| {
+            anyhow!(
+                "report bug, abi vec does not have {}",
+                GET_LATEST_PARENT_FINALITY_FUNC_NAME
+            )
+        })?;
```
aakoshh, line 153: This is the third time these 15 lines appear in this module. It seems easy to turn it into a method to be called with an `Abi` and function name.

fendermint/vm/topdown/src/disabled.rs
```diff
@@ -0,0 +1,108 @@
...
+
+impl ParentFinalityProvider for MaybeDisabledProvider {
+    fn last_committed_finality(&self) -> StmResult<IPCParentFinality, Error> {
+        match self {
+            MaybeDisabledProvider::Disabled => disabled!(),
+            MaybeDisabledProvider::Enabled(p) => p.last_committed_finality(),
+        }
+    }
+
+    fn next_proposal(&self) -> StmResult<Option<IPCParentFinality>, Error> {
```
aakoshh, line 88: So you had two options when you implemented a `Toggle`: 
* return an error, like you do
* return an "empty" result, which can be slightly for each method

The problem with returning an error are:
* I assume you had to extend the error types with yet another thing the caller has to reason about 
* where you[ call it from](https://github.com/consensus-shipyard/fendermint/pull/231/files#diff-38169d6d71afc04b5bc631226485edb63af1309907ec52553d956d0c55e86259R101-R104) in the `ChainMessageInterpreter` during `prepare_proposal` in the previous PR it's not expected to return an error, and returning one will immiately crash Fendermint

This is why I am begging you to not return errors where an empty result would do, don't return errors for things that can be expected. Only return errors where they are unexpected if you plan to use `?` inside interpreters, because they are fatal errors. 

The condition of not having a parent is not fatal. You could pass to the interpreter an `Option<InMemoryFinalityProvider>` which would tell the interpreter that there isn't any parent, but instead it looks like a normal provider it can call, except it bombs. 

And even on the `Enabled` case, why should it be an error that the syncer doesn't yet have data? We don't _have_ to propose a finality, we can just handle `None` and move on. Yet this one returns an error and it kills the app, or forces everyone to think whether its' enabled or disabled. 

I think it should return empties: it doesn't have any data about any height, and never proposes anything new. You can also change the `check` function to return `true` or `false` instead of an error, and then disabled can always return `false`. The only question is what to return from `last_committed_finality` but I think that can even be made into an internal method. The only thing that can return an error is `on_committed_finality` because there should be no commits as there is not parent.

fendermint/vm/topdown/src/finality.rs
```diff
@@ -163,14 +180,27 @@ impl ParentFinalityProvider for InMemoryFinalityProvider {
...
+    /// not be able to provide an existing finality from the storage. This provider requires an
+    /// existing committed finality. Providing the finality will enable other functionalities.
+    pub fn uninitialized(config: Config) -> Self {
+        Self::new(config, None)
+    }
+
+    // creates an initialized provider
+    pub fn initialized(config: Config, committed_finality: IPCParentFinality) -> Self {
+        Self::new(config, Some(committed_finality))
+    }
```
aakoshh, line 201: Okay so you have a conundrum because you wan to hand the finality provider to the application before it can be initialized, because it can at best get its data from genesis. 

I'd argue this is not the case. I'd say it's always possible as an operator to come up with a hash and height on the parent chain that we consider final, and add it to the `[ipc.topdown]` settings in the config file. And this doesn't have to be part of genesis, because if we start our node later, we can just use whatever is the latest checkpoint or final block from yesterday. 

There should be no reason to create an `uninitialized` provider.

Before the its' started, we can see if we already have in the ledger something newer than the config file, and use whichever is the higher height.

fendermint/vm/topdown/src/sync.rs
```diff
@@ -15,19 +15,55 @@ use ipc_sdk::subnet_id::SubnetID;
...
+                continue;
+            }
+            Err(e) => {
+                tracing::warn!("cannot get committed finality: {e}");
+                tokio::time::sleep(Duration::from_secs(5)).await;
+                continue;
+            }
+        };
+
+        atomically_or_err(|| view_provider.set_new_finality(finality.clone())).await?;
```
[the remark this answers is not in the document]
cryptoAtwill: I feel adding the height into config should do it, but still think we need the initialisation process. Because in Step 2, `Create an already initialized provider from it.`, one still has to wait to read from the ledger to confirm what's the starting height. This is an initialisation process.

fendermint/vm/topdown/src/disabled.rs
```diff
@@ -0,0 +1,108 @@
...
+        match self {
+            MaybeDisabledProvider::Disabled => disabled!(),
+            MaybeDisabledProvider::Enabled(p) => p.next_proposal(),
+        }
```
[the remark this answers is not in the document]
cryptoAtwill: very cool tip! Learned!

fendermint/app/src/ipc.rs
cryptoAtwill: I have updated the decoding and encoding all together and they are working now

fendermint/app/src/cmd/run.rs
cryptoAtwill: yep, updated to something along this line.

fendermint/app/src/ipc.rs
cryptoAtwill: nice tip, but this might be called repeatedly, so still have to hold the `App`.

fendermint/app/src/settings.rs
cryptoAtwill: I removed them from `default.toml`, I assume not everyone wants to start `ipc` in the first place. Adding to `test.toml` seems to be a better idea.

fendermint/vm/interpreter/src/chain.rs
cryptoAtwill: They all are simplified, they will not return `Option` but instead exponentially backoff retry or util it fails.

fendermint/vm/topdown/src/disabled.rs
cryptoAtwill: Errors are internalized and returns only option.

fendermint/vm/topdown/src/finality.rs
cryptoAtwill: I feel we should still keep the `uninitialized` and remove the `initialized`. First of all, I feel we should not use `config` because human errors are pretty common, like putting a value that is greater than the last committed finality. Also with the simplified error handling, if it's not initialized, it will just skip the current round of proposing.

cryptoAtwill:
@aakoshh I have included fixes from integration testing and also simplifications we discussed before. Please take a look again.

docs/running.md
```diff
@@ -154,6 +154,32 @@ The public key was spliced in as it was, in base64 format, which is how it would
...
+
+If you need ipc related function, let's add the subnet info to the Genesis with deployed subnet id: /r31415926
+
+```shell
+cargo run -p fendermint_app --release -- \
+      genesis --genesis-file test-network/genesis.json \
+      ipc \
+      gateway --subnet-id /r31415926 \
+      --bottom-up-check-period 10 --top-down-check-period 10 \
+      --msg-fee 1 --majority-percentage 60;
```
aakoshh, line 167:
```suggestion
      --msg-fee 1 --majority-percentage 65;
```
At least that number if higher than 2/3. With 60 I wonder if this is something other than BFT, does it tolerate 40% adversary, etc. But please correct me if I'm wrong and this has nothing to do with the classical restriction of Byzantine power.

fendermint/app/src/app.rs
```diff
@@ -275,15 +281,65 @@ where
...
     async fn modify_exec_state<T, F, R>(&self, f: F) -> Result<T>
     where
-        F: FnOnce((CheckpointPool, FvmExecState<SS>)) -> R,
-        R: Future<Output = Result<((CheckpointPool, FvmExecState<SS>), T)>>,
+        F: FnOnce(
+            (
+                CheckpointPool,
+                Arc<Toggle<CachedFinalityProvider>>,
+                FvmExecState<SS>,
+            ),
```
aakoshh, line 289: Maybe a type alias would help, as I see this repeated further down as well.

fendermint/app/src/cmd/mod.rs
```diff
@@ -105,3 +110,19 @@ fn settings(opts: &Options) -> anyhow::Result<Settings> {
...
 }
+
+pub(crate) fn set_network(network: u8) {
+    let network = Network::from_u8(network).unwrap();
+    set_current_network(network);
+}
+
+pub(crate) fn set_network_from_env() -> anyhow::Result<()> {
+    let network_raw = std::env::var("NETWORK")
+        // default to testnet
```
aakoshh, line 121: Please let's add an option [exactly like the log_level](https://github.com/consensus-shipyard/fendermint/blob/48a70ce5c10d8519bffe30b01e7fd9e39eaa433c/fendermint/app/src/options/mod.rs#L46-L48), like I asked :pray: 

The problem with this is manifold:
* You changed the default from main to testnet, when `running.md` is already full of `f0...` examples. Why should testnet be the default when the `Network::Mainnet` is the default. It sounds like a choice made because it's convenient for you with IPC testing with Eudico, rather than something that our users will ever want. It's inconsistent with the underlying type. 
* It is a hidden environment variable that nobody will ever know about unless they read the code _and_ find their way into `cmd/mod.rs`.
* All the env vars which are configuration file based start with `FM_`, with a few exceptions: `LOG_LEVEL` maybe a sort of "universal" env var that users might expect and get right in any application; `TENDERMINT_RPC_URL` already existed in the library I think, but it tells you exactly what it is. 
* Compared to those `NETWORK=1` is super cryptic, and the chance that anyone knows to set it is very low. 
* You need this env var in every terminal you open, and remember that this global affects all the address queries, so any CLI interaction where this value doesn't exist will end in frustration.

Whereas if you moved this to be a CLI parameter, then:
* It is discoverable with `--help`, rather than not something hidden in the codebase
* We can give it a suitable env var name such as `LOTUS_NETWORK` or whatever that helps recognise it for what it is.
* Give it a `default_value=0` to maintain backwards compatibility and match `Network`
* We can pass it as `--network=1` or use the env var in our infra scripts with Eudico. We should suffer, not our users.

fendermint/app/src/cmd/mod.rs
```diff
@@ -12,13 +12,18 @@ use async_trait::async_trait;
...
+use num_traits::FromPrimitive;
 
 pub mod eth;
 pub mod genesis;
 pub mod key;
 pub mod rpc;
 pub mod run;
 
+/// FVM address network, mainnet is 0
+const TEST_NETWORK: &str = "1";
```
aakoshh, line 25:
```suggestion
```
Move to CLI option and use 0, no need for special handling.

fendermint/app/src/cmd/run.rs
```diff
@@ -52,8 +88,27 @@ async fn run(settings: Settings) -> anyhow::Result<()> {
...
+    if let Some((agent_proxy, config)) = ipc_tuple {
+        let app_parent_finality_query = AppParentFinalityQuery::new(app.clone());
+        tokio::spawn(async move {
+            match launch_polling_syncer(
+                &settings.ipc.subnet_id,
+                app_parent_finality_query,
+                config,
+                parent_finality_provider,
+                agent_proxy,
+            )
```
aakoshh, line 103: The initialisation above looks cool, but as we discussed in our hudl I think the polling approach is a recipe for a race condition. I am sorry, I knew I recommended in the first place. 

So my proposal is that:
* We create the provider uninitialized, as you do. 
* To initialize it is the responsibility of the `App` - specifically the Genesis interpreter or the Snapshot Loader has to call `set_committed_finality` when they learn the final height in any way. 
* On subsequent runs, the two above won't run, in which case we can do our query once - the ledger should be available. This can be done here.
* We panic if any method is called that requires init, before it's actually initialized. This should not happen. 

By polling we risk that the App receives the block to process but, alas, the polling is still asleep and the query hasn't been performed.

fendermint/app/src/ipc.rs
aakoshh: You would still be able to hold the app if you used 

```rust
let qapp = app.clone()
AppParentFinalityQuery::new(move || qapp.new_read_only_exec_state())
```

The closure takes ownership of the app. But the signature is much simpler as you don't need to tell.

fendermint/app/src/main.rs
```diff
@@ -8,10 +8,15 @@ mod cmd;
...
 mod settings;
 
+use crate::cmd::set_network_from_env;
 use options::Options;
 
 #[tokio::main]
 async fn main() {
+    // Need to set env before Options is parsed because Options contains fvm Address related
+    // parameter, such as subnet id. Setting the network after Options is parsed will lead to error
+    set_network_from_env().expect("cannot set fvm address network from env");
```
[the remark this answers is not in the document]
aakoshh: I thought I'd try this out:
```
    /// Set the FVM Address Network.
    #[arg(short, long, default_value = "0", env = "NETWORK", value_parser = parse_network)]
    pub network: Network,

...
pub fn parse_network(s: &str) -> Result<Network, String> {
    let n: u8 = s
        .parse()
        .map_err(|_| format!("`{s}` isn't a network number"))?;
    let n = Network::from_u8(n).ok_or_else(|| format!("`{s}` isn't a recognised network"))?;
    fvm_shared::address::set_current_network(n);
    Ok(n)
}
```

What I accidentally discovered is that it works with `--network=1` but not `--network 1` or env var :disappointed: 
I'll try to figure out why `=` is special.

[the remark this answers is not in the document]
aakoshh: I will spend some cycles on this in https://github.com/consensus-shipyard/fendermint/issues/263

fendermint/app/src/settings.rs
aakoshh: Maybe add it to the `default.toml` but commented out, so people can still discover these and don't have to guess the syntax. You can also provide comments above them, which is useful.

In a similar situation with the IPLD Resolver I decided to include it, so people don't have to remember the dozens of fields that need to be filled in if they want to use it, with suitable default values, but deduct when it's disabled based on the absence of fields which don't have reasonable defaults. That way all they have to do is provide that field in an env var, and they are done.

fendermint/vm/interpreter/src/chain.rs
```diff
@@ -91,30 +103,52 @@ where
...
+                block_hash: proposal.block_hash,
+            })))
+        }
+
         Ok(msgs)
     }
 
     /// Perform finality checks on top-down transactions and availability checks on bottom-up transactions.
-    async fn process(&self, pool: Self::State, msgs: Vec<Self::Message>) -> anyhow::Result<bool> {
+    async fn process(&self, state: Self::State, msgs: Vec<Self::Message>) -> anyhow::Result<bool> {
```
aakoshh, line 119:
```suggestion
    async fn process(&self, (pool, provider): Self::State, msgs: Vec<Self::Message>) -> anyhow::Result<bool> {
```
More readable than `state.0.get_status`

fendermint/vm/interpreter/src/chain.rs
```diff
@@ -91,30 +103,52 @@ where
...
-            }
+                ChainMessage::Ipc(IpcMessage::TopDownExec(ParentFinality {
+                    height,
+                    block_hash,
+                })) => {
+                    let prop = IPCParentFinality {
+                        height: height as u64,
+                        block_hash,
+                    };
+                    return Ok(atomically(|| state.1.check_proposal(&prop)).await);
```
aakoshh, line 148:
```suggestion
                    let is_final = atomically(|| state.1.check_proposal(&prop)).await
                    if !is_final {
                      return Ok(false);
                    }
```

This seems to have regressed. Remember that we have lots of messages to check and this is just one.

fendermint/vm/interpreter/src/chain.rs
```diff
@@ -173,32 +211,55 @@ where
...
+                    Ok(((pool, provider, state), ChainMessageApplyRet::Signed(ret)))
                 }
                 IpcMessage::BottomUpExec(_) => {
                     todo!("#197: implement BottomUp checkpoint execution")
                 }
-                IpcMessage::TopDown => {
-                    todo!("implement TopDown handling; this is just a placeholder")
+                IpcMessage::TopDownExec(p) => {
+                    if !provider.is_enabled() {
+                        return Err(anyhow!("toggle on ipc top down to process messages"));
```
aakoshh, line 221:
```suggestion
                        return Err(anyhow!("cannot execute IPC top-down message: parent provider disabled"));
```
Sorry, I found that message a confusing, so I assume someone less familiar with the code would do even more so.

fendermint/vm/topdown/src/convert.rs
```diff
@@ -0,0 +1,122 @@
...
+
+impl TryFrom<IPCParentFinality> for gateway_router_facet::ParentFinality {
+    type Error = anyhow::Error;
+
+    fn try_from(value: IPCParentFinality) -> Result<Self, Self::Error> {
+        if value.block_hash.len() != 32 {
+            return Err(anyhow!("invalid block hash length, expecting 32"));
+        }
+
+        let mut array = [0u8; 32];
```
aakoshh, line 26: Minor but if you call this `block_hash` then it carries the meaning, it still does not shadow `value.block_hash`, and you don't have to type `block_hash: array` when assigning in `Self`, just its name.

fendermint/vm/topdown/src/finality.rs
```diff
@@ -3,85 +3,174 @@
...
+
+                    let to_sleep = Duration::from_secs(wait);
+                    tokio::time::sleep(to_sleep).await;
+
+                    wait *= 2;
+                    continue;
+                }
+            }
+
+            break res;
```
aakoshh, line 60: Nice, I don't think I ever saw `break` return a value from the loop :+1:

fendermint/vm/topdown/src/finality.rs
```diff
@@ -106,114 +195,64 @@ impl ParentViewProvider for InMemoryFinalityProvider {
...
-            let (block_hash, _, _) = v;
-            block_hash.clone()
+        let latest_height = if let Some(h) = self.cached_data.latest_height()? {
+            h
         } else {
-            return abort(Error::HeightNotFoundInCache(height));
+            // latest height is not found, meaning we dont have any prefetched cache, we just be
+            // optimistic and vote yes as the previous latest committed finality check is passed
+            // and it's valid.
+            return Ok(true);
```
aakoshh, line 219: I could be wrong but this doesn't compute to me. 

Looking at `check_block_hash`, a proposal _will_ be rejected if we don't know have data about that height, that is, we don't consider it final. For the same reason, we should reject this as well: we simply don't know! The "previous latest commmitted finality check" only says that the new proposal is higher than what we considered final last time, but that could be the tip of the chain, or something that doesn't even exist.

I don't think we should vote "yes" just because we know we will vote "no" in the other method, and we know the actual check method calls both. They should be self-contained IMO.

fendermint/vm/topdown/src/sync.rs
```diff
@@ -3,139 +3,245 @@
...
+        let mut finality = match query.get_latest_committed_finality() {
+            Ok(Some(finality)) => finality,
+            Ok(None) => {
+                tracing::debug!("app not ready for query yet");
+                tokio::time::sleep(Duration::from_secs(5)).await;
+                continue;
+            }
+            Err(e) => {
+                tracing::warn!("cannot get committed finality: {e}");
+                tokio::time::sleep(Duration::from_secs(5)).await;
```
aakoshh, line 61: Okay I see that there are no panics and no errors returned from the provider, so no race condition :+1:

fendermint/vm/topdown/src/sync.rs
```diff
@@ -3,139 +3,245 @@
...
         tokio::spawn(async move {
             loop {
-                // Syncing with parent with the below steps:
-                // 1. Get the latest height in cache or latest height committed increment by 1 as the
-                //    starting height
-                // 2. Get the latest chain head height deduct away N blocks as the ending height
-                // 3. Fetches the data between starting and ending height
-                // 4. Update the data into cache
-                if let Err(e) = sync_with_parent(&config, &agent, &provider).await {
+                let mut interval =
```
aakoshh, line 138: Any reason this is created inside the loop rather than out?

fendermint/vm/topdown/src/sync.rs
```diff
@@ -195,6 +306,14 @@ impl IPCAgentProxy {
         Ok(height as BlockHeight)
     }
 
+    pub async fn get_genesis_epoch(&self, _subnet_id: &SubnetID) -> anyhow::Result<ChainEpoch> {
```
aakoshh, line 309: Can you add some docstrings to explain the rationale of this method? 
* Why does it take `SubnetID` parameter if it doesn't use it?
* What is `self.child_subnet` of an agent?
* What is a `genesis_epoch` of a `ValidatorSet`?
* What does the `get_validator_set` return when it's called with `None`?

fendermint/vm/interpreter/src/chain.rs
```diff
@@ -173,32 +211,55 @@ where
...
+                    }
+
```
aakoshh, line 223: Something I haven't considered before but we must handle is that currently a validator can propose multiple finalities in the same block, and do so in arbitrary order. We must decide whether to allow this, and if so how to proceed during execution. 

Say the last committed finality is 100, and the next validator proposes multiple ones: [120, 110] in that order. Say we look at 120, it's final, then 110, obviously that is final too, so we vote "yes". Then during execution we do 120 first, okay, and then what about 110? We should skip it. We can check in the cache, because the previous run should have called `set_new_finality`.

The other option is to not vote if there are multiple records in the same block, which prevents ordering problems.

aakoshh: Yet another option, which may be actually the easiest and cleanest to implement, is to reject backwards going finality in the contract, and then to [inspect the return value](https://github.com/consensus-shipyard/fendermint/pull/236/files#diff-38169d6d71afc04b5bc631226485edb63af1309907ec52553d956d0c55e86259R233) and if it's in an error then do not execute the cross message, and also do not update the cache. 

This avoids any potential double execution of cross-messages.

fendermint/app/src/cmd/run.rs
cryptoAtwill: I think adding to `Genesis Interpreter` might introduce extra code burden as we have to introduce `ipc-agent` related code to get the subnet genesis epoch. Also the ChainInterpreter init chain does not have the AppState initialized. Maybe we add to `App::init_chain`, but somehow, we have to deal with the extra imported crate the `App` needs to hold. 

I feel the ideal case would be starting the `Polling` in the background then we run the `App`. But because we need to query the `AppState` that we have to init the `App` first. We can add a barrier  in `App` to signal start the service only when `Polling` has started. That way we dont pollute `App` with too many dependencies, but also starting the `App` only when `Polling` is running.

Btw, I'm wondering if the race is indeed an issue. Because the polling is not ready, that means we dont have anything, maybe dont propose and we dont vote for any proposals. Does this impact `tendermint` a lot?

aakoshh: Thanks, I can't find it now but I commented somewhere that I agree that because you are never returning errors due to "data not ready" any more, polling is fine, we just skip a round of proposals, no biggie, no race :+1:

fendermint/vm/topdown/src/sync.rs
```diff
@@ -3,139 +3,245 @@
...
+use std::cmp::min;
 use std::sync::Arc;
 use std::time::Duration;
 
+/// The max number of blocks polling should query each parent view update. If the number of blocks
+/// polled equals this value, it would stop polling for this iteration and commit the result to cache.
+const MAX_PARENT_VIEW_BLOCK_GAP: BlockHeight = 100;
+/// When polling parent view, if the number of top down messages exceeds this limit,
+/// the polling will stop for this iteration and commit the result to cache.
+const TOPDOWN_MSG_LEN_THRESHOLD: usize = 500;
```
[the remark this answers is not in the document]
cryptoAtwill: It is possible to place it in the config but I think they should be aligned by the protocol. Say if the `MAX_PARENT_VIEW_BLOCK_GAP` is reached, the fetched data would be committed into cache and parent finality would start producing proposals. In a way, these parameters control what the height would be in the proposals, if we use config, then anyone can input some value, I'm not sure how this would impact the consensus. If there are no impact, then in config should work too.

fendermint/vm/interpreter/src/chain.rs
cryptoAtwill: The backward rejection is implement in the Membership PR with configuration number, see https://github.com/consensus-shipyard/ipc-solidity-actors/blob/3332383c11dfee0dc60d29a913c733d8e25752e8/src/lib/LibGateway.sol#L79.

fendermint/vm/topdown/src/sync.rs
cryptoAtwill: copy paste error...

fendermint/vm/topdown/src/finality.rs
cryptoAtwill: I changed to `false`

fendermint/vm/topdown/src/sync.rs
cryptoAtwill: I assumed everyone is aware... Doc string added.

fendermint/vm/interpreter/src/chain.rs
aakoshh: Okay that ignores the membership but we also need to actually reject the block hash and height, rather than go backwards.

cryptoAtwill: this one; https://github.com/consensus-shipyard/ipc-solidity-actors/blob/876b5fa56ce5dc5222589f761b8afddef41ca4f4/src/lib/LibGateway.sol#L60

fendermint/vm/topdown/src/convert.rs
```diff
@@ -0,0 +1,122 @@
...
+    validator_set: ValidatorSet,
+) -> anyhow::Result<Vec<u8>> {
+    let commit_function = get_evm_function(COMMIT_PARENT_FINALITY_FUNC_NAME)?;
+
+    let validators = validator_set.validators.unwrap_or_default();
+
+    let mut addresses = vec![];
+    let mut weights = vec![];
+    for validator in validators {
+        let raw_address = validator.worker_addr.unwrap_or(validator.addr);
```
aakoshh, line 82: What's a `worker_addr`?

fendermint/vm/topdown/src/convert.rs
```diff
@@ -0,0 +1,122 @@
...
+        Payload::Secp256k1(v) => ethers::types::Bytes::from(v),
+        Payload::Delegated(d) => {
+            let addr = d.subaddress();
+            let b = ethers::abi::encode(&[Token::Tuple(vec![
+                Token::Uint(U256::from(d.namespace())),
+                Token::Uint(U256::from(addr.len())),
+                Token::Bytes(addr.to_vec()),
+            ])]);
+            ethers::types::Bytes::from(b)
+        }
```
aakoshh, line 58: I don't know how we can handle delegated addresses. At the end of the day we need to give Tendermint a public key and this is the only place I will learn it from.

fendermint/vm/topdown/src/convert.rs
cryptoAtwill: Explained in slack, hope it clarifies.

fendermint/vm/topdown/src/convert.rs
cryptoAtwill: From the discussion on Slack, we can try creating our own address namespace.

fendermint/vm/topdown/src/sync.rs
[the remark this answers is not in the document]
aakoshh: My thinking on issues like this is that:
* if something affects consensus, ie. it's is vital that we all use the same value otherwise we might be rejecting proposals, then it should be in the genesis block
* if it's something validators can choose then it should be in the config, because otherwise it's an arbitrary number in the code we cannot easily change when we realise there is a better value

In this case it depends on whether we want these be a basis of rejection. 

Like I commented somewhere I think we won't be able to apply limits directly on these, more like on gas, and even then it's impossible to say up front what's a good value because it depends on what else is in the block, so I recommended that we actually append values into a queue rather than execute the whole checkpoint as one, and then chip away at it deterministically in the upcoming blocks.

So for now I think these are more about protecting the process from out-of-memory problems, and should be in the config.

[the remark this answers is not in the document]
cryptoAtwill: This works for me!

fendermint/vm/topdown/src/convert.rs
aakoshh: Yes so for the record in Eudico we had an Ethereum address for joining a subnet and a `worker_addr` for mining with, which may have been a different kind of address. 

We are discussing to use `worker_addr` to carry the public key of the validator, so that we can hand it to Tendermint inside validator set changes.

Cargo.toml
```diff
@@ -122,6 +122,8 @@ tendermint-proto = { version = "0.31" }
 
 # Using the IPC SDK without the `fil-actor` feature so as not to depend on the actor `Runtime`.
 ipc-sdk = { git = "https://github.com/consensus-shipyard/ipc-actors.git", default-features = false, branch = "main" }
+ipc-agent = { git = "https://github.com/consensus-shipyard/ipc-agent.git", branch = "main" }
+ipc-agent-sdk = { git = "https://github.com/consensus-shipyard/ipc-agent.git", branch = "dev" }
```
aakoshh, line 126: We should probably use the same branch for anything coming from the same repo, no?

fendermint/app/src/cmd/mod.rs
aakoshh: If you merge from `main` you can get rid of this now.

aakoshh (approved):
I think this is fine now :see_no_evil: :crossed_fingers:

fendermint/vm/topdown/src/sync.rs
aakoshh: Can't see the docs :thinking:

Cargo.toml
cryptoAtwill: I accidentally committed the experiment, it's updated. But now with the new `ipc-agent` dev, everything broke again.

PR #237, opened 2023-09-04
Update async stm to 0.4

cryptoAtwill:
As the title suggests...


aakoshh (approved):
Cheers! 

I was confused because I thought I already did this, but it's only in https://github.com/consensus-shipyard/fendermint/pull/210/files :+1:

adlrocha (approved):

PR #264, opened 2023-09-12
FM-263: Set network

### LINKED ISSUE #263: Try to find a way to unify the FVM Network setting with the rest
See:
* https://github.com/consensus-shipyard/fendermint/pull/236/files#r1321503591
* https://github.com/consensus-shipyard/fendermint/pull/236/files#r1321521809

So the issue is that the FVM `Network` type relies on a global setting which can be either `Mainnet` or `Testnet`, and depending on that value parsing `SubnetID` or in general the `Address` type is rejected if the prefix (`f` for main, `t` for test) doesn't match the global. 

My initial idea was to add a `--network` CLI parameter with a suitable default and env var, however it looks like `clap` provides no guarantees that this will be parsed before other fields, and the subnet IDs can still fail to parse. I tried doing a partial parse first with `#[clap(ignore_errors=true, disable_help_flag=true)]` but it wont' handle the defaults in the presence of other options that it should not care about :disappointed: 

There must be a way to do it though. 

Another issue which could be solved here is that it takes way too long to compile `app` after any tiny change. Perhaps moving the `config` bits into a separate crate would help.

aakoshh:
Closes #263 

The PR adds a `--network` CLI parameter next to `--log-level` which accepts `main`, `mainnet`, `test`, `testnet`, `f`, `t`, `0` or `1` as values and calls `fvm_shared::address::set_current_network` with the value. The default value is `mainnet`, to match the default of `Network` itself, and it's possible to set it via the `FVM_NETWORK` env var as well.

The way I was able to achieve capturing the `--network` first, while ignoring other options is to add a `GlobalOptions` that gets parsed first and has only the `network` as an option, treating any trailing arguments and options as just strings. Then we use this value to prime the FVM, before going on to parse `Options`. To facilitate this a `parse` method has been added to the `options` module itself. 

The `--network` is also present on `Options` for discoverability, but there it's basically ignored. I decided against setting the global value in the parser itself, to emphasize that this needs to be done explicitly and up front - it turned out that commands are parsed from the inside out, so this was the last to be parsed (although using `--network=1` instead of `--network 1` seemed to have influenced that, I did not investigate why).

I also moved the entire `options` module into its own `fendermint_app_options` crate to reduce build times, so it doesn't take a minute to try any change.


fendermint/app/options/src/lib.rs
```diff
@@ -0,0 +1,149 @@
...
+    #[arg(short, long, default_value = "dev")]
+    pub mode: String,
+
+    /// Set the logging level.
+    #[arg(short, long, default_value = "info", value_enum, env = "LOG_LEVEL")]
+    pub log_level: LogLevel,
+
+    /// Global options repeated here for discoverability.
+    #[command(flatten)]
+    pub global: GlobalArgs,
```
cryptoAtwill, line 88: what does it mean by `discoverability`? Do we have to config this parameter?

aakoshh: I'll change it to `FM_NETWORK`, maybe that's more what people would expect.

PR #267, opened 2023-09-13
FM-244: GatewayCaller

### LINKED ISSUE #244: Method in the Gateway to return the checkpoint interval
The Fendermint interpreter needs to know the checkpointing boundary to know when to construct a checkpoint.

An alternative to this is if the Gateway has an "end block" method and does the height checks on its own, but then it needs to return much more than if we break these out into methods, and Fendermint would tell from the return value whether a transition has taken place or not.

aakoshh:
Closes #244 

The PR adds a `fendermint_vm_interpreter::fvm::state::fevm::ContractCaller` component to help calling EVM methods, and on top of that a `fendermint_vm_interpreter::fvm::state::ipc::GatewayCaller` to act as a facade for the Gateway facets. A test demonstrates getting the bottom-up checkpoint period.


aakoshh:
@adlrocha @dnkolegov my unit test failed because I didn't know about [these minimum values](https://github.com/consensus-shipyard/ipc-solidity-actors/blob/main/src/GatewayDiamond.sol#L19-L23) applied in Solidity. 

Do you think its worth keeping those minimums there? They look quite arbitrary: top-down is not respected any more, and the bottom up is like, why 10?

adlrocha:
@aakoshh, agree, we should remove them. These are inherited from `eudico` and they were chosen as the minimum number of epochs that you needed to wait on our PoW implementation to avoid re-orgs. 

@dnkolegov, do you mind adding a PR removing these? (you'll probably have a faster turnaround than I do as I am flying back home tonight)

fendermint/vm/interpreter/src/fvm/state/fevm.rs
```diff
@@ -0,0 +1,96 @@
...
+            params: calldata,
+            gas_limit: fvm_shared::BLOCK_GAS_LIMIT,
+            gas_fee_cap: TokenAmount::from_atto(0),
+            gas_premium: TokenAmount::from_atto(0),
+        };
+
+        let (ret, _) = state.execute_implicit(msg)?;
+
+        if !ret.msg_receipt.exit_code.is_success() {
+            bail!(
```
cryptoAtwill, line 77: Is there a way to return the error instead of `bail`? For `bail`, it will crash the app? Because if it fails, the caller might handle the error from contract?

aakoshh: The [bail!](https://docs.rs/anyhow/latest/anyhow/macro.bail.html) macro is essentially just a `return Err(anyhow!(...));`, so it doesnt' crash immediately. 

The choice here was whether to turn an unsuccessful exit code into an `Err` or not. If we treated it as success, what should it return, because I want the `T` from the method call. An `Option<T>` would mask the error, a nested `anyhow::Result<Result<T, (ExitCode, String)>>` could be more trouble than worth (we can crash if we fail to call the FVM but inspect the result for errors - we can always add a `try_call` method for that. I thought this may be the simplest case: when we call the thing should succeed, or else it's big trouble.

We just need to remember that if we call methods that can revert or otherwise are expected to fail, we need to add a `try_call` and use that with them.

aakoshh: Added `try_call`

PR #272, opened 2023-09-15
FM-252: Adopt next configuration at the end of the checkpoint period

### LINKED ISSUE #252: ExecInterpreter::end adopt next validator set and return updates to Tendermint
As a subnet operator, I want the  Fendermint interpreter to call #244 to determine the checkpoint interval; if the current block height is divisible by the checkpoint interval, the interpreter should call #245 and #246 to figure out the change in the validator set (if any), and return them in the appropriate format to CometBFT to initiate the power transfer.

### LINKED ISSUE #256: Relayer subscribe to events from the checkpoint contract

aakoshh:
Closes #252


aakoshh:
Closed it because there were so many changes that it's easier to reimplement then merge.

PR #275, opened 2023-09-21
migrate ipc actor binding

cryptoAtwill:
Migrate the ipc actor rust binding to use the one from `ipc-solidity-actor` repo.


Cargo.toml
```diff
@@ -122,6 +122,7 @@ tendermint-proto = { version = "0.31" }
 
 ipc-sdk = { git = "https://github.com/consensus-shipyard/ipc-agent.git", branch = "dev" }
 ipc_ipld_resolver = { git = "https://github.com/consensus-shipyard/ipc-ipld-resolver.git", branch = "main" }
+ipc_actors_abis = { git = "https://github.com/consensus-shipyard/ipc-solidity-actors.git", branch = "ci" }
```
cryptoAtwill, line 125: branch is `ci` as it's not merged into dev yet, once merged with update to dev

adlrocha (approved):

Makefile
```diff
@@ -89,15 +89,6 @@ $(BUILTIN_ACTORS_BUNDLE): $(BUILTIN_ACTORS_CODE)
 	rustup target add wasm32-unknown-unknown && \
 	cargo run --release -- -o output/$(shell basename $@)
 
-
-# Compile the ABI artifacts and Rust bindings for the IPC Solidity actors.
-ipc-actors-abi: $(IPC_ACTORS_ABI)
```
aakoshh, line 94: I think you could leave this in and only remove the `cargo build ...` stuff, similarly to how the `actor-bundle` target works. The reason is that it's still needed for the docker-build and it's just a more convenient command to have, in case it's needed.

aakoshh (approved):
Thanks, looks good :+1: 

As per my only comment, I'd leave the target for compiling the ABI as long as we actually need it for the Docker image.

aakoshh:
https://github.com/consensus-shipyard/ipc-solidity-actors/pull/214 has updated Solidity but not the ABI, something isn't right there.

Makefile
```diff
@@ -89,15 +89,6 @@ $(BUILTIN_ACTORS_BUNDLE): $(BUILTIN_ACTORS_CODE)
...
-	cargo build --release -p fendermint_vm_ipc_actors
-
-# Force reompilation of the IPC actors.
-clean-ipc-actors-abi:
```
aakoshh, line 98: I'd also leave this target while we need the ABI files for docker, as otherwise I'd have to go into the other repo and do a `git pull` to refresh it, because otherwise it doesn't do it.

It would be nice if we could find a way to link to the source that cargo downloaded due to the project reference and get the stuff from there, so there's no chance for a mismatch.

PR #276, opened 2023-09-21
FM-255: Sign checkpoint

### LINKED ISSUE #255: ExecInterpreter::end sign the BottomUpCheckpoint and broadcast transaction
As a subnet _validator_ (not just a full node), I want the Fendermint interpreter to check if I was a validator in the checkpointing period which just ended (see #245). If I was a validator, I want the interpreter to sign the checkpoint created in #254 and broadcast a transaction with a message to #249 to CometBFT

aakoshh:
Closes #255 
Depends on: 
* https://github.com/consensus-shipyard/fendermint/pull/278 
* https://github.com/consensus-shipyard/fendermint/pull/273
* https://github.com/consensus-shipyard/fendermint/pull/280

The PR adds a step after creating a checkpoint and adding it to the ledger to create and broadcast a transaction that adds the validator signature.

### DIFF
--- a/fendermint/vm/interpreter/src/fvm/broadcast.rs
+++ b/fendermint/vm/interpreter/src/fvm/broadcast.rs
@@ -3,13 +3,14 @@
 
 use anyhow::{anyhow, bail, Context};
 use ethers::types as et;
-use fendermint_rpc::client::FendermintClient;
+use fendermint_rpc::message::GasParams;
 use fendermint_rpc::query::QueryClient;
-use fendermint_vm_actor_interface::evm;
-use fendermint_vm_message::{chain::ChainMessage, query::FvmQueryHeight, signed::SignedMessage};
-use fvm_ipld_encoding::{BytesSer, RawBytes};
+use fendermint_rpc::tx::{CallClient, TxClient, TxCommit};
+use fendermint_rpc::{client::FendermintClient, message::MessageFactory};
+use fendermint_vm_message::query::FvmQueryHeight;
 use fvm_shared::{address::Address, chainid::ChainID, econ::TokenAmount, BLOCK_GAS_LIMIT};
-use libsecp256k1::SecretKey;
+use libsecp256k1::{PublicKey, SecretKey};
+use num_traits::Zero;
 use tendermint_rpc::Client;
 
 /// Broadcast transactions to Tendermint.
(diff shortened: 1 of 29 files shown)


fendermint/vm/interpreter/src/fvm/broadcast.rs
```diff
@@ -0,0 +1,143 @@
...
+/// This is typically something only active validators would want to do
+/// from within Fendermint as part of the block lifecycle, for example
+/// to submit their signatures to the ledger.
+///
+/// The broadcaster encapsulates the tactics for figuring out the nonce,
+/// the gas limit, potential retries, etc.
+#[derive(Clone)]
+pub struct Broadcaster<C> {
+    client: FendermintClient<C>,
+    secret_key: SecretKey,
```
dnkolegov, line 27: Do we zeroize these secret bytes?

dnkolegov (approved):
I do not have comments other than a question related to zeroization.

Should we create an issue to not forget about that in the future?

fendermint/vm/interpreter/src/fvm/broadcast.rs
aakoshh: Can you elaborate? I have an idea of what you mean, that if a program crashes it should not be recoverable from a memory dump, however I don't see when I could do this given that it's needed for each submission. What is the recommended way of using secret keys? Should it be read from disk every time? Should it be protected by a password? How should that password be protected?

dnkolegov: You are right. The best description of the problem can be found here - https://github.com/veorq/cryptocoding#clean-memory-of-secret-data

In Rust, zeroize - https://crates.io/crates/zeroize - is a de facto standard for clearing memory.

And this is an example of usage of `zeroize` - https://github.com/matrix-org/vodozemac/pull/43/files

aakoshh: Thanks, I created https://github.com/consensus-shipyard/fendermint/issues/279 to deal with this.

fendermint/vm/interpreter/src/fvm/mod.rs
```diff
@@ -23,18 +24,42 @@ pub use fendermint_vm_message::query::FvmQuery;
...
 
+pub use self::broadcast::Broadcaster;
 use self::state::ipc::GatewayCaller;
 
 pub type FvmMessage = fvm_shared::message::Message;
 
+#[derive(Clone)]
+pub struct ValidatorContext<C> {
+    /// The secret key the validator uses to produce blocks.
+    secret_key: SecretKey,
```
adlrocha, line 35: Out of a curiosity, how are we passing the `secret_key` to the context?

adlrocha (approved):
A few clarifications, but LGTM. Thanks!

fendermint/vm/interpreter/src/fvm/broadcast.rs
```diff
@@ -0,0 +1,144 @@
...
+    C: Client + Clone + Send + Sync,
+{
+    pub fn new(
+        client: C,
+        secret_key: SecretKey,
+        gas_fee_cap: TokenAmount,
+        gas_premium: TokenAmount,
+    ) -> Self {
+        let client = FendermintClient::new(client);
+        // TODO: We could use f410 addresses to send the transaction, but the `MessageFactory` assumes f1.
```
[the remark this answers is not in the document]
aakoshh: Hm, okay, I will think of a way to add support for both. 

I thought that since we discussed that the user who bonds with their Ethereum account to become a validator has to provide a public key they will use to sign blocks, it's not a big ask for them to fund such an f1 account with tokens. But if that's awkward, we can add something to show whether the secret key should be treated as an f410 or f1 address.

fendermint/vm/interpreter/src/fvm/broadcast.rs
```diff
@@ -0,0 +1,144 @@
...
+
+        let factory = MessageFactory::new(self.secret_key.clone(), sequence, chain_id)
+            .context("failed to create MessageFactory")?;
+
+        // Using the bound client as a one-shot transaction sender.
+        let mut client = self.client.clone().bind(factory);
+
+        // TODO: Maybe we should implement something like the Ethereum facade for estimating fees?
+        // I don't want to call the Ethereum API directly (it would be one more dependency).
+        // Another option is for Fendermint to recognise transactions coming from validators
```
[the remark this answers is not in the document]
aakoshh: Yeah, I agree it would be nice. The only question is whether validators would try to submit other transactions knowing they will get in front of others. But if we also look at the `to` of the message being the gateway, I reckon that should weed out the chaff.

fendermint/vm/interpreter/src/fvm/mod.rs
aakoshh: If the settings are pointing at validator key file, we read it and create a context from it here: https://github.com/consensus-shipyard/fendermint/pull/276/files#diff-6d8c711ef368a10a52b47f7702c1eff949ea4bac181370ffae4e8f49f3e1f9a2R46

fendermint/vm/interpreter/src/fvm/broadcast.rs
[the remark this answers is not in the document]
aakoshh: Created https://github.com/consensus-shipyard/fendermint/issues/283

PR #284, opened 2023-09-26
Check network name variable

dnkolegov:
This PR fixes a minor bug in infra's `run.sh`.

### DIFF
--- a/infra/run.sh
+++ b/infra/run.sh
@@ -5,6 +5,13 @@ if [ $# -ne 1 ]; then
   exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


dnkolegov:
> Is the network name same as the subnet ID?

It should be. NETWORK_NAME should be equal to CHAIN_NAME

adlrocha:
> It should be. NETWORK_NAME should be equal to CHAIN_NAME
Can we force `NETWORK_NAME == CHAIN_NAME` so that we don't have inconsistencies?

adlrocha (approved):

PR #285, opened 2023-09-27
FM-283: Add validator account kind setting

### LINKED ISSUE #283: Support f410 address for broadcasting transactions
https://github.com/consensus-shipyard/fendermint/pull/276#discussion_r1337349267

Somehow make it configurable whether the address used to sign transactions is an Ethereum or a regular address.

aakoshh:
Closes #283 

Changes the `validator_key` setting from a `PathBuf` into a struct with a `path` and a `kind` field, where the kind can be `regular` or `ethereum`, which is how accounts are added to the genesis file too. This is then used to derive the `addr` which is passed to the `Broadcaster` and the `MessageFactory`, which allows validators to use f410 accounts to pay for transactions. 

I thought about introducing two of these fields, one for the validator (needs to sign the checkpoint, but there the account kind isn't needed) and an optional one for the broadcaster (can be used to override the validator key for sending transactions) but decided to hold on, since we discussed in a comment with @adlrocha that recognising validator transactions could be something we use to subsidize their transactions.

### DIFF
--- a/fendermint/app/src/cmd/run.rs
+++ b/fendermint/app/src/cmd/run.rs
@@ -1,10 +1,13 @@
 // Copyright 2022-2023 Protocol Labs
 // SPDX-License-Identifier: Apache-2.0, MIT
 
-use anyhow::{anyhow, Context};
+use anyhow::{anyhow, bail, Context};
 use fendermint_abci::ApplicationService;
 use fendermint_app::{App, AppConfig, AppStore, BitswapBlockstore};
+use fendermint_app_settings::AccountKind;
+use fendermint_crypto::SecretKey;
 use fendermint_rocksdb::{blockstore::NamespaceBlockstore, namespaces, RocksDb, RocksDbConfig};
+use fendermint_vm_actor_interface::eam;
 use fendermint_vm_interpreter::{
     bytes::{BytesMessageInterpreter, ProposalPrepareMode},
     chain::{ChainMessageInterpreter, CheckpointPool},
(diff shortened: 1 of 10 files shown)


fendermint/app/src/cmd/run.rs
```diff
@@ -237,3 +250,11 @@ fn to_resolver_config(settings: &Settings) -> anyhow::Result<ipc_ipld_resolver::
 
     Ok(config)
 }
+
+fn to_address(sk: &SecretKey, kind: &AccountKind) -> anyhow::Result<Address> {
```
adlrocha, line 254: Why do you need to re-implement [here](https://github.com/consensus-shipyard/fendermint/pull/285/files#diff-7d11b1e9a3cc49dd08f1497ba1c3e20ccd2a8b82f5618a46df2e58eb8cb4f870R384)?

adlrocha (approved):
Nice one, thanks!

fendermint/app/src/cmd/run.rs
aakoshh: Because these are two distinct `AccountKind` types at the moment: 
* one in the `fendermint_app_settings` which deals with deserializing the config file
* the other in `fendermint_app_options` which is about CLI parameters

These two crates have no reference to each other at the moment and sharing an `AccountKind` didn't seem important enough to introduce it. Not sure where else to put it at the moment. The one in `options` needs a `clap` specific attribute too, which is not something I would like to see spreading, although it might work with `serde` I don't know.

PR #286, opened 2023-09-27
FM-281: Retrieve cross messages and put their hash into the checkpoint

### LINKED ISSUE #281: Retrieve and hash the cross messages
Related to https://github.com/consensus-shipyard/fendermint/issues/254 when creating a checkpoint we need to produce a hash over the cross-messages and put it into the `BottomUpCheckpoint`, so that it can be validated in https://github.com/consensus-shipyard/fendermint/issues/192

aakoshh:
Closes #281 

The PR populates `BottomUpCheckpointNew::cross_messages_hash` with the hash of the ABI encoded `Vec<CrossMsg>` retrieved by calling the new `bottomUpMessages` on `GatewayGetterFacet` with the height of the current checkpoint, which was added in https://github.com/consensus-shipyard/ipc-solidity-actors/pull/214

This is based on my assumption how the contract works:
* [getNextEpoch](https://github.com/consensus-shipyard/ipc-solidity-actors/blob/0b587608d77ec659f7a31f31f15c05560c20196b/src/lib/LibVoting.sol#L84) calculates the next checkpoint height based on the current one as `((height / period) + 1) * period` which gives some multiple of `period`, which should match the way we [check](https://github.com/consensus-shipyard/fendermint/blob/main/fendermint/vm/interpreter/src/fvm/checkpoint.rs#L128) for the end-of-period height: `height % period == 0`, so we'll use the same height values.
* [commitBottomUpMsg](https://github.com/consensus-shipyard/ipc-solidity-actors/blob/0b587608d77ec659f7a31f31f15c05560c20196b/src/lib/LibGateway.sol#L180) pushes the message into the next checkpointed height


dnkolegov:
> Closes #281
> 
> The PR populates `BottomUpCheckpointNew::cross_messages_hash` with the hash of the ABI encoded `Vec<CrossMsg>` retrieved by calling `bottomUpCheckpoints` on `GatewayGetterFacet` with the height of the current checkpoint.
> 
> This is based on my assumption how the contract works:
> 
> * [getCurrentBottomUpCheckpoint](https://github.com/consensus-shipyard/ipc-solidity-actors/blob/0b587608d77ec659f7a31f31f15c05560c20196b/src/lib/LibGateway.sol#L35) inserts a checkpoint at the given block height if it doesn't exist
> * [commitBottomUpMsg](https://github.com/consensus-shipyard/ipc-solidity-actors/blob/0b587608d77ec659f7a31f31f15c05560c20196b/src/lib/LibGateway.sol#L180) pushes the message into the checkpoint
> * [getNextEpoch](https://github.com/consensus-shipyard/ipc-solidity-actors/blob/0b587608d77ec659f7a31f31f15c05560c20196b/src/lib/LibVoting.sol#L84) calculates the next checkpoint height based on the current one as `((height / period) + 1) * period` which gives some multiple of `period`, which should match the way we [check](https://github.com/consensus-shipyard/fendermint/blob/main/fendermint/vm/interpreter/src/fvm/checkpoint.rs#L128) for the end-of-period height: `height % period == 0`, so we'll use the same height values.

Why is [createBottomUpCheckpoint](https://github.com/consensus-shipyard/ipc-solidity-actors/blob/0b587608d77ec659f7a31f31f15c05560c20196b/src/gateway/GatewayRouterFacet.sol#L26) not used to create a checkpoint?

dnkolegov:
>
- commitBottomUpMsg

-  is there some method in the contract to call to get the cross messages ?

-  plus the getCurrentBottomUpCheckpoint is internal. How does one go about retrieving them to Rust?

@aakoshh It would be very helpful if you could describe the flow of creating, populating, and signing a checkpoint from "Fendermint as a client" perspective.

I am afraid that the current implementation (from Limechain) is incompatible with what you want to do, and we must reimplement Gateway API.

Another question is is there any relation with populating BottomUpCheckpoint and collecting signatures from validators?

aakoshh:
> Why is `createBottomUpCheckpoint` not used to create a checkpoint?

@dnkolegov it _is_ used, [here](https://github.com/consensus-shipyard/fendermint/blob/33f83aa1309e2e06529fabde1ec7ccb4d37476ef/fendermint/vm/interpreter/src/fvm/checkpoint.rs#L87), with the Solidity method called [by the GatewayCaller](https://github.com/consensus-shipyard/fendermint/blob/33f83aa1309e2e06529fabde1ec7ccb4d37476ef/fendermint/vm/interpreter/src/fvm/state/ipc.rs#L121). 

What I was trying to describe there were my understanding of why, when a contract calls `commitBottomUpMsg`, a _legacy_ checkpoint will be automatically created with empty fields, and the cross-message appended to it. It looks like it will be retrievable with the height that I will actually create the _new_ checkpoint with, at which point I can retrieve the accumulated cross messages from it.

aakoshh:
> It would be very helpful if you could describe the flow of creating, populating, and signing a checkpoint from "Fendermint as a client" perspective.

Okay, I will open another PR with a sequence diagram.

aakoshh:
> I am afraid the current implementation (from Limechain) is incompatible with what you want to do, and we must reimplement Gateway API.

Can you explain why it's incompatible? I may be misunderstanding what the code is doing, but like I said `commitBottomUpMsg` does not inspect the `exists` return value, so it looks like this is normal behaviour and it will stash the messages as intended. I have no argument against reimplementing it, but  it looks like it should technically work this way as well.

aakoshh:
> Another question is, is there any relation between populating BottomUpCheckpoint with cross messages and collecting signatures for this checkpoint from validators?

As you can see in https://github.com/consensus-shipyard/fendermint/pull/276, the signatures are over the hash of the checkpoint, which incorporates the hash of the cross messages, so the signatures are implicitly over the cross messages as well.

fendermint/vm/interpreter/src/fvm/state/ipc.rs
```diff
@@ -75,11 +77,33 @@ impl<DB: Blockstore> GatewayCaller<DB> {
         self.getter.call(state, |c| c.bottom_up_check_period())
     }
 
+    /// Fetch the bottom-up messages enqueued in a given checkpoint.
+    pub fn bottom_up_msgs(
+        &self,
+        state: &mut FvmExecState<DB>,
+        height: u64,
+    ) -> anyhow::Result<Vec<getter::CrossMsg>> {
+        // NOTE: This particular method is marked as deprecated but it's currently the only known way to retrieve
```
adlrocha, line 86: Why is this method deprecated (//cc @dnkolegov)? I would say that this is the right way to get the bottom-up cross-net messages. 
* When there is a new bottomup message, [`commitBottomUpMsg`](https://github.com/consensus-shipyard/ipc-solidity-actors/blob/bbf1d9b2bf7250a0c5432894e15ed5f4b0453692/src/lib/LibGateway.sol#L178) pushes the message to the height of the next checkpoint to be created.
* When we need to populate that checkpoint we can directly get the bottomup messages as you are doing.

adlrocha (approved):

fendermint/vm/interpreter/src/fvm/state/ipc.rs
dnkolegov: It was deprecated because it returned `BottomUpMessageLegacy` several days ago.
Now, it looks like it's usual - https://github.com/consensus-shipyard/ipc-solidity-actors/blob/bbf1d9b2bf7250a0c5432894e15ed5f4b0453692/src/gateway/GatewayGetterFacet.sol#L58

aakoshh: Ah, yes, I should have removed this comment, thanks!

PR #288, opened 2023-09-27
FM-282: Retry broadcast

### LINKED ISSUE #282: Retry failed signature broadcast
https://github.com/consensus-shipyard/fendermint/pull/276#discussion_r1337356926

Follow up https://github.com/consensus-shipyard/fendermint/issues/255 by retrying the transaction up to a configurable number of times if it failed for reasons that can be expected to succeed next time (e.g. if the sequence didn't match).

aakoshh:
Closes #282 

Retries broadcasting the transaction up to 5 times if it fails with nonce problems.

### DIFF
--- a/fendermint/app/settings/src/lib.rs
+++ b/fendermint/app/settings/src/lib.rs
@@ -189,12 +189,6 @@ impl Settings {
         &self.home_dir
     }
 
-    /// Indicate whether we have configured the IPLD Resolver to run.
-    pub fn resolver_enabled(&self) -> bool {
-        !self.resolver.connection.listen_addr.is_empty()
-            && self.resolver.subnet_id != *ipc_sdk::subnet_id::UNDEF
-    }
-
     /// Tendermint RPC URL from the environment or the config file.
     pub fn tendermint_rpc_url(&self) -> anyhow::Result<Url> {
         // Prefer the "standard" env var used in the CLI.
@@ -260,13 +254,13 @@ mod tests {
     #[test]
     fn parse_default_config() {
         let settings = parse_config("");
-        assert!(!settings.resolver_enabled());
+        assert!(!settings.resolver.enabled());
     }
 
     #[test]
     fn parse_test_config() {
         let settings = parse_config("test");
-        assert!(settings.resolver_enabled());
+        assert!(settings.resolver.enabled());
     }
 
     #[test]
--- a/fendermint/app/settings/src/lib.rs
+++ b/fendermint/app/settings/src/lib.rs
@@ -6,16 +6,17 @@ use config::{Config, ConfigError, Environment, File};
 use fvm_shared::econ::TokenAmount;
 use ipc_sdk::subnet_id::SubnetID;
 use serde::Deserialize;
-use serde_with::serde_as;
 use std::path::{Path, PathBuf};
 use tendermint_rpc::Url;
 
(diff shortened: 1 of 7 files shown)


fendermint/app/settings/src/lib.rs
```diff
@@ -83,28 +84,13 @@ pub struct DbSettings {
     pub state_hist_size: u64,
 }
 
-#[serde_as]
+/// Settings affecting how we deal with failures in trying to send transactions to the local CometBFT node.
+/// It is not expected to be unavailable, however we might get into race conditions about the nonce which
```
adlrocha, line 88: Do you mind elaborating on this a bit. What you mean is that if you retry after some other transaction is broadcast to CometBFT we may get a race as the retry has a lower nonce than the second one which was (for some reason) accepted first?  
```suggestion
/// It is not expected to be unavailable, however we might get into race conditions about the nonce which
```

fendermint/vm/interpreter/src/fvm/broadcast.rs
```diff
@@ -49,87 +72,112 @@ where
...
-        // TODO: Maybe we should implement something like the Ethereum facade for estimating fees?
-        // I don't want to call the Ethereum API directly (it would be one more dependency).
-        // Another option is for Fendermint to recognise transactions coming from validators
-        // and always put them into the block to facilitate checkpointing.
-        let mut gas_params = GasParams {
-            gas_limit: BLOCK_GAS_LIMIT,
-            gas_fee_cap: self.gas_fee_cap.clone(),
-            gas_premium: self.gas_premium.clone(),
-        };
+            // TODO: Maybe we should implement something like the Ethereum facade for estimating fees?
```
adlrocha, line 102: IIRC, we discussed this in some previous PR, I think always including validator messages is a good idea. I don't think they should be competing against other user transactions for block space.

fendermint/vm/interpreter/src/fvm/broadcast.rs
```diff
@@ -139,3 +187,20 @@ where
         }
     }
 }
+
+/// Decide if it's worth retrying the transaction.
+fn can_retry(code: tendermint::abci::Code) -> bool {
```
adlrocha, line 192: Nice one! This can prevent a lot of useless work!

adlrocha (approved):
LGTM. Thanks!

fendermint/app/settings/src/lib.rs
aakoshh: Ahm :dizzy_face: Maybe...? :sweat_smile: 

What I meant was that:
* CometBFT is expected to be there (it's calling us after all), so I'm not trying any exponential backoff here to wait until it's available, which is why the settings are unsophisticated.
* The race conditions arise from the fact that the operator might be using the keys to sign other transactions, such as releasing funds to the parent network, which is practically the only failure scenario I could think of where it's worth retrying a broadcast immediately with a fresh nonce. The retries can get into trouble if we were to use simultaneous broadcasts from different places - but arguably if we know we are doing that then we should add an internal queue to the broadcaster.

fendermint/vm/interpreter/src/fvm/broadcast.rs
aakoshh: Yes I agree, that seems to cut to the heart of the problem much better. The fee estimation would be accidental complexity.

PR #289, opened 2023-09-28
DOCS: Checkpointing

aakoshh:
Created a sequence diagram to showcase checkpointing, as promised in https://github.com/consensus-shipyard/fendermint/pull/286#issuecomment-1738733735


adlrocha (approved):
This is amazing, @aakoshh. It was long overdue. Thanks!
 
- The role of the IPC agent has been removed. It is now a library that Fendermint and the relayer can leverage to directly call the relevant child/parent subnets. It may be better to just remove it from the diagram and depict the relayer and Fendermint directly calling the relevant nodes?

- Also, it may be also worth having a diagram of how top-down messages are executed (@cryptoAtwill or me can pick this one in a follow-up PR).

aakoshh:
> The role of the IPC agent has been removed. It is now a library that Fendermint and the relayer can leverage to directly call the relevant child/parent subnets.

Okay, I only went by [what happened](https://github.com/consensus-shipyard/fendermint/pull/236/files#diff-6d8c711ef368a10a52b47f7702c1eff949ea4bac181370ffae4e8f49f3e1f9a2R37-R44) in the now deprecated PRs where IIRC the agent had its own URL, but I could be wrong on that one. Let's fix the diagram when @cryptoAtwill again introduces the syncer. Until then we can treat this diagram as speculative.

PR #290, opened 2023-09-29
FM-255: Skip broadcasting checkpoint signature if syncing

aakoshh:
Related to #255 

As the title says, when a node re-syncs from genesis or catching up and encounters checkpoints which were already finished without its assistance, there's no need to send the signature again.

### DIFF
--- a/fendermint/vm/interpreter/src/fvm/exec.rs
+++ b/fendermint/vm/interpreter/src/fvm/exec.rs
@@ -142,29 +142,32 @@ where
                     .find(|v| v.public_key.0 == ctx.public_key)
                     .cloned()
                 {
-                    let secret_key = ctx.secret_key.clone();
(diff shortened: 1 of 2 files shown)


adlrocha (approved):
I feel this `syncing` method will come really handy in the future. Thanks!

aakoshh:
@adlrocha yes, there are many use cases where the `syncing` would be useful, for example we don't need to sync with the parent from the last committed final height up to the current tip while we are catching up with our own chain, we can just wait to learn what we need to execute. 

It's a bit unfortunate that it's placed under the `FvmMessageInterpreter` because that's not accessible in general, but it's easy enough for anyone to query if they have a Tendermint RPC client. Maybe we can find a better place for it, perhaps under `FendermintClient`.

PR #291, opened 2023-09-29
CHORE: Update ABI bindings after BottomUpCheckpointNew rename

aakoshh:
Updates the ABI after https://github.com/consensus-shipyard/ipc-solidity-actors/pull/214/


dnkolegov (commented):
LGTM

PR #292, opened 2023-09-29
FM-259: Sign all incomplete checkpoints

### LINKED ISSUE #259: Upon restart list incomplete checkpoints and re-broadcast signatures if necessary
As a Fendermint operator, if my node crashes after #244 but before #255 was included in the next blocks, I want the system to retrieve incomplete checkpoints from the actor using #258, re-sign and re-broadcast them, to make sure we don't stall by never sending a checkpoint to the parent.

aakoshh:
Closes #259 

The PR changes the end-of-period checkpoint signature broadcasting by validators to apply on all incomplete (a.k.a. pending) checkpoints where they were validators.

So, instead of doing this on node start, it's done on each checkpoint period end. This has the following benefits:
* `main.rs` stays simpler and the checkpointing logic is kept in the interpreter
* a validator doesn't have to restart their node to fill in their signatures if e.g. they ran out of funds, they just need to get some tokens and it will automatically be retried

### DIFF
--- a/fendermint/vm/interpreter/src/fvm/checkpoint.rs
+++ b/fendermint/vm/interpreter/src/fvm/checkpoint.rs
@@ -12,14 +12,14 @@ use fvm_shared::{address::Address, chainid::ChainID};
 use fendermint_crypto::SecretKey;
 use fendermint_vm_actor_interface::ipc::BottomUpCheckpoint;
 use fendermint_vm_genesis::{Power, Validator, ValidatorKey};
-use fendermint_vm_ipc_actors::gateway_router_facet::SubnetID;
+use fendermint_vm_ipc_actors::gateway_getter_facet as getter;
+use fendermint_vm_ipc_actors::gateway_router_facet as router;
 
 use super::{
     broadcast::Broadcaster,
     state::{ipc::GatewayCaller, FvmExecState},
+    ValidatorContext,
 };
-
-pub type Checkpoint = BottomUpCheckpoint;
 
 /// Validator voting power snapshot.
 #[derive(Debug, Clone)]
@@ -38,7 +38,7 @@ pub async fn maybe_create_checkpoint<C, DB>(
     client: &C,
     gateway: &GatewayCaller<DB>,
     state: &mut FvmExecState<DB>,
-) -> anyhow::Result<Option<(Checkpoint, PowerTable, PowerUpdates)>>
+) -> anyhow::Result<Option<(router::BottomUpCheckpoint, PowerTable, PowerUpdates)>>
 where
     C: Client + Sync + Send + 'static,
     DB: Blockstore + Sync + Send + 'static,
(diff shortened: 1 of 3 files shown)


dnkolegov (approved):

fendermint/vm/interpreter/src/fvm/checkpoint.rs
```diff
@@ -87,11 +87,66 @@ where
...
+            .await
+            .context("failed to get power table")?;
+
+        if let Some(validator) = power_table
+            .0
+            .iter()
+            .find(|v| v.public_key.0 == validator_ctx.public_key)
+            .cloned()
+        {
+            // TODO: Code generation in the ipc-solidity-actors repo should cater for this.
```
adlrocha, line 114: @cryptoAtwill, can you double-check that this is the case once you migrate to the `ipc-solidity-actors` bindings?

fendermint/vm/interpreter/src/fvm/exec.rs
```diff
@@ -129,45 +129,49 @@ where
...
+                        "the current checkpoint is incomplete"
+                    );
+
+                    let client = self.client.clone();
+                    let gateway = self.gateway.clone();
+                    let chain_id = state.chain_id();
+                    let height = checkpoint.block_height;
+                    let validator_ctx = ctx.clone();
+
+                    tokio::spawn(async move {
```
adlrocha, line 161: Why is it possibel to do the broadcast of incomplete signatures in parallel here without resulting in nonce races? (I also ask because in the comment above you mention that broadcasts can't be done in parallel).

adlrocha (approved):

fendermint/vm/interpreter/src/fvm/checkpoint.rs
aakoshh: I don't remember seeing that this has been implemented already, we just talked about the fact that the bindings generate as many versions of `SubnetID` and whatnot as there are facets. So it's just a note for the future that if this gets out of hand, we should try to solve it there.

fendermint/vm/interpreter/src/fvm/exec.rs
aakoshh: Sorry, it looks like my comment wasn't as clear as it could have been. 

What I meant was that we should not try to kick off a separate background task for each pending checkpoint like this:
```rust
for cp in pending_checkpoints {
  tokio::spawn(async move {
    checkpoint::broadcast_signature(broadcaster, cp).await;
  });
}
```
That's because every time the `broadcaster` is asked to send a transaction it will query the state for the nonce, and in this case that will surely mean all of the checkpoint submissions will get the same nonce and only one will go through. 

Instead, what we have is this:
```rust
for cp in pending_checkpoints {

    checkpoint::broadcast_signature(broadcaster, cp).await;
  });
}
```

PR #297, opened 2023-10-03
CHORE: Use the ABI bindings from the Solidity repo

cryptoAtwill:
Migrate ipc_actor rust binding to `ipc-solidity-actor` repo.

PR #298, opened 2023-10-03
CHORE: Generate the ABI under the IPC actor solidity checkout

aakoshh:
This is a followup for https://github.com/consensus-shipyard/fendermint/pull/297

Before that PR, `make ipc-actors-abi` checked out the `ipc-solidity-actors` repository and compiled the ABI artifacts, which were then used to:
1. generate Rust bindings and 
2. copy the contracts into the Docker file. 

Which tag to check out was defined in the `Makefile` as `IPC_ACTORS_TAG` which could be overridden as an env var.

After that PR, we no longer had to do 1), but still needed the checkout for 2). However now the `IPC_ACTORS_TAG` could have differed from the tag defined in `Cargo.toml` where the project dependency was defined.

This PR removes the need for checking out the `ipc-solidity-actors` repo by finding where `cargo` did the checkout and generating the ABI bindings there.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ docker/.artifacts
 cometbft
 test-network
 .idea
+.make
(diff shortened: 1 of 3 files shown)


dnkolegov (approved):

adlrocha (approved):
High-level looks good. However, there seems to be a slight issue with the compilation directory as the CI is not happy. Feel free to merge when this is resolved. Thanks!

cryptoAtwill (approved):

PR #307, opened 2023-10-06
FM-301: IPC genesis validators

### LINKED ISSUE #301: Bootstrap child subnet
Related: https://github.com/consensus-shipyard/ipc-solidity-actors/pull/223

Once the minimum requirements set by a subnet to bootstrap (mainly minimum number of validators, and minimum collateral in the reference implementation) are fulfilled, the subnet actor in the parent emits an event to notify that the infrastructure for the subnet can be spawned. This event can be used to create deterministically the Fendermint genesis, setting the initial validators of the subnet to the ones determined by the parent, and thus deploying the infrastructure for the child subnet according to the information in the parent.

The genesis of the child subnet should: 
- Initialize the gateway of the with the initial validators that have bootstrapped the network (from there on, updates to membership are handled through top-down proof-of-finality). 
- Initialize  the initial validators in Fendermint so they are picked up right away to start creating new blocks.

aakoshh:
Closes #301 

The PR uses the `genesis.validators` to derive a new `validators` field in the IPC Gateway constructor. 

It assumes the following change(s) in the system:
* That each validator has an Ethereum (a.k.a. f410) account corresponding to its public key, that is, we no longer support using a public key that is different from what the account providing the collateral is using.
* That we recognise that each validator address _has_ to be an f410 address as long as IPC is implemented in Solidity - you simply can't send a transaction to Ethereum from an f1 address.
* That the public key can be passed to the gateway constructor so that the ledger remembers it and can return it when we next have to do update validators.

With those assumptions we use the `PublicKey` in the validators of the genesis file to produce a `H160` address which we pass to the constructor along with the `PublicKey` and the collateral. 

# Collateral to Power conversion

`Power` is limited to `u64` in CometBFT, if use used it for `atto`, which represents 10**-18 FIL, we'd be limited to ~18 FIL maximum collateral, which is not enough. The validator updates in IPC come as collateral, so we have to convert them to `u64` somehow. 

To do so, the PR adds a new `genesis.power_scale` field which the number of decimals used to convert FIL to power. 0 means 1 FIL gives 1 power, 3 means it gives 1000, -1 would mean 10 FIL gives 1 power. 

The `power_scale` is saved to the `AppState`, so that we can use it every time there is a new checkpoint. Ultimately it will have to be defined in the parent SubnetActor, where it could also be used to carry out quorum checks in exactly the same way as they happen on the child subnet.


docs/running.md
```diff
@@ -29,9 +29,17 @@ mkdir test-network
...
 First, create a new `genesis.json` file devoid of accounts and validators. The `--base-fee` here is completely arbitrary.
+The `--power-scale` value of `0` means we'll grant 1 voting power per 1 FIL; to use more precision, we can set it to `3`
+to use milliFIL for example.
 
 ```shell
-cargo run -p fendermint_app -- genesis --genesis-file test-network/genesis.json new --chain-name test --base-fee 1000 --timestamp 1680101412
+cargo run -p fendermint_app -- \
+  genesis --genesis-file test-network/genesis.json \
+  new \
+  --chain-name test \
```
adlrocha, line 39: I don't know if this is the case already, but I would force this one to be a `SubnetID` and make `chain_name == subnet_id` rejecting any `chain_name` that is not a valid `SubnetID`.

fendermint/vm/actor_interface/src/ipc.rs
```diff
@@ -145,10 +149,14 @@ impl ValidatorMerkleTree {
...
     use ethers::contract::{EthAbiCodec, EthAbiType};
-    use ethers::core::types::{H160, U256};
+    use ethers::core::types::{Bytes, H160, U256};
     use fendermint_vm_genesis::ipc::GatewayParams;
+    use fendermint_vm_genesis::{Collateral, Validator};
+    use fvm_shared::address::Error as AddressError;
     use fvm_shared::address::Payload;
     use fvm_shared::econ::TokenAmount;
+    // TODO: This is what the contract _should_ take in the constructor.
+    use ipc_actors_abis::subnet_actor_getter_facet::GenesisValidator;
```
adlrocha, line 159: I don't think I fully agree with this one. 
* When we bootstrap a child subnet with Fendermint, we should use the `IpcProvider` to get the genesis validators from the bootstrap  event (or the storage), This is a `GenesisValidator` type. 
* Fendermint uses this `GenesisValidator` to extract the public key and construct the gateway. I can pass a `GenesisValidator` as a construct parameter of the gateway if its easier for you, but we wouldn't be using the metadata and public keys for anything there.

fendermint/vm/actor_interface/src/ipc.rs
```diff
@@ -167,35 +175,55 @@ pub mod gateway {
...
+                    _ => return Err(AddressError::InvalidPayload),
                 };
                 route.push(H160::from(addr.0))
             }
+
+            // Every validator has an Ethereum address.
+            let validators = validators
+                .into_iter()
+                .map(|v| {
+                    let pk = v.public_key.0.serialize();
```
adlrocha, line 205: I will update the [PR](https://github.com/consensus-shipyard/ipc-solidity-actors/pull/228) so you can use the right type.

docs/running.md
aakoshh: IPC is optional with Fendermint; I think at the moment there is nothing to check that you are using a subnet ID for the name. IIRC the name is passed to CometBFT as-is so that it can use it in whatever way it wants, and within Fendermint it's effectively destroyed as soon as it's hashed into a number, we will never want to parse it. So, it's just a convention we agreed that we should use the subnet ID for the name because it's unique.

fendermint/vm/actor_interface/src/ipc.rs
aakoshh: > When we bootstrap a child subnet with Fendermint, we should use the IpcProvider to get the genesis validators from the bootstrap event
That should happen in https://github.com/consensus-shipyard/fendermint/issues/306 to produce the JSON file. 

> I can pass a GenesisValidator as a construct parameter of the gateway if its easier for you, but we wouldn't be using the metadata and public keys for anything there.

You wouldn't be using it in the constructor, however we will need the public key again and again every time we create a new checkpoint and need to update the CometBFT power table in https://github.com/consensus-shipyard/fendermint/issues/252 

New joiners will have their public keys appear there, so it is only natural that _all_ validators should have it in the ledger. Otherwise I'd constantly have to go back to the parent to figure out whether some validators who don't have metadata might have been one of the genesis validators. 

The metadata and the public key are one and the same thing at the moment.

Cargo.toml
```diff
@@ -126,10 +126,10 @@ tendermint-proto = { version = "0.31" }
 
 ipc-sdk = { git = "https://github.com/consensus-shipyard/ipc-agent.git", branch = "dev" }
 ipc_ipld_resolver = { git = "https://github.com/consensus-shipyard/ipc-ipld-resolver.git", branch = "main" }
-ipc_actors_abis = { git = "https://github.com/consensus-shipyard/ipc-solidity-actors.git", branch = "dev" }
+ipc_actors_abis = { git = "https://github.com/consensus-shipyard/ipc-solidity-actors.git", branch = "ipc-221-follow-up" }
```
adlrocha, line 129: New branch with power_scale and apply finality changes https://github.com/consensus-shipyard/ipc-solidity-actors/pull/235

docs/running.md
```diff
@@ -29,9 +29,17 @@ mkdir test-network
 ### Create a new Genesis file
 
 First, create a new `genesis.json` file devoid of accounts and validators. The `--base-fee` here is completely arbitrary.
+The `--power-scale` value of `0` means we'll grant 1 voting power per 1 FIL; to use more precision, we can set it to `3`
```
adlrocha, line 32: To clarify, in the contract I ended going with `power_scale` as the accuracy for the power table so as you mention `0 == FIL, 3 == miliFil, 12 == atto`.

fendermint/vm/genesis/src/lib.rs
```diff
@@ -17,6 +18,9 @@ use fendermint_vm_encoding::IsHumanReadable;
 #[cfg(feature = "arb")]
 mod arb;
 
+/// Power conversion decimal points, e.g. 3 decimals means 1 power per milliFIL.
+pub type PowerScale = i32;
```
adlrocha, line 22: I am using `uint8` for `power_scale` in the contract (I guess it shouldn't be an issue).

adlrocha (approved):

adlrocha (commented):
Subjet to the update of the constructor parameters for the new version of the contracts, LGTM. Thanks :pray:

fendermint/vm/genesis/src/lib.rs
aakoshh: I allow negative values because -1 makes perfect sense depending on the value of your token: in our case you'd get a power for every 10 FIL. It would be great not to have an exception trying to pass it in.

docs/running.md
aakoshh: atto would be 18, but otherwise sounds like you're using it the same way.

adlrocha: Already fixed in the contract, I keep saying that atto is 12 because I am stupid :)

fendermint/vm/genesis/src/lib.rs
adlrocha: Gotcha, fixed in the contract.

PR #308, opened 2023-10-09
Topdown

cryptoAtwill:
Migrating previous `topdown` crate to new validator change set implementation.

@aakoshh It is the same as it was last approved, only we query the validator change set now instead of full validators.


Cargo.toml
```diff
@@ -124,9 +124,10 @@ tendermint-config = "0.33.0"
 tendermint-rpc = { version = "0.31", features = ["secp256k1", "http-client", "websocket-client"] }
 tendermint-proto = { version = "0.31" }
 
-ipc-sdk = { git = "https://github.com/consensus-shipyard/ipc-agent.git", branch = "dev" }
+ipc-sdk = { git = "https://github.com/consensus-shipyard/ipc-agent.git", branch = "validator-metadata-change" }
```
cryptoAtwill, line 127: will update to `dev` once merged the depending PRs.

fendermint/vm/topdown/src/finality.rs
```diff
@@ -0,0 +1,527 @@
...
+{
+    fn next_proposal(&self) -> Stm<Option<IPCParentFinality>> {
+        let height = if let Some(h) = self.cached_data.latest_height()? {
+            h
+        } else {
+            return Ok(None);
+        };
+
+        // safe to unwrap as latest height exists
+        let block_hash = self.cached_data.block_hash(height)?.unwrap();
```
aakoshh, line 112: It would be safer to bundle the height and the block into a single `atomically`, to avoid all possibility that between the two calls something else removes that block. It is what it's about, after all.

fendermint/vm/topdown/src/lib.rs
```diff
@@ -0,0 +1,73 @@
...
+pub type Bytes = Vec<u8>;
+pub type BlockHash = Bytes;
+
+#[derive(Debug, Clone, Deserialize)]
+pub struct Config {
+    /// The number of blocks to delay before reporting a height as final on the parent chain.
+    /// To propose a certain number of epochs delayed from the latest height, we see to be
+    /// conservative and avoid other from rejecting the proposal because they don't see the
+    /// height as final yet.
+    pub chain_head_delay: BlockHeight,
```
aakoshh, line 34: I don't fully understand the description. 

Say we have this chain, and we consider blocks 3 deep as final.
```
B0 <- B1 <- B2 <- B3 <- B4 <- B5 <- B6 <- B7 <- B8 <- B9
```
So in our case that means `B9` is the tip of the chain and `B6` is final, because there are 3 blocks built on top of it. 

We have a concern with using ABCI++ as a voting mechanism: if we propose `B6` too aggressively, it might not be seen as final for a quorum to be achieved, and the block doesn't get accepted. It's worth delaying the proposal a little to reduce this possibility. For example the 6 second delay of gossipsub seems reasonable, but I also thought about a height based approach, where we say "propose at 4 deep, accept at 3 deep", but it would add 30 seconds in case of Lotus, so maybe it's not as good as the wall clock based one.

Does `chain_head_delay` correspond to 3 in this scenario?

cryptoAtwill: Yep, it's corresponding to 3.

fendermint/vm/topdown/src/lib.rs
```diff
@@ -0,0 +1,73 @@
...
+    /// Parent syncing cron period, in seconds
+    pub polling_interval_secs: u64,
+    /// The ipc agent url
+    pub ipc_agent_url: String,
```
adlrocha, line 38: There is no `ipc_agent` anymore, think you can remove this one for the config.

fendermint/vm/topdown/src/sync.rs
```diff
@@ -0,0 +1,273 @@
...
+pub trait ParentFinalityStateQuery {
+    /// Get the latest committed finality from the state
+    fn get_latest_committed_finality(&self) -> anyhow::Result<Option<IPCParentFinality>>;
+}
+
+/// Constantly syncing with parent through polling
+pub struct PollingParentSyncer<T> {
+    config: Config,
+    parent_view_provider: Arc<Toggle<CachedFinalityProvider<IPCProviderProxy>>>,
+    agent: Arc<IPCProviderProxy>,
```
adlrocha, line 36: Do we want to still call this one `agent`? Should we may be call it `ipc` instead?

fendermint/vm/topdown/src/sync.rs
```diff
@@ -0,0 +1,273 @@
...
+    );
+
+    if last_recorded_height == ending_height {
+        tracing::debug!(
+            "the parent has yet to produce a new block, stops at height: {last_recorded_height}"
+        );
+        return Ok(());
+    }
+
+    // we are going backwards in terms of block height, the latest block height is lower
```
adlrocha, line 186: I guess for now this is good enough, it may have a bit of bad performance but I am personally fine with it.

adlrocha (approved):
High-level LGTM. I haven't gone in depth into the implementation of some of the function because I assume they are the same that we reviewed a while ago. Let me know if this is not the case and there's is something you want me to look depeer into.

fendermint/vm/topdown/src/sync.rs
```diff
@@ -0,0 +1,273 @@
...
+    // in `provider` and start from scratch
+    if last_recorded_height > ending_height {
+        let finality = query_starting_finality(query, agent_proxy).await?;
+        atomically(|| provider.reset(finality.clone())).await;
+        return Ok(());
+    }
+
+    // we are adding 1 to the height because we are fetching block by block, we also configured
+    // the sequential cache to use increment == 1.
+    let starting_height = last_recorded_height + 1;
```
aakoshh, line 197: I still think it would be a cheap and worthy step to start fetching from `last_recorded_height`, rather than one after, to be able to detect reorgs. 

The previous check `last_recorded_height > ending_height` established that the parent chain hasn't got shorter. If it did, the comment says it could be a reorg (or the parent node is re-syncing). But it could be longer and still have gone through a reorg:

```
B0 - B1 - B3 - B4 - B5
        \
          B3' - B4' - B5' - B6'
```

Say that `last_recorded_height` is 5 because we have fetched B5, and there was a reorg. Depending on when we poll, the following things could happen:
* we see `B4'` with height 4 and we clear the cache
* we see `B5'` with height 5 and we skip saying there are no new blocks
* we see `B6'` with height 6 and we fetch it and append it after our B5

It wouldn't be difficult to just fetch the block header at the `last_recorded_height` 5 and:
* if it doesn't exist any more because the chain got shorter, it's a reorg
* if it exists but differs, it's a reorg
* if it exists and matches, we can go and append new blocks

I know you commented earlier that reorgs are difficult to deal with but it seems perplexing to see the cache being reset on one signal of it happening but the obviously wrong thing being done on a different case, all just to save re-fetching one header.

fendermint/vm/topdown/src/sync.rs
aakoshh: Maybe `parent_client`?

Cargo.toml
adlrocha: Changed to `dev` after rebasing from `main`

fendermint/vm/topdown/src/lib.rs
adlrocha: fixed!

fendermint/vm/topdown/src/sync.rs
adlrocha: fixed!

fendermint/vm/topdown/src/sync.rs
```diff
@@ -0,0 +1,273 @@
...
+    }
+
+    let query = Arc::new(query);
+
+    let finality = query_starting_finality(&query, &agent).await?;
+
+    atomically(|| view_provider.set_new_finality(finality.clone())).await;
+
+    let poll = PollingParentSyncer::new(config, view_provider, agent, query);
+    poll.start();
```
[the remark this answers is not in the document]
adlrocha: fixed!

fendermint/vm/topdown/src/sync.rs
aakoshh: Created a more elaborate proposal in https://github.com/consensus-shipyard/fendermint/issues/314

adlrocha: Let me also track it in the code, so it is clear. Thanks!

fendermint/vm/topdown/src/sync.rs
```diff
@@ -0,0 +1,304 @@
...
+        } else {
+            None
+        };
+        Ok(b)
+    })
+    .await;
+    if let Some(blk) = last_committed_block {
+        if blk
+            != parent_proxy
+                .get_block_hash(last_recorded_height)
```
aakoshh, line 207: I think you are comparing different things here. 

`parent_proxy.get_block_hash(last_recorded_height)` is the correct thing to fetch, but it shouldn't be compared to the `provider.last_committed_finality()` because that one only moves when something is marked as final, while the other one moves every time we poll and there are new blocks. They do not necessarily point at the same thing. 

Somehow you would have to get from the cache the hash of the `last_recorded_height` and validate that it's still valid, and can be extended with new blocks.

aakoshh: You can see here that they aren't the same: https://github.com/consensus-shipyard/fendermint/pull/308/files#diff-2d785f545d077d6b6d959f0ccac11391a65407b67a677c47ff6ba83192edf7f0R254-R257

adlrocha: I see. I will revert the change and let @cryptoAtwill handle this one (I feel he has the full context and I am reading something wrong).

fendermint/vm/topdown/src/sync.rs
```diff
@@ -231,53 +241,77 @@ async fn reset_cache<T: ParentFinalityStateQuery + Send + Sync + 'static>(
...
     .await;
 
-    Ok(height)
+    Ok(result)
 }
 
 /// Obtain the new parent views for the input block height range
 async fn get_new_parent_views(
     parent_proxy: &Arc<IPCProviderProxy>,
+    previous_hash: BlockHash,
```
aakoshh, line 263: If you want you can just declare this as `mut previous_hash: BlockHash,` and then the `let mut previous_hash = previous_hash;` is not needed.

fendermint/vm/topdown/src/sync.rs
```diff
@@ -231,53 +241,77 @@ async fn reset_cache<T: ParentFinalityStateQuery + Send + Sync + 'static>(
...
         BlockHash,
         Vec<StakingChangeRequest>,
         Vec<CrossMsg>,
     )>,
+    Error,
 > {
     let mut block_height_to_update = vec![];
     let mut total_top_down_msgs = 0;
+
+    let mut previous_hash = previous_hash;
```
aakoshh, line 278: Maybe I'm blind: I can't see this every being e-assigned in the loop.

cryptoAtwill: Yeah, I forgot to add it, it was added.

PR #310, opened 2023-10-10
Topdown finality proposal and execution

cryptoAtwill:
Integrate topdown proposal and execution.

@aakoshh This is the previous approved PR with:
- Using validator change set
- Update configuration
- Initial simplified version for top down message execution. I understand there are more sophisticated methods for execution, but the current one is for integration testing to see execution with contract works. Once this is ok, will update to better methods such as left over gas in the block explicitly.


adlrocha (approved):
Looking good. Thanks! I am missing the gas management, but I would get this one merged first so we can start testing the end-to-end.

fendermint/app/config/default.toml
```diff
@@ -168,3 +168,8 @@ event_buffer_capacity = 100
 rate_limit_bytes = 0
 # Length of the time period at which the consumption limit fills. 0 means no limit.
 rate_limit_period = 0
+
+# IPC related configuration parameters
+[ipc]
+subnet_id = "/r0"
+network_name = "dev"
```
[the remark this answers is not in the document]
aakoshh: What is a `networ_name`?  IIRC @dnkolegov said it's the same as the subnet ID.
Sorry I can't keep it in my head :disappointed:

fendermint/app/settings/src/lib.rs
```diff
@@ -93,6 +97,52 @@ pub struct BroadcastSettings {
...
 
+#[derive(Debug, Deserialize, Clone)]
+pub struct TopDownConfig {
+    /// The number of blocks to delay before reporting a height as final on the parent chain.
+    /// To propose a certain number of epochs delayed from the latest height, we see to be
+    /// conservative and avoid other from rejecting the proposal because they don't see the
+    /// height as final yet.
+    pub chain_head_delay: BlockHeight,
+    /// Parent syncing cron period, in seconds
+    pub polling_interval_secs: u64,
```
aakoshh, line 108: Let's use the established way here: https://github.com/consensus-shipyard/fendermint/blob/main/fendermint/app/settings/src/eth.rs#L16-L17

fendermint/app/settings/src/lib.rs
```diff
@@ -93,6 +97,52 @@ pub struct BroadcastSettings {
...
+    /// Top down exponential back off retry base
+    pub exponential_back_off_secs: u64,
+    /// The max number of retries for exponential backoff before giving up
+    pub exponential_retry_limit: usize,
+    /// The ipc provider url
+    pub ipc_provider_url: String,
+    /// The parent registry address
+    #[serde(deserialize_with = "deserialize_eth_address_from_str")]
+    pub registry_address: Address,
```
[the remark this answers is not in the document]
aakoshh: The gateway and the registry do not have stable addresses on Lotus, right?

fendermint/app/settings/src/lib.rs
```diff
@@ -93,6 +97,52 @@ pub struct BroadcastSettings {
...
+    /// The parent gateway address
+    #[serde(deserialize_with = "deserialize_eth_address_from_str")]
+    pub gateway_address: Address,
+}
+
+#[serde_as]
+#[derive(Debug, Deserialize, Clone)]
+pub struct IPCSettings {
```
aakoshh, line 125:
```suggestion
pub struct IpcSettings {
```

So that we match existing capitalization [here](https://github.com/consensus-shipyard/fendermint/blob/main/fendermint/app/options/src/genesis.rs#L122) and [here](https://github.com/consensus-shipyard/fendermint/blob/main/fendermint/vm/genesis/src/lib.rs#L152] for example.

fendermint/app/src/app.rs
```diff
@@ -276,15 +283,55 @@ where
...
+            state,
+        ))
+        .await?;
         self.put_exec_state(state);
         Ok(ret)
     }
 
+    /// Get a read only fvm execution state. This is useful to perform query commands targeting
+    /// the latest state.
+    pub fn new_read_only_exec_state(
```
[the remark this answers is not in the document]
aakoshh: You may have seen this before in the predecessor PR that got closed, or maybe the one in the `check_tx` method. 

It doesn't exist in exactly the same format.

fendermint/app/src/cmd/run.rs
```diff
@@ -15,14 +15,45 @@ use fendermint_vm_interpreter::{
...
 use crate::cmd::key::read_secret_key;
 use crate::{cmd, options::run::RunArgs, settings::Settings};
 
+fn create_ipc_provider_proxy(settings: &Settings) -> anyhow::Result<IPCProviderProxy> {
+    let topdown_config = settings.ipc.topdown_config()?;
+
+    let url = topdown_config
+        .ipc_provider_url
+        .parse()
+        .context("invalid agent URL")?;
```
aakoshh, line 38: So, no more agent? 

It would be nice to move this parsing into the `settings` and use a `Url` type instead of `String`, to catch problems early. There is a type in `tendermint_rpc` we already use [here](https://github.com/consensus-shipyard/fendermint/blob/main/fendermint/app/settings/src/lib.rs#L107), perhaps it can do the trick.

fendermint/app/src/ipc.rs
```diff
@@ -0,0 +1,77 @@
...
+            let msg = encode_to_fvm_implicit(&evm_params)?;
+            tracing::debug!("query gateway parent finality message: {msg:?}");
+
+            let (apply_ret, _) = exec_state.execute_implicit(msg)?;
+
+            let data = apply_ret.msg_receipt.return_data.to_vec();
+            let decoded = fvm_ipld_encoding::from_slice::<BytesDe>(&data)
+                .map(|bz| bz.0)
+                .map_err(|e| anyhow!("failed to deserialize bytes returned by FEVM: {e}"))?;
+            Some(decode_parent_finality_return(decoded.as_slice())?)
```
aakoshh, line 70: Could this be added to the [GatewayCaller](https://github.com/consensus-shipyard/fendermint/blob/main/fendermint/vm/interpreter/src/fvm/state/ipc.rs#L56)?

fendermint/eth/hardhat/src/lib.rs
```diff
@@ -316,7 +316,11 @@ mod tests {
         let hardhat = test_hardhat();
 
         // Not giving any dependency should result in a failure.
-        let result = hardhat.bytecode("GatewayDiamond.sol", "GatewayDiamond", &Default::default());
```
aakoshh, line 319: Does this no longer have any dependencies?

fendermint/vm/interpreter/src/chain.rs
```diff
@@ -177,32 +217,87 @@ where
...
+                    if ret.is_err() {
+                        // TODO: how to handle error here? Do we return Err or stash them into one ret?
+                        todo!()
+                    }
+
+                    // stash validator changes
+
+                    // error happens if we cannot get the validator set from ipc agent after retries
+                    let validator_changes = provider.validator_changes(p.height as u64).await?;
+                    let msg = encode_to_fvm_implicit(&encode_store_validator_changes_call(
```
aakoshh, line 253: `encode_store_validator_changes_call` can go to `GatewayCaller` too.

fendermint/vm/interpreter/src/chain.rs
```diff
@@ -177,32 +217,87 @@ where
...
+                    }
+
+                    // Execute top down messages
+
+                    // error happens if we cannot get the validator set from ipc agent after retries
+                    let messages = provider.top_down_msgs(p.height as u64).await?;
+                    let msg = encode_to_fvm_implicit(&encode_apply_cross_messages_call(messages)?)?;
+                    let (state, ret) = self
+                        .inner
+                        .deliver(state, VerifiableMessage::NotVerify(msg))
```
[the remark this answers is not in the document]
aakoshh: Probably not here, though. We use implicit message execution, and expect the gateway to charge users for gas. By the time this returns it should be handled like all other transactions by the `inner`.

fendermint/vm/interpreter/src/chain.rs
```diff
@@ -177,32 +217,87 @@ where
...
+                    };
+                    let msg = encode_to_fvm_implicit(&encode_commit_parent_finality_call(
+                        finality.clone(),
+                    )?)?;
+                    let (state, ret) = self
+                        .inner
+                        .deliver(state, VerifiableMessage::NotVerify(msg))
+                        .await?;
+                    if ret.is_err() {
+                        // TODO: how to handle error here? Do we return Err or stash them into one ret?
```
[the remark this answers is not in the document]
aakoshh: Ah, wait, we are using the terms too loosely here. What do we mean by "return" exactly? 

My initial thought we can just  `return Err(anyhow!("failed to apply finality"))` and by doing that crash Fendermint, because if a legitimately voted upon finality fails to be added to the contract, we have a serious bug in the system. 

But you might argue it's better to keep chugging along, effectively severed from the parent subnet, in which case we can `return Ok(ret)` and go into terra incognita, see if we end up with a consensus failure or live to fight another day. 

Whichever it is, I don't think we should continue with the execution of messages.

fendermint/vm/interpreter/src/chain.rs
```diff
@@ -177,32 +217,87 @@ where
...
+                    let validator_changes = provider.validator_changes(p.height as u64).await?;
+                    let msg = encode_to_fvm_implicit(&encode_store_validator_changes_call(
+                        validator_changes,
+                    )?)?;
+                    let (state, ret) = self
+                        .inner
+                        .deliver(state, VerifiableMessage::NotVerify(msg))
+                        .await?;
+                    if ret.is_err() {
+                        // TODO: how to handle error here? Do we return Err or stash them into one ret?
```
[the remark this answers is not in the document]
aakoshh: What can be a legitimate legitimate reasons for this to fail?

So again we could crash saying this must be a bug, since these changes have been added to the parent ledger already, how can we _not_ add them on the child? I'm not sure how we could fix the contract though, if there was a bug, but maybe the bug is in the syncer?

Or we can choose to take it on the chin and `return Ok(ret)`, however the problem is that we have already committed the finality itself, so the syncer will think it doesn't have to re-fetch these changes, and then we have two possibilities:
* We have a bug in the syncer of this node instance, and everyone else applied the changes, in which case it's a consensus failure
* Nobody applied these changes, in which case it creates a gap in the ledger that might not exist on the parent, which can lead to unaccepted checkpoints

Quite possibly the best thing would be to apply all 3 changes as one atomic call to the contracts. I don't know about a way to selectively rolling back first transaction at this point.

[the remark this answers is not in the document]
aakoshh: I would choose to crash for the same reason I mentioned [here](https://github.com/consensus-shipyard/fendermint/pull/310/files#r1352782553).

fendermint/vm/interpreter/src/chain.rs
```diff
@@ -177,32 +217,87 @@ where
...
+
+                    // error happens if we cannot get the validator set from ipc agent after retries
+                    let messages = provider.top_down_msgs(p.height as u64).await?;
+                    let msg = encode_to_fvm_implicit(&encode_apply_cross_messages_call(messages)?)?;
+                    let (state, ret) = self
+                        .inner
+                        .deliver(state, VerifiableMessage::NotVerify(msg))
+                        .await?;
+                    if ret.is_err() {
+                        // TODO: how to handle error here? Do we return Err or stash them into one ret?
```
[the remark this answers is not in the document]
aakoshh: Once again, crashing might be our best option to avoid invalid state. 

I agree that the messages in the batch should be executed separately and their result should be emitted as events regardless of the outcome - at least I can't think of a better way of surfacing them. 

But the overall execution itself should not fail. If there is a reason it fails, we should fix it and resume from there, rather than sometimes skip entire batches because of a rotten apple or who knows what causes it to fail.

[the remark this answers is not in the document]
aakoshh: For example say it fails because even implicit executions respect the gas limit, which we set at `BLOCK_GAS_LIMIT`. Even though we talked about it when I proposed that messages should be appended to a queue, not immediately executed, and then executed as and when there is unused gas in the block, at this point it's just one big bang and it can fail. 

What then, do we just say oh well, sometimes depending on how many messages there are we skip entire batches? Or should we say we quickly patch the application to try to break it up into multiple batches? 

Of course crashing all nodes probably means the chain outage is going to be unacceptable, which leads me to back to my point that the best thing to do would be to treat the entire top-down execution as an atomic unit and mark it as failed, so that it can be re-tried again and again until it's fixed, but without crashing Fendermint or halting all other applications. 

Sorry, I know I am going back and forth from crashing to marking as failed, but in this case there is a legitimate reason to for failure, which is this overall gas limit. What we should not do is partial failure IMO.

fendermint/vm/interpreter/src/chain.rs
```diff
@@ -318,3 +413,23 @@ fn relayed_bottom_up_ckpt_to_fvm(
...
 }
+
+/// Encode to fvm implicit message
+pub fn encode_to_fvm_implicit(bytes: &[u8]) -> anyhow::Result<FvmMessage> {
+    let params = RawBytes::serialize(BytesSer(bytes))?;
+    let msg = FvmMessage {
+        version: 0,
+        from: system::SYSTEM_ACTOR_ADDR,
+        to: ipc::GATEWAY_ACTOR_ADDR,
+        value: TokenAmount::zero(),
```
aakoshh, line 424: @adlrocha can you walk me through what happens when somebody wants to `fund` their account by locking up tokens in the parent subnet and sending them to the child? 

I'd expect that somewhere we need to parse messages and recognise which ones are top-down messages (to=gateway, method=apply...), parse the cross messages, find the `fund` types, sum up their values and add them to the `circ_supply`. 

A similar thing should happen with bottom-up messages: `release` should lower the `circ_supply`. 

These are missing at the moment. 

Nevertheless, the message goes to the Gateway who is supposed to send the amount to the recipient Ethereum account. But can the Gateway _mint_ tokens, or can it only send what it has? Should perhaps this `value` here be used as a minting mechanism, so that incoming messages increase the balance of the `Gateway` itself, in order for it to be able to credit it to the funded accounts? Can the system account mint tokens?

fendermint/app/settings/src/lib.rs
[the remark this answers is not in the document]
adlrocha: Correct, that is why we need to offer a config for that (for L2+ subnets this will be easier as we can point to the deterministic address by Fendermint)

fendermint/app/src/app.rs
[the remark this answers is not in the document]
adlrocha: That may be it, thanks!

fendermint/vm/interpreter/src/chain.rs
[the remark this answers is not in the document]
aakoshh: Another relatively easy to implement option to prevent running out of gas while handling the batches is to do https://github.com/consensus-shipyard/fendermint/issues/261

fendermint/vm/interpreter/src/chain.rs
adlrocha: 1. When a user wants to fund an account in a subnet it calls [`fund` in the gateway of the parent](https://github.com/consensus-shipyard/ipc-solidity-actors/blob/e2624a563072a33b96ede3d3aafe0d78f87ed4ce/src/gateway/GatewayManagerFacet.sol#L130)
2. This function commits the [top-down message](https://github.com/consensus-shipyard/ipc-solidity-actors/blob/e2624a563072a33b96ede3d3aafe0d78f87ed4ce/src/gateway/GatewayManagerFacet.sol#L140), and stores the message for propagation in [the top down queue, updating the circulating supply, nonces, etc.](https://github.com/consensus-shipyard/ipc-solidity-actors/blob/63b3bf1f5f2774f31b25e4894982925d5a7a4672/src/lib/LibGateway.sol#L161-L165).

Fund is always top-down and release bottom-up, but when an arbitrary message arrives to the actor [`commitCrossMessage` is the one responsible for routing it as a top-down or a bottom-up](https://github.com/consensus-shipyard/ipc-solidity-actors/blob/63b3bf1f5f2774f31b25e4894982925d5a7a4672/src/gateway/GatewayMessengerFacet.sol#L76).

> But can the Gateway mint tokens, or can it only send what it has? Should perhaps this value here be used as a minting mechanism, so that incoming messages increase the balance of the Gateway itself, in order for it to be able to credit it to the funded accounts? Can the system account mint tokens?

This is a great question, what we did in Lotus is pre-allocate the maximum circulating supply that we want to allow in the subnet to the gateway, so the circulating supply is pre-minted and made available to the gateway to mind an burn when needed (actually, when we say burn what we should actually do is send the balance back to the gateway to avoid reducing the circulating supply).

That being said, now that we don't have the limitations from Lotus, I agree that it would be best to have ad-hoc minting of new funds here. It is a more elegant solution, and it prevents all of the pre-allocation and burning dance that we currently have.

fendermint/app/settings/src/lib.rs
```diff
@@ -93,6 +98,54 @@ pub struct BroadcastSettings {
...
+#[derive(Debug, Deserialize, Clone)]
+pub struct TopDownConfig {
+    /// The number of blocks to delay before reporting a height as final on the parent chain.
+    /// To propose a certain number of epochs delayed from the latest height, we see to be
+    /// conservative and avoid other from rejecting the proposal because they don't see the
+    /// height as final yet.
+    pub chain_head_delay: BlockHeight,
+    /// Parent syncing cron period, in seconds
+    #[serde_as(as = "DurationSeconds<u64>")]
+    pub polling_interval_secs: Duration,
```
aakoshh, line 111:
```suggestion
    pub polling_interval: Duration,
```
No need for saying anything about units if it's `Duration`.

fendermint/app/settings/src/lib.rs
```diff
@@ -93,6 +98,54 @@ pub struct BroadcastSettings {
...
+    /// Top down exponential back off retry base
+    #[serde_as(as = "DurationSeconds<u64>")]
+    pub exponential_back_off_secs: Duration,
```
aakoshh, line 114:
```suggestion
    pub exponential_back_off: Duration,
```

fendermint/vm/topdown/src/lib.rs
```diff
@@ -33,11 +34,9 @@ pub struct Config {
     /// height as final yet.
     pub chain_head_delay: BlockHeight,
     /// Parent syncing cron period, in seconds
-    pub polling_interval_secs: u64,
-    /// The endpoint to connect to the parent subnet
-    pub ipc_parent_endpoint: String,
+    pub polling_interval_secs: Duration,
     /// Top down exponential back off retry base
-    pub exponential_back_off_secs: u64,
+    pub exponential_back_off_secs: Duration,
```
aakoshh, line 39:
```suggestion
    pub exponential_back_off: Duration,
```

aakoshh (approved):
For posterity: I see we are crashing if any of the 3 parts of the execution of top-down finality goes wrong. 

I think that's not a bad way to go in the next two weeks, but I maintain that we should rethink our approach to execution to respect gas limits:
* by not voting on stuff over the limit https://github.com/consensus-shipyard/fendermint/pull/310#discussion_r1354360921
* by deferring execution to later blocks with free gas capacity
* by putting all three aspects of finality into a single call that fails together, so can retry in a later block if it failed, rather than crash

fendermint/vm/topdown/src/lib.rs
adlrocha: fixed.

fendermint/app/settings/src/lib.rs
adlrocha: fixed.

fendermint/app/settings/src/lib.rs
adlrocha: fixed.

adlrocha:
@cryptoAtwill, I addressed @aakoshh comments (remember to update you local Fendermint config when you test again), but I see this [PR](https://github.com/consensus-shipyard/fendermint/pull/321) depends on this one. @aakoshh, do you mind checking that one too see if we can merge it? Thanks :pray:

PR #311, opened 2023-10-11
FM-252: Apply validator changes and calculate diff

### LINKED ISSUE #252: ExecInterpreter::end adopt next validator set and return updates to Tendermint

aakoshh:
Closes #252 

At the end of the checkpointing period, call the validator update transition function in the ledger, then calculate the validator updates we can return to CometBFT.


fendermint/vm/interpreter/src/fvm/checkpoint.rs
```diff
@@ -212,3 +241,164 @@ where
...
+fn u256_to_tokens(value: U256) -> TokenAmount {
+    let mut bz = [0u8; 32];
+    value.to_big_endian(&mut bz);
+    let atto = BigInt::from_bytes_be(Sign::Plus, &bz);
+    TokenAmount::from_atto(atto)
+}
+
+/// Calculate the difference between the current and the next power table, to return to CometBFT only what changed.
+fn power_diff(current: &PowerTable, next: PowerTable) -> PowerUpdates {
+    let current = current
```
adlrocha, line 272: Maybe there's a type mismatch that I am missing, but why can't you use `into_power_map` here?

fendermint/vm/interpreter/src/fvm/checkpoint.rs
```diff
@@ -212,3 +241,164 @@ where
...
+        let insert = match current.get(&k) {
+            Some(w) if **w == v => None,
+            _ => Some(v),
+        };
+        if let Some(insert) = insert {
+            diff.push(insert);
+        }
+    }
+
+    PowerUpdates(diff)
```
adlrocha, line 308: Oh, I was having a hard time to see how this diff was applied. So you insert all of the changes beween `current` and `next`, indicating if something wants to be removed by setting their power by `0` (it wasn't obvious to me how you were reflecting validators leaving the membership) :)

adlrocha (approved):
LGTM. This is exciting, I think we are close to having a testable end-to-end :)

fendermint/vm/interpreter/src/fvm/checkpoint.rs
```diff
@@ -212,3 +241,164 @@ where
...
+                public_key: ValidatorKey(pk),
+                power: Collateral(c).into_power(power_scale),
+            })
+        }
+    }
+
+    PowerTable(pt)
+}
+
+fn u256_to_tokens(value: U256) -> TokenAmount {
```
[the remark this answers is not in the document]
aakoshh: Nice! I'm sure I have done this somewhere too... heah, [here](https://github.com/consensus-shipyard/fendermint/blob/995e193f3b97e816b2b9e020760312a4a5a25bab/fendermint/eth/api/src/conv/from_tm.rs#L45). I'm sure it's the easiest to debug, while this may be faster, who knows. 

I did it this way because it was symmetrical with [tokens_to_u256](https://github.com/consensus-shipyard/fendermint/blob/main/fendermint/vm/actor_interface/src/ipc.rs#L234), but the other way I also have a [safer](https://github.com/consensus-shipyard/fendermint/blob/995e193f3b97e816b2b9e020760312a4a5a25bab/fendermint/vm/message/src/conv/from_fvm.rs#L29) conversion that checks maximum boundaries. 

It would be nice if these were in a common place, just haven't found a good one yet.

fendermint/vm/interpreter/src/fvm/checkpoint.rs
aakoshh: Thanks for asking, the only reason is that `current` is a reference, while `into_power_map` consumes the input. I actually wanted to check if there is a way to unify these two because they are otherwise doing the same thing. 

`current` is a reference because:
* it's used later when constructing the Merkle tree, although maybe the tree could be constructed first
* I thought the values in it would not be used most of the time: the public key is only cloned if a validator is deleted

`next` is taken by value because:
* the value for new validators is going to be pushed into the output as-is
* it's not used for anything else

But I will have a second look because this is certainly jarring.

fendermint/vm/interpreter/src/fvm/checkpoint.rs
aakoshh: Yes, that's what CometBFT expects: https://github.com/informalsystems/tendermint-rs/blob/bcc0b377812b8e53a02dff156988569c5b3c81a2/rpc/src/dialect/end_block.rs#L12-L14

fendermint/vm/interpreter/src/fvm/checkpoint.rs
```diff
@@ -55,17 +65,35 @@ where
...
 
-            // TODO #252: Take the next validator changes from the gateway.
-            let power_updates = PowerUpdates(Vec::new());
+            // Apply any validator set transitions.
+            let next_configuration_number = gateway
+                .apply_validator_changes(state)
+                .context("failed to apply validator changes")?;
+
+            // Figure out the power updates if there was some change in the configuration.
+            let power_updates = if next_configuration_number == 0 {
```
aakoshh, line 80: By the way @adlrocha do you think it will cause any issues for Solidity if we send 0 in the checkpoint to show "nothing changed"?

fendermint/vm/interpreter/src/fvm/checkpoint.rs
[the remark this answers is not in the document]
adlrocha: Yep, I think you approach is better. Maybe we can have these either in the actors bindings or in the `ipc-agent/sdk`. That being said, for now I think is fine where it is.

fendermint/vm/interpreter/src/fvm/checkpoint.rs
adlrocha: I don't think is a problem (that is what I am returning myself when applying the finality). What I am wondering now is where are we confirming the new membership in the parent when the bottom-up checkpoint is committed, I can't find anything [here](https://github.com/consensus-shipyard/ipc-solidity-actors/blob/f492531ae7d441b4bb00ca930f5d066956fadf4f/src/subnet/SubnetActorManagerFacet.sol#L40).

@cryptoAtwill, @dnkolegov, do you mind chiming in here. I am missing something or are we not confirming the changes in membership yet and we should? (I think it is the latter after reading the code, this should be an easy fix, but is a must :) )

aakoshh: Thanks for the link @adlrocha , it indeed looks like the membership is not altered to reflect the new configuration number. We discussed that once that is done, we can't use `validateActiveQuorumSignature` any more and Denis updated the method you link to compare hashes, but the actual transition is missing :mag:

adlrocha: @dnkolegov can you pick this one up, if not I think I can pick it up when I wrap my current PR (I am updating the contract bindings in `ipc-agent`): https://github.com/consensus-shipyard/ipc-agent/pull/334

fendermint/vm/interpreter/src/fvm/checkpoint.rs
aakoshh: Decided to rearrange the code so both arguments are consumed. Originally I also returned the power table from `maybe_create_checkpoint` because it was used to construct the proof, but since we re-send the validator signature for every potentially pending checkpoint, I just re-fetch it and ignored the returned value to keep the code simpler. Not a big deal.

fendermint/vm/interpreter/src/fvm/checkpoint.rs
dnkolegov: ok

dnkolegov: Here - https://github.com/consensus-shipyard/ipc-solidity-actors/blob/6322ae6a236859cedcc32d6ff7dabf4d709695e8/src/subnet/SubnetActorManagerFacet.sol#L73

PR #317, opened 2023-10-16
FM-306: Create child subnet genesis from parent

### LINKED ISSUE #306: Generate genesis.json file from parent
Once the minimum requirements to bootstrap a subnet are met in the parent in https://github.com/consensus-shipyard/ipc-solidity-actors/pull/223 and https://github.com/consensus-shipyard/ipc-solidity-actors/pull/228 we can generate a Genesis file for the subnet. 

Currently this is in the form of a `genesis.json` file produced by the `fendermint` CLI. The idea is that instead of putting it together by adding accounts and validators, setting IPC parameters, etc, all the data required is fetched from the parent chain and written to the JSON file, with which CometBFT can be started as we have done before. 

One possible way to do this is:
* There should be some CLI in Fendermint that can produce a genesis.json file from some subnet at a specific height in the parent subnet.
* It needs as input the SubnetID, and optionally the block height
* Ideally it should verify that at that height the "bootstrapped" event was emitted.
* It could fetch the logs of the SubnetActor contract and check at which block height it was emitted, so the user doesn't need to know.
* It should query some methods in the SubnetActor to fetch the IPC parameters such as: the bottomUpCheckpointPeriod, the initial validator set, ...
* Set the chain name to the subnet ID in string form.
* Set the circulating supply to 0
* We can assume no account have any balance in the genesis, and will fund their accounts later with top-down messages
* We can create accounts (f1 or f410) for the initial validators according to the address they use in the parent.

Questions:
* Do we go to the Ethereum API directly, or do we use the IPC SDK for this? Or both?
* Which Solidity methods need to be called?
* How can we figure out the Ethereum delegated address of a SubnetActor? Ask the SubnetRegistry? How?

This ticket enables #301

adlrocha:
Closes #306 

This PR implements a new genesis command to generate the genesis file for a child subnet from the information available for the child in the parent (after it has been bootstrapped successfully).

### DIFF
--- a/docs/running.md
+++ b/docs/running.md
@@ -667,3 +667,43 @@ $ cargo run -p fendermint_rpc --release --example simplecoin -- --secret-key tes
 ```
 
 Note that the script figures out the Alice's nonce on its own, so we don't have to pass it in. It also has an example of running an EVM view method (which is read-only) either as as a distributed read-transaction (which is included on the chain and costs gas) or a query anwered by our node without involving the blockchain. Both have their uses, depending on our level of trust.
+
+## Deploy IPC child subnet
+
+### Crate genesis from parent
+Fendermint includes a command to automatically create the genesis file for an IPC child subnet according to the information for the subnet available in its parent. Here's an example of the generation of a genesis file for a subnet that has already been bootstrapped in the parent.
+```console
+cargo run -p fendermint_app -- --network=test genesis --genesis-file test-network/genesis.json ipc child-genesis --subnet-id <CHILD_SUBNET_ID> -p <PARENT_ENDPOINT> --parent-gateway <PARENT_GATEWAY_CONTRACT> --parent-registry <PARENT_REGISTRY_CONTRACT>
+```
+
+Here's a sample execution of the command for an already bootstrapped subnet in `/r314159`:
+```console
+cargo run -p fendermint_app -- --network=test genesis --genesis-file test-network/genesis.json ipc child-genesis --subnet-id /r314159/t410fdoh27lsddz4my2v3e77qnxdp5vsjxkdfokc7sti -p https://api.calibration.node.glif.io/rpc/v1 --parent-gateway 0x56948d2CFaa2EF355B8C08Ac925202db212146D1 --parent-registry 0x6A4884D2B6A597792dC68014D4B7C117cca5668e
+```
+
+Leading to the following genesis file:
+```console
+{
+  "chain_name": "/r314159/t410fdoh27lsddz4my2v3e77qnxdp5vsjxkdfokc7sti",
+  "timestamp": 1007560,
+  "network_version": 18,
+  "base_fee": "1000",
+  "power_scale": 3,
+  "validators": [
+    {
+      "public_key": "BDOFw7mriml8177GymI8vdD+oSk+i0ZN+CWxBOtYpEzI76zGo0grhmuF7N9zS11O9UlXN96zSGJc5qNVNhQtKVU=",
+      "power": "1000000000000000000"
+    }
+  ],
+  "accounts": [],
+  "ipc": {
+    "gateway": {
+      "subnet_id": "/r314159/t410fdoh27lsddz4my2v3e77qnxdp5vsjxkdfokc7sti",
+      "bottom_up_check_period": 30,
+      "msg_fee": "1000000000000",
+      "majority_percentage": 60,
+      "min_collateral": "1000000000000000000",
+      "active_validators_limit": 100
+    }
+}
+```
--- a/docs/running.md
+++ b/docs/running.md
@@ -672,14 +672,25 @@ Note that the script figures out the Alice's nonce on its own, so we don't have 
 
 ### Crate genesis from parent
 Fendermint includes a command to automatically create the genesis file for an IPC child subnet according to the information for the subnet available in its parent. Here's an example of the generation of a genesis file for a subnet that has already been bootstrapped in the parent.
-```console
-cargo run -p fendermint_app -- --network=test genesis --genesis-file test-network/genesis.json ipc from-parent --subnet-id <CHILD_SUBNET_ID> -p <PARENT_ENDPOINT> --parent-gateway <PARENT_GATEWAY_CONTRACT> --parent-registry <PARENT_REGISTRY_CONTRACT>
+```shell
+cargo run -p fendermint_app -- \
+    --network=test \
+    genesis --genesis-file test-network/genesis.json \
+    ipc from-parent --subnet-id <CHILD_SUBNET_ID> -p <PARENT_ENDPOINT> \
+    --parent-gateway <PARENT_GATEWAY_CONTRACT> \
+    --parent-registry <PARENT_REGISTRY_CONTRACT>
 ```
 
 Here's a sample execution of the command for an already bootstrapped subnet in `/r314159`:
-```console
-cargo run -p fendermint_app -- --network=test genesis --genesis-file test-network/genesis.json ipc from-parent --subnet-id /r314159/t410fdoh27lsddz4my2v3e77qnxdp5vsjxkdfokc7sti -p https://api.calibration.node.glif.io/rpc/v1 --parent-gateway 0x56948d2CFaa2EF355B8C08Ac925202db212146D1 --parent-registry 0x6A4884D2B6A597792dC68014D4B7C117cca5668e
-```
+```shell
+cargo run -p fendermint_app -- \
+    --network=test \
+    genesis --genesis-file test-network/genesis.json \
+    ipc from-parent \
+    --subnet-id /r314159/t410fdoh27lsddz4my2v3e77qnxdp5vsjxkdfokc7sti \
+    -p https://api.calibration.node.glif.io/rpc/v1 \
+    --parent-gateway 0x56948d2CFaa2EF355B8C08Ac925202db212146D1 \
+    --parent-registry 0x6A4884D2B6A597792dC68014D4B7C117cca5668e```
 
 Leading to the following genesis file:
 ```console
--- a/docs/running.md
+++ b/docs/running.md
@@ -673,12 +673,12 @@ Note that the script figures out the Alice's nonce on its own, so we don't have 
 ### Crate genesis from parent
 Fendermint includes a command to automatically create the genesis file for an IPC child subnet according to the information for the subnet available in its parent. Here's an example of the generation of a genesis file for a subnet that has already been bootstrapped in the parent.
 ```console
-cargo run -p fendermint_app -- --network=test genesis --genesis-file test-network/genesis.json ipc child-genesis --subnet-id <CHILD_SUBNET_ID> -p <PARENT_ENDPOINT> --parent-gateway <PARENT_GATEWAY_CONTRACT> --parent-registry <PARENT_REGISTRY_CONTRACT>
+cargo run -p fendermint_app -- --network=test genesis --genesis-file test-network/genesis.json ipc from-parent --subnet-id <CHILD_SUBNET_ID> -p <PARENT_ENDPOINT> --parent-gateway <PARENT_GATEWAY_CONTRACT> --parent-registry <PARENT_REGISTRY_CONTRACT>
 ```
 
 Here's a sample execution of the command for an already bootstrapped subnet in `/r314159`:
 ```console
-cargo run -p fendermint_app -- --network=test genesis --genesis-file test-network/genesis.json ipc child-genesis --subnet-id /r314159/t410fdoh27lsddz4my2v3e77qnxdp5vsjxkdfokc7sti -p https://api.calibration.node.glif.io/rpc/v1 --parent-gateway 0x56948d2CFaa2EF355B8C08Ac925202db212146D1 --parent-registry 0x6A4884D2B6A597792dC68014D4B7C117cca5668e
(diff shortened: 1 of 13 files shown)


fendermint/app/src/cmd/genesis.rs
```diff
@@ -246,3 +252,64 @@ fn set_ipc_gateway(genesis_file: &PathBuf, args: &GenesisIpcGatewayArgs) -> anyh
...
+            majority_percentage: genesis_info.majority_percentage,
+            active_validators_limit: genesis_info.active_validators_limit,
+        },
+    };
+    let genesis = Genesis {
+        // We set the genesis epoch as the genesis timestamp so it can be
+        // generated deterministically by all participants
+        // genesis_epoch should be a positive number, we can afford panicking
+        // here if this is not the case.
+        timestamp: Timestamp(genesis_info.genesis_epoch.try_into().unwrap()),
```
aakoshh, line 292: Slightly unfortunate that we call it `epoch`, which is height in Filecoin. I try to avoid using that name in Fendermint for other stuff. What is `genesis_epoch` exactly?

fendermint/app/src/cmd/genesis.rs
```diff
@@ -246,3 +252,64 @@ fn set_ipc_gateway(genesis_file: &PathBuf, args: &GenesisIpcGatewayArgs) -> anyh
...
+        chain_name: args.subnet_id.to_string(),
```
aakoshh, line 293: Shouldn't this be derived from the parent and the address of the subnet actor?

fendermint/app/src/cmd/genesis.rs
```diff
@@ -246,3 +252,64 @@ fn set_ipc_gateway(genesis_file: &PathBuf, args: &GenesisIpcGatewayArgs) -> anyh
...
+        accounts: Vec::new(),
+        ipc: Some(ipc_params),
+    };
+
+    for v in genesis_info.validators {
+        let pk = PublicKey::parse_slice(&v.metadata, None)?;
+        let c = to_fvm_tokens(&v.weight);
+        genesis.validators.push(Validator {
+            public_key: ValidatorKey(pk),
+            power: Collateral(c).into_power(genesis.power_scale),
```
aakoshh, line 307: Maybe I am confusing myself here, but AFAIR `genesis.validators` is has type `Vec<Validator<Collateral>>`. Why are you converting to `Power` here? I thought this wouldn't even compile.

fendermint/app/src/cmd/genesis.rs
adlrocha: We cal it epoch because it comes directly from the state of the subnet actor in the parent, and we call it `epoch` in the actors. We could change it in the `IpcProvider` but it seems like a unnecessary overhead (unless you strongly think otherwise).

fendermint/app/src/cmd/genesis.rs
adlrocha: It has already been derived in the parent, we just pass the result as an argument here. There are lots of subnet actors deployed in the parent, and we need to signal for which one we want to generate the genesis file for.

fendermint/app/src/cmd/genesis.rs
adlrocha: Outdated code, it was not pushed before. Sorry.

fendermint/app/src/cmd/genesis.rs
aakoshh: Sorry I should have been more specific. What I thought was happening is that we have a parent subnet, say, `/r123` and in there we create a subnet which will:
* be an EVM contract with an ID such as `f4101a2b3c...4d5e6f`
* have the subnet ID `/r123/f4101a2b3c...4d5e6f`

Now we want to grab its state to produce a `genesis.json` file for the child subnet via CLI. The set of parameters that minimises any error potential would look something like this:
```
ipc-cli genesis from-parent --parent-url http://mylotus.com:3456 --subnet-actor-addr f4101a2b3c...4d5e6f
```

The `/r123` part of the subnet ID is functionally defined by the `--parent-url`, so all you need is to find the subnet actor, which has a unique ID within this ledger. 

If, on the other hand, we use this:
```
ipc-cli genesis from-parent --parent-url http://mylotus.com:3456 --subnet-id /r123/f4101a2b3c...4d5e6f
```
and then put the subnet ID from the `args` into `genesis.json` then it's a bit more suspicious whether the subnet ID really belongs to the actor. It depends on how it was fetched, ie. whether the full path was taken into account, or just the last leg.

I'm sure it's fine, just wanted to point out that it's a bit over specified.

fendermint/app/src/cmd/genesis.rs
```diff
@@ -246,3 +252,64 @@ fn set_ipc_gateway(genesis_file: &PathBuf, args: &GenesisIpcGatewayArgs) -> anyh
...
+
+async fn new_child_genesis(
+    genesis_file: &PathBuf,
+    args: &GenesisIpcChildGenesisArgs,
+) -> anyhow::Result<()> {
+    // provider with the parent.
+    let parent_provider = IpcProvider::new_with_subnet(
+        None,
+        ipc_provider::config::Subnet {
+            id: args.subnet_id.clone(),
```
aakoshh, line 263: What does `id` do? Why isn't the the subnet ID of the _parent_ subnet, within which we can inquire about the child subnets?

fendermint/app/src/cmd/genesis.rs
aakoshh: Oh well. I wouldn't have chosen this contested name for the timestamp, but ho hum.

fendermint/app/options/src/parse.rs
```diff
@@ -94,3 +94,10 @@ pub fn parse_network(s: &str) -> Result<Network, String> {
         }
     }
 }
+
+pub fn parse_eth_address(s: &str) -> Result<Address, String> {
+    match primitives::EthAddress::from_str(&s) {
```
aakoshh, line 99: What was wrong with the `ethers::core::Address` type?

fendermint/app/src/cmd/genesis.rs
aakoshh: I will not argue against using full subnet IDs whenever they appear, seems like a decent default choice, even though strictly speaking when all operations can only go either to the parent, or to one of the immediate children, an Ethereum address would also do, but at least this way it's easy to tell what it is, and also easy to figure out if it's pointing at the parent or a child. I'll leave UX to others. 

Just have to make sure we always validate that the current subnet is on the path.

adlrocha: I see your point, and you are right. UX is something I am not happy about, tbh, and that needs more care. Hopefully we get help for this soon. 

In any case, I will tinker a bit with it, let's see what is more natural.

fendermint/app/src/cmd/genesis.rs
adlrocha: Bug from my side, already fixed.

fendermint/app/options/src/parse.rs
adlrocha: `ethers::core::Address` didn't work directly, and for the provider config we use `f4` addresses, so it was easier to leverage our auxiliary type. But I am not strongly attached to it if you feel there is a simpler way of doing it.

fendermint/app/src/cmd/genesis.rs
adlrocha: I actually renamed the command to `from-parent`, I think is better. Will push in a minute. I am testing it

docs/running.md
```diff
@@ -667,3 +667,43 @@ $ cargo run -p fendermint_rpc --release --example simplecoin -- --secret-key tes
 ```
 
 Note that the script figures out the Alice's nonce on its own, so we don't have to pass it in. It also has an example of running an EVM view method (which is read-only) either as as a distributed read-transaction (which is included on the chain and costs gas) or a query anwered by our node without involving the blockchain. Both have their uses, depending on our level of trust.
+
+## Deploy IPC child subnet
+
+### Crate genesis from parent
+Fendermint includes a command to automatically create the genesis file for an IPC child subnet according to the information for the subnet available in its parent. Here's an example of the generation of a genesis file for a subnet that has already been bootstrapped in the parent.
+```console
```
aakoshh, line 675:
```suggestion
```shell
```

Let's reserve `console` when you want to show output, in which case your command would be prefixed with `$`.

docs/running.md
```diff
@@ -667,3 +667,43 @@ $ cargo run -p fendermint_rpc --release --example simplecoin -- --secret-key tes
...
+cargo run -p fendermint_app -- --network=test genesis --genesis-file test-network/genesis.json ipc from-parent --subnet-id <CHILD_SUBNET_ID> -p <PARENT_ENDPOINT> --parent-gateway <PARENT_GATEWAY_CONTRACT> --parent-registry <PARENT_REGISTRY_CONTRACT>
+```
+
+Here's a sample execution of the command for an already bootstrapped subnet in `/r314159`:
+```console
```
aakoshh, line 680:
```suggestion
```shell
```

docs/running.md
```diff
@@ -667,3 +667,43 @@ $ cargo run -p fendermint_rpc --release --example simplecoin -- --secret-key tes
 ```
 
 Note that the script figures out the Alice's nonce on its own, so we don't have to pass it in. It also has an example of running an EVM view method (which is read-only) either as as a distributed read-transaction (which is included on the chain and costs gas) or a query anwered by our node without involving the blockchain. Both have their uses, depending on our level of trust.
+
+## Deploy IPC child subnet
+
+### Crate genesis from parent
+Fendermint includes a command to automatically create the genesis file for an IPC child subnet according to the information for the subnet available in its parent. Here's an example of the generation of a genesis file for a subnet that has already been bootstrapped in the parent.
+```console
+cargo run -p fendermint_app -- --network=test genesis --genesis-file test-network/genesis.json ipc from-parent --subnet-id <CHILD_SUBNET_ID> -p <PARENT_ENDPOINT> --parent-gateway <PARENT_GATEWAY_CONTRACT> --parent-registry <PARENT_REGISTRY_CONTRACT>
```
aakoshh, line 676:
```suggestion
cargo run -p fendermint_app -- \
  --network=test \
  genesis --genesis-file test-network/genesis.json ipc \
    from-parent \
      --subnet-id <CHILD_SUBNET_ID> \
      -p <PARENT_ENDPOINT> \
      --parent-gateway <PARENT_GATEWAY_CONTRACT> \
      --parent-registry <PARENT_REGISTRY_CONTRACT>
```
So readers don't have to scroll in the box. 

By the way I found out that there is some way to tell `clap` that it should accept the options regardless of where they appear, so instead of `fendermint --network ... genesis --genesis-file ... ipc from-parent ...` we could type `fendermint genesis ipc from-parent --network ... --genesis-file ...`. Not sure which one is easier to get right.

docs/running.md
```diff
@@ -667,3 +667,43 @@ $ cargo run -p fendermint_rpc --release --example simplecoin -- --secret-key tes
...
+```
+
+Here's a sample execution of the command for an already bootstrapped subnet in `/r314159`:
+```console
+cargo run -p fendermint_app -- --network=test genesis --genesis-file test-network/genesis.json ipc from-parent --subnet-id /r314159/t410fdoh27lsddz4my2v3e77qnxdp5vsjxkdfokc7sti -p https://api.calibration.node.glif.io/rpc/v1 --parent-gateway 0x56948d2CFaa2EF355B8C08Ac925202db212146D1 --parent-registry 0x6A4884D2B6A597792dC68014D4B7C117cca5668e
```
aakoshh, line 681:
```suggestion
cargo run -p fendermint_app -- \
  --network=test \
  genesis --genesis-file test-network/genesis.json \
  ipc from-parent \
    --subnet-id /r314159/t410fdoh27lsddz4my2v3e77qnxdp5vsjxkdfokc7sti \
    -p https://api.calibration.node.glif.io/rpc/v1 \
    --parent-gateway 0x56948d2CFaa2EF355B8C08Ac925202db212146D1 \
    --parent-registry 0x6A4884D2B6A597792dC68014D4B7C117cca5668e
```

fendermint/app/options/src/genesis.rs
```diff
@@ -149,3 +149,34 @@ pub struct GenesisIpcGatewayArgs {
...
+pub struct GenesisFromParentArgs {
+    /// Child subnet for with the genesis file is being created
+    #[arg(long, short)]
+    pub subnet_id: SubnetID,
+
+    /// Endpoint to the RPC of the child subnet's parent
+    #[arg(long, short)]
+    pub parent_endpoint: url::Url,
+
+    /// IPC gateway of the parent
```
aakoshh, line 163:
```suggestion
    /// IPC gateway of the parent; 20 byte Ethereum address in 0x prefixed hex format
```

fendermint/app/options/src/genesis.rs
```diff
@@ -149,3 +149,34 @@ pub struct GenesisIpcGatewayArgs {
...
+    #[arg(long, value_parser = parse_eth_address)]
+    pub parent_gateway: Address,
+
+    /// IPC registry of the parent
```
aakoshh, line 167:
```suggestion
    /// IPC registry of the parent; 20 byte Ethereum address in 0x prefixed hex format
```

aakoshh: Should we give these the default values we use Fendermint?

fendermint/app/options/src/genesis.rs
aakoshh: Maybe worth mentioning how one can figure this out when they have to deal with Lotus.

aakoshh (approved):
Looks good! :checkered_flag:

PR #319, opened 2023-10-18
Test staking: Part 2

aakoshh:
Continuation of https://github.com/consensus-shipyard/fendermint/pull/299

- [x] apply update immediately until activated
- [x] activate on collateral reached
- [x] `join` and `leave` command
- [x] implement join and leave on the model state
- [x] [implement `unstake` in the contract](https://github.com/consensus-shipyard/ipc-solidity-actors/issues/256)
- [x] call the SUT for `join`
- [x] call the SUT for `stake`
- [x] call the SUT for `leave`
- [x] call the SUT for `unstake`
- [x] call the SUT for `checkpoint`
- [x] validate balances
- [x] validate stakes
- [x] validate top N validators
- [x] generate a non-empty cross message vector to check hashing

## Fixes:
* On the contract side https://github.com/consensus-shipyard/ipc-solidity-actors/pull/255 
* Use non-masked ID for gateway address
* Pass the checkpoint as a tuple to ABI hashing
* Add 27 to the signature recovery byte
* Various fixes on the staking test model

## Running

The tests can be executed as follows:

```shell
cargo test --release -p contract-test --test smt_staking 
```

To run them on modified Solidity contracts, change the `Cargo.toml` file like so:

```toml
[patch."https://github.com/consensus-shipyard/ipc-solidity-actors.git"]
ipc_actors_abis = { path = "../ipc-solidity-actors/binding" }
```

And then run `make ipc-actors-abi` to recompile the ABI artifacts.

### Ranking

I cannot replicate the exact ordering of validators when multiple validators have the same collateral, which can lead to disagreements between Rust and Solidity about who exactly are the top N validators, when validator N and N+1 both have the same collateral. 

I used a stable sorting mechanism hoping that it will work fairly: given a sorted list of validators, we update the collateral, and the move them up or down to their new place, leaving others in place. 

However, in Solidity the priority queues are unstable. For example:
* Say we we have the following values in a min-heap: `[1a, 2b, 2c, 3d, 3e, 3f]`, where the letter is the validator identity and the number is their collateral
* Say we want to pop/delete the minimum value `1a`
* The algorithm first swaps `1a` with the last element `3f`, resulting in `[3f, 2b, 2c, 3d, 3e, 1a]`. Then it removes the last element and makes the queue shorter. Finally it sinks the first element back into its place, resulting in `[2b, 2c, 3f, 3d, 3e]`. 
* So `3f` moved from being the last to the top of the "threes". 
* By contrast, in Rust we just remove `1a` from the queue if it leaves, leaving the others in their relative order.

If this situation arises, I stop testing the system further and start with a fresh state. I don't think it's worth trying to replicate this exact ordering mechanism.


fendermint/testing/contract-test/src/ipc/subnet.rs
```diff
@@ -50,4 +56,104 @@ impl<DB: Blockstore> SubnetCaller<DB> {
...
+
+    /// Try to join the subnet as a validator.
+    pub fn try_join(
+        &self,
+        state: &mut FvmExecState<DB>,
+        validator: &Validator<Collateral>,
+    ) -> TryCallResult<()> {
+        let public_key = validator.public_key.0.serialize();
+        let addr = EthAddress::new_secp256k1(&public_key)?;
+        let deposit = from_fvm::to_eth_tokens(&validator.power.0)?;
```
adlrocha, line 68: I like this `from_fvm` and `from_eth` conversion utils. I feel we should expose them as a crate and make them available to the `ipc` repo. I feel we have a lot of redundant conversion functions here and there :) 

//cc @cryptoAtwill, thoughts?

fendermint/testing/contract-test/tests/staking/machine.rs
```diff
@@ -122,65 +176,278 @@ impl StateMachine for StakingMachine {
...
+            }
+            StakingCommand::Leave(addr) => {
+                if !pre_state.has_staked(addr) {
+                    result.expect_err("must call join before leave");
+                } else {
+                    result.expect("leave should succeed");
+                }
+            }
+            StakingCommand::Unstake(_addr, _value) => {
+                todo!("implement unstake in the contract")
```
adlrocha, line 396: https://github.com/consensus-shipyard/ipc-solidity-actors/issues/256

adlrocha (approved):
These tests are actual gold, @aakoshh. Thanks! This makes testing significantly easier than what we were doing pre-Fendermint :heart:

fendermint/testing/contract-test/src/ipc/subnet.rs
aakoshh: They are exposed in the `messages` crate, but some low level ones might be useful on their own for someone who doesn't want to deal with Fendermint message types.

Issue #312, opened 2023-10-12
Use bootstrap nodes for child subnets


adlrocha:
## Background
As brought up by @aakoshh, we shouldn't expose validator's network addresses publicly to avoid potential attacks. This leaves us with the only option of using bootstrap nodes for new validators to be able to discover the rest of members in the subnet. 

## Solution
My proposed solution is to: 
* Allow validators to deploy their own bootstrap nodes and advertise them through the subnet actor in the parent.
* We should include a `set_bootstrap(net_address: String)` function in the the Subnet actor for validators to be able to add their bootstrap. 
* When a validator leaves the subnet, we should remove their bootstrap to ensure that we keep fresh bootsraps in the list. 
* As part of this task, we should also add some scripts that aide with the deployment of bootstraps, and test the configuration that we need to set in validators to enable an easy deployment of a child subnets infra. 

## Discussion
I don´t have a strong opinion on the fact that we are only allowing to set bootsraps to validators. It felt right to just allow running bootstraps with some "skin in the game" to avoid users DDoS with fake bootstraps, but we can simplify a bit the model if needed.

aakoshh:
Just to move some of my comments here:
* My initial thought was an optional address at the time the subnet is created, but it's not clear who could change it later
* The validator's voluntary signaling answers that question
* The [IPLD Resolver](https://github.com/consensus-shipyard/ipc-ipld-resolver) gossips around which node can serve data for what subnets - if it was possible to derive a CometBFT address from the mutliaddress where the IPLD Resolver itself is listening, it could be used to discover nodes to connect to on any subnet. However, there is no guarantee that they have `pex` enabled, or they are running on the port we assume they do. Nevertheless this would mean there would be only 2 seed addresses for the entire hierarchy. 

At the risk of stating the obvious, these addresses should be removed if the validator leaves the subnet completely.

dnkolegov:
> 
When a validator leaves the subnet, we should remove their bootstrap to ensure that we keep fresh bootsraps in the list.

What if there is only one bootstrap node and the corresponding validator is leaving?

adlrocha:
> What if there is only one bootstrap node and the corresponding validator is leaving?

My take is that then validators should worry about booting up a new bootstrap if they want to be discoverable. We could think of an economic incentive for bootstrap, where validators with bootstrap get higher rewards, but at this point I would keep it simple and consider it a best effort system.

dnkolegov:
[...]

PR #320, opened 2023-10-20
IPC-333: child validator and bootstrap deployment scripts

adlrocha:
Closes https://github.com/consensus-shipyard/ipc-agent/issues/333 and https://github.com/consensus-shipyard/fendermint/issues/312

This PR:
- Adds convenient `cargo-make` scripts to deploy a bootstrap node for child subnets and to spawn child subnet validators, and a set of bug fixes.

## Bug fixes
 - [x] Ensuring that the `chain_name` doesn't overflow the allows `MAX_LENGTH`. 
 - [x] Bundling the IPC contracts into the docker container. 
 - [x] In the exposure of ports by the infra script.


fendermint/app/src/cmd/genesis.rs
```diff
@@ -189,9 +191,17 @@ where
 fn into_tendermint(genesis_file: &PathBuf, args: &GenesisIntoTendermintArgs) -> anyhow::Result<()> {
     let genesis = read_genesis(genesis_file)?;
     let genesis_json = serde_json::to_value(&genesis)?;
+    // Fendermint chainIDs can only be 50 bytes long. To avoid
+    // arbitrary chain names from exceeding the limit, we compute
+    // the chainID as the 54 most significant bytes of the SHA256
```
dnkolegov, line 196: Where are the other 4 bytes?

infra/scripts/node.toml
```diff
@@ -1,19 +1,44 @@
...
     "cometbft-stop",
     "cometbft-rm"
 ]
 
+
 [tasks.new-node]
 dependencies = [
     "node-down",
-    "node-init",
+    "bootstrap-mkdir",
```
dnkolegov, line 35: why is it called `bootstrap-mkdir` and is used in `new-node`?

fendermint/app/src/cmd/genesis.rs
adlrocha: This is a typo, it should say 50 in both, I will update accordingly. Fixed!

infra/scripts/node.toml
adlrocha: I thought `new-node` was being used by some other script. It turns out it is not. I just wanted to do `node-mkdir` without creating a fendermint directory, this is why I wrote a new action for the bootstrap, but I realized it is not even needed because no one else is using it. I updated it accordingly.

dnkolegov (approved):

docker/local.Dockerfile
```diff
@@ -37,6 +37,6 @@ ENV FM_ETH__LISTEN__HOST=0.0.0.0
 # We could build the actor bundles in the `builder` as well,
 # but we should be able to copy it from somewhere.
 COPY docker/.artifacts/bundle.car $FM_HOME_DIR/bundle.car
-COPY docker/.artifacts/contracts $FM_HOME_DIR/contracts
+COPY docker/.artifacts/contracts/out $FM_HOME_DIR/contracts
```
aakoshh, line 40: I found a different fix for this, [here](https://github.com/consensus-shipyard/fendermint/pull/319/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R15). 

The problem was that the way `IPC_ACTORS_OUT` was constructed found a multitude of `out` dirs since `bindings` were added. But the way it's [used](https://github.com/consensus-shipyard/fendermint/blob/8a30ea017a2071ab234faedcc42f8df25eb05db2/Makefile#L66) is clear: we intend to copy the contents of the out directory itself. 

So I'd say this fix isn't needed, it's not the right fix, because `docker/.artifacts/contracts` should contain the contracts directly.

fendermint/app/src/cmd/genesis.rs
```diff
@@ -189,9 +191,18 @@ where
...
     let genesis = read_genesis(genesis_file)?;
     let genesis_json = serde_json::to_value(&genesis)?;
+
+    // Fendermint chainIDs can only be 50 bytes long. To avoid
+    // arbitrary chain names from exceeding the limit, we compute
+    // the chainID as the 50 most significant bytes of the hex enconding
+    // of the SHA256 of the chain name
+    let mut hasher = sha2::Sha256::new();
+    hasher.update(genesis.chain_name);
+    let h_chain_id = hex::encode(hasher.finalize());
```
aakoshh, line 201: I thought we agreed that we will use the numeric chain ID itself, which we calculate [here](https://github.com/consensus-shipyard/fendermint/blob/b1dc6c1218331b39553c5990db2f6f8e6513f5ff/fendermint/vm/interpreter/src/fvm/genesis.rs#L84). I think it's preferable because that's what we'll show to the Ethereum API as well. No need to introduce yet another kind of ID for the chain.

infra/Makefile.toml
```diff
@@ -5,17 +5,34 @@ extend = [
...
+ETHAPI_HOST_PORT = { value = "8545", condition = { env_not_set = ["ETHAPI_HOST_PORT"] }}
+
+# IPC subnet related parameters
+# Use calibration as default value
+NODE_NAME = { value = "ipc-node", condition = { env_not_set = ["NODE_NAME"] } }
+PARENT_ENDPOINT = { value = "https://api.calibration.node.glif.io/rpc/v1", condition = { env_not_set = ["PARENT_ENDPOINT"] } }
+PARENT_GATEWAY = { value = "0x56948d2CFaa2EF355B8C08Ac925202db212146D1", condition = { env_not_set = ["PARENT_GATEWAY"] } }
+PARENT_REGISTRY = { value = "0x6A4884D2B6A597792dC68014D4B7C117cca5668e", condition = { env_not_set = ["PARENT_REGISTRY"] } }
+NETWORK_TYPE = { value = "test", condition = { env_not_set = ["NETWORK_TYPE"] } }
+VALIDATOR_PRIV_KEY = { value = "~/.ipc/priv.key", condition = { env_not_set = ["VALIDATOR_PRIV_KEY"] } }
```
aakoshh, line 32: There already exists a setting for the [validator_key](https://github.com/consensus-shipyard/fendermint/blob/b1dc6c1218331b39553c5990db2f6f8e6513f5ff/fendermint/app/settings/src/lib.rs#L110) and it is [relative to the home directory](https://github.com/consensus-shipyard/fendermint/blob/b1dc6c1218331b39553c5990db2f6f8e6513f5ff/fendermint/app/settings/src/lib.rs#L70), which [by default](https://github.com/consensus-shipyard/fendermint/blob/b1dc6c1218331b39553c5990db2f6f8e6513f5ff/fendermint/app/options/src/lib.rs#L73) is `~/.fendermint`. 

Why should we suddenly have a separate copy under an `~/.ipc` directory too?

aakoshh: Are all infra scripts using this different home direcory?

infra/scripts/cometbft.toml
```diff
@@ -34,7 +34,7 @@ dependencies = ["cometbft-pull", "docker-network-create"]
 script = """
 id=`docker exec ${CMT_CONTAINER_NAME} cometbft show-node-id`
 ip=`docker inspect -f '{{range.NetworkSettings.Networks}}{{.IPAddress}}{{end}}' ${CMT_CONTAINER_NAME}`
-echo $id@$ip:26656
+echo $id@$ip:${CMT_HOST_PORT}
```
aakoshh, line 37: Is this correct? Looks like we changed from the P2P to the RPC port.

infra/scripts/cometbft.toml
```diff
@@ -21,7 +21,7 @@ docker run \
   --name ${CMT_CONTAINER_NAME} \
   --user $(id -u) \
   --network ${NETWORK_NAME} \
-  --publish 26657:${CMT_HOST_PORT} \
+  --publish ${CMT_HOST_PORT}:26657 \
```
aakoshh, line 24: Oops, this is embarrassing :disappointed_relieved: 
Could I ask you to fix my mistakes [here](https://github.com/consensus-shipyard/fendermint/blob/2dc335ac1fc2c901f005017c8104c447bddb0f90/fendermint/testing/smoke-test/Makefile.toml#L132) and [here](https://github.com/consensus-shipyard/fendermint/blob/2dc335ac1fc2c901f005017c8104c447bddb0f90/fendermint/testing/smoke-test/Makefile.toml#L182) as well? :pray:

infra/scripts/cometbft.toml
aakoshh: I thought this was for something like the peer file, which enables containers to connect to each other without peer discovery, and 26656 was the port for that, which should work internally.

infra/scripts/subnet.toml
```diff
@@ -0,0 +1,57 @@
...
+    "subnet-genesis",
+    "subnet-fm-prepare",
+]
+
+
+[tasks.subnet-genesis]
+cwd = "./target/release"
+script = """
+
+./fendermint --network=${NETWORK_TYPE} genesis --genesis-file ${GENESIS_FILE} ipc from-parent --subnet-id ${CHAIN_NAME} -p ${PARENT_ENDPOINT}  --parent-gateway ${PARENT_GATEWAY}  --parent-registry ${PARENT_REGISTRY} --base-fee ${BASE_FEE} --power-scale ${POWER_SCALE}
```
aakoshh, line 45: Do these scripts run on the host, or inside a container?

docs/ipc.md
```diff
@@ -0,0 +1,55 @@
+# IPC
+
+This documentation will guide you through the different utils provided in Fendermint for the deployment of Fendermint-based IPC subnets. This docs are only focused on the infrastructure deployment, for an end-to-end walk through of spawning IPC subnets refer to the [IPC quickstart](https://github.com/consensus-shipyard/ipc/blob/main/docs/quickstart-calibration.md).
+
+## Deploy subnet bootstrap
+In order not to expose directly the network address information from validators, subnets leverage the use of bootstraps (or `seeds` in CometBFT parlance), for new nodes to discover peers in the network and connect to the subnet's validators. To run a bootstrap node you can run the following command from the root of the repo:
```
aakoshh, line 6:
```suggestion
In order not to expose directly the network address information from validators, subnets leverage the use of bootstrap nodes (or `seeds` in CometBFT parlance), for new nodes to discover peers in the network and connect to the subnet's validators. To run a bootstrap node you can run the following command from the root of the repo:
```

docs/ipc.md
```diff
@@ -0,0 +1,55 @@
...
+```bash
+cargo make --makefile infra/Makefile.toml bootstrap
```
aakoshh, line 8: The docs could mention if this will run on the host, or inside Docker, any prerequisites..

docs/ipc.md
```diff
@@ -0,0 +1,55 @@
...
+[cargo-make] INFO - Build Done in 13.38 seconds.
+```
+
+If at any time you need to query the endpoint of your bootstrap, you can run:
+```bash
+cargo make --makefile infra/Makefile.toml bootstrap-id
+```
+
+`cargo-make bootstrap` supports the following environment variables to customize the deployment:
+- `CMT_HOST_PORT` (optional): Specifies the listening port in the localhost for COMETBFT.
```
aakoshh, line 24:
```suggestion
- `CMT_HOST_PORT` (optional): Specifies the listening port in the localhost for CometBFT.
```

aakoshh: Previously I thought this was the RPC port, but now it looks like you wanted the P2P port. Perhaps would be useful to map both so the user can access the API to hit the info URLs.

infra/scripts/cometbft.toml
aakoshh: Isn't this IP address only available inside the container?

docs/ipc.md
```diff
@@ -0,0 +1,55 @@
...
+## Deploy subnet bootstrap
+In order not to expose directly the network address information from validators, subnets leverage the use of bootstraps (or `seeds` in CometBFT parlance), for new nodes to discover peers in the network and connect to the subnet's validators. To run a bootstrap node you can run the following command from the root of the repo:
+```bash
+cargo make --makefile infra/Makefile.toml bootstrap
+```
+You'll see that by the end of the output, this command should output the network address of your bootstrap. You can use this endpoint to include this bootstrap node as a seed in the `seeds` configuration of CometBFT.
+```console
+[cargo-make] INFO - Running Task: cometbft-wait
+[cargo-make] INFO - Running Task: cometbft-node-id
+2b23b8298dff7711819172252f9df3c84531b1d9@172.26.0.2:26650
```
aakoshh, line 14: This IP doesn't look like something the user can actually share with others on the network. By the looks of it it's the IP of the container inside the docker network, not the one visible externally over the internet.

aakoshh: Of course the docs doesn't explicitly say sharing it with the world was the intention. It's not exactly clear whether these infra scripts are for local testing or to connect to a real network.

docs/ipc.md
```diff
@@ -0,0 +1,55 @@
...
+
+
+## Deploy child subnet validator
+Once a child subnet has been bootstrapped in its parent, its subnet actor has been deployed, and has fulfilled its minimum requirements in terms of validators and minimum collateral, validators in the subnet can deploy their infrastructure to spawn the child subnet.
+
+In order to spawn a validator node in a child subnet, you need to run:
+```bash
+cargo make --makefile infra/Makefile.toml -e VALIDATOR_PRIV_KEY=<VALIDDATOR_PRIV_KEY> -e CHAIN_NAME=<SUBNET_ID> -e CMT_HOST_PORT=<COMETBFT_PORT> -e COMMA_SEPARATED_BOOTSTRAPS=<BOOTSTRAP_NODE1>,<BOOTSTRAP_NODE2> -e ETHAPI_HOST_PORT=<ETH_RPC_PORT> child-validator
+```
+This command will run the infrastructure for a Fendermint validator in the child subnet. It will generate the genesis of the subnet from the information in its parent, and will run the validator's infrastructure with the specific configuration passed in the command.
```
aakoshh, line 39: How does it know where the parent is?

docs/ipc.md
```diff
@@ -0,0 +1,55 @@
...
+
+`cargo-make child-validator` supports the following environment variables to customize the deployment:
+- `CMT_HOST_PORT` (optional): Specifies the listening port in the localhost for COMETBFT.
```
aakoshh, line 42:
```suggestion
- `CMT_HOST_PORT` (optional): Specifies the listening port in the localhost for CometBFT.
```

docs/ipc.md
```diff
@@ -0,0 +1,55 @@
...
+- `ETHAPI_HOST_PORT` (optional): Specifies the listening port in the localhost for the ETH RPC of the node.
+- `NODE_NAME` (optional): Name for the node deployment. Along with `CMT_HOST_PORT` and `ETHAPI_HOST_PORT`, these variables come really handy for the deployment of several validator nodes over the same system.
+- `VALIDATOR_PRIV_KEY`: Path of the private key for your validator (it should be the corresponding one used to join the subnet in the parent).
+- `CHAIN_NAME`: SubnetID for the child subnet.
```
aakoshh, line 46: Maybe it should actually ask for `SUBNET_ID` instead. I can imagine this business with the chain name will get confusing, given that we immediately hash it and use it as a chain ID, but we want them to pass subnet ID. Why not just call a spade a spade?

docs/ipc.md
```diff
@@ -0,0 +1,55 @@
...
+- `COMMA_SEPARATED_BOOTSTRAPS`: Comma separated list of bootstraps (or seeds in CometBFT parlance).
```
aakoshh, line 47: Perhaps including `COMMA_SEPARATED_` in the name is a bit pedantic, and simply `BOOTSTRAPS` would do, with the docs mentioning the format for including multiple?

docs/ipc.md
```diff
@@ -0,0 +1,55 @@
...
+- `PARENT_ENDPOINT`: Public endpoint that the validator should use to connect to the parent.
```
aakoshh, line 48: Perhaps an example in case of Lotus would help.

docs/ipc.md
aakoshh: I can't actually find this among the [env vars](https://github.com/consensus-shipyard/fendermint/blob/2dc335ac1fc2c901f005017c8104c447bddb0f90/infra/Makefile.toml#L15).

infra/scripts/cometbft.toml
adlrocha: `26656` is the internal port, but if we want to run more than one CometBFT instance locally (which will definitely be the case if you are running a validator and a bootstrap), we need to use different ports for each of them. What we show here is the external port being exposed in the host to connect with this CometBFT instance. That way external peers have public visibility of this node, thus the `CMT_HOST_PORT`

infra/scripts/cometbft.toml
adlrocha: I actually didn't want to touch it in case I was making a mistake myself and I was waiting for your review to confirm haha I have to admit that this bug was driving me crazy at first, but fixed :)

infra/scripts/subnet.toml
adlrocha: They run on the host, I followed what @dnkolegov was doing in the rest of his infra scripts, but I see that Raul already opened a PR to fix this: https://github.com/consensus-shipyard/fendermint/pull/329

I will let Raul fix that one and I will propagate the fix here.

docs/ipc.md
adlrocha: fixed

docs/ipc.md
adlrocha: fixed.

docs/ipc.md
adlrocha: Oh, I see. I thought I was exposing the P2P port here, but you are right that we may also want to expose the API. Can you remind me the default port for the API?

infra/scripts/cometbft.toml
aakoshh: Okay, I was confused because in my smoke test script `CMT_HOST_PORT` is used to expose the RPC connection. Perhaps we can have a different name for this one, like `CMT_P2P_HOST_PORT` or `CMT_HOST_P2P_PORT`, and either keep this for RPC or add an variant for that with RPC in the name. 

The docs don't say, but if you intend these CometBFT nodes to be deployed on the Internet then you'd probably want the RPC node exposed too, otherwise nobody can send a transaction, at least not to this node.

docs/ipc.md
adlrocha: fixed.

docker/local.Dockerfile
adlrocha: Gotcha. I included your fix and it works. Thanks!

docs/ipc.md
adlrocha: We need to pass the `PARENT_ENDPOINT` env variable. I was testing using calibration (which is used as the default endpoint), and that is why I missed it. I will add it.

docs/ipc.md
adlrocha: I inherited it from Denis scripts and didn't want to have to debug the rename as I was developing. But that makes sense. Fixed.

docs/ipc.md
adlrocha: Yes, this is in purpose, so that it is the empty string if not explicitly. I can add them to the list though with a default value of an empty string if you think it is clearer like that.

fendermint/app/src/cmd/genesis.rs
adlrocha: I think I misunderstood you, then, but your suggestion makes more sense. Nevermind, already fixed.

infra/Makefile.toml
adlrocha: So the goal of this variable is to pick up the export of a private key from the `ipc-cli` (so initially it can be persisted anywhere and not necessarily inside the `fendermint` path or with a Fendermint format). [The scripts](https://github.com/consensus-shipyard/fendermint/pull/320/files#diff-fb7d01e502bec77131b22c42a51ca3fa92060ae09be7538f10bf29584a6f746eR55-R56) will pick up this secret key, convert it into a Fendermint private key, and put it into the right location. 

This is why I decided to use a different env variable for this, as this is an external key that is picked up and converted into a Fendermint format to use it for the validator.

infra/scripts/cometbft.toml
adlrocha: Fixed.

docs/ipc.md
```diff
@@ -0,0 +1,78 @@
...
+```bash
+cargo make --makefile infra/Makefile.toml \
+    -e VALIDATOR_PRIV_KEY=<VALIDATOR_PRIV_KEY> \
+    -e SUBNET_ID=<SUBNET_ID> \
+    -e CMT_P2P_HOST_PORT=<COMETBFT_P2P_PORT> \
+    -e CMT_RPC_HOST_PORT=<COMETBFT_RPC_PORT> \
+    -e ETHAPI_HOST_PORT=<ETH_RPC_PORT> \
+    -e BOOTSTRAPS=<BOOTSTRAP_ENDPOINT>
+    -e PARENT_REGISTRY=<PARENT_REGISTRY_CONTRACT_ADDR> \
+    -e PARENT_GATEAY=<GATEWAY__REGISTRY_CONTRACT_ADDR> \
```
aakoshh, line 58:
```suggestion
    -e PARENT_GATEWAY=<PARENT_GATEWAY_CONTRACT_ADDR> \
```

docs/ipc.md
adlrocha: > When I last asked @dnkolegov whether this IP was accessible at all even from the host, he said no, it's internal to Docker, so this sentence seems misleading:

hmm, I wasn't aware of this. How can we expose the bootstrap to the outside world? @dnkolegov, do you mind chiming in? 

> Why don't you print the host IP instead, e.g. the output of curl ifconfig.me ?

Maybe we should just output a placeholder and let the user choose the relevant IP? I though about `ifconfig.me` but what if this is being deployed over an AWS instance with all of their networking madness? But you are definitely right, and giving the internal IP for Docker is the most confusing option.

infra/Makefile.toml
```diff
@@ -5,18 +5,36 @@ extend = [
...
+  { path = "scripts/subnet.toml" }
 ]
 
 [config]
 default_to_workspace = false
 
 [env]
-CHAIN_NAME = { value = "root", condition = { env_not_set = ["CHAIN_NAME"] }}
+# General network-specific parameters
+SUBNET_ID = { value = "root", condition = { env_not_set = ["SUBNET_ID"] }}
```
aakoshh, line 17:
```suggestion
SUBNET_ID = { value = "/r0", condition = { env_not_set = ["SUBNET_ID"] }}
```

That's the default we use [here](https://github.com/consensus-shipyard/fendermint/blob/66ca42d08c1708c99ea4d0444d9d2ac24a0516d8/fendermint/app/config/default.toml#L97). My understanding is all subnet IDs start with `/r` and 0 means it's default, unspecified.

infra/Makefile.toml
```diff
@@ -5,18 +5,36 @@ extend = [
...
+NODE_NAME = { value = "ipc-node", condition = { env_not_set = ["NODE_NAME"] } }
+PARENT_ENDPOINT = { value = "https://api.calibration.node.glif.io/rpc/v1", condition = { env_not_set = ["PARENT_ENDPOINT"] } }
+PARENT_GATEWAY = { value = "0x56948d2CFaa2EF355B8C08Ac925202db212146D1", condition = { env_not_set = ["PARENT_GATEWAY"] } }
+PARENT_REGISTRY = { value = "0x6A4884D2B6A597792dC68014D4B7C117cca5668e", condition = { env_not_set = ["PARENT_REGISTRY"] } }
+NETWORK_TYPE = { value = "test", condition = { env_not_set = ["NETWORK_TYPE"] } }
+VALIDATOR_PRIV_KEY = { value = "~/.ipc/priv.key", condition = { env_not_set = ["VALIDATOR_PRIV_KEY"] } }
+TOPDOWN_CHAIN_HEAD_DELAY = { value = "10", condition = { env_not_set = ["TOPDOWN_CHAIN_HEAD_DELAY"] } }
+
+# Deployment-related
+BASE_DIR="${HOME}/.ipc/${SUBNET_ID}"
```
aakoshh, line 37: Just to confirm, the `SUBNET_ID` is something like `/r123/f410123...abc/f064`, and they will actually map to a nice nested directory structure, right?

infra/scripts/cometbft.toml
```diff
@@ -34,7 +35,7 @@ dependencies = ["cometbft-pull", "docker-network-create"]
 script = """
 id=`docker exec ${CMT_CONTAINER_NAME} cometbft show-node-id`
```
aakoshh, line 36: This would be a good spot for grabbing the host IP, although I am not exactly sure if that can be used inside docker.

docs/ipc.md
adlrocha: Fixed.

infra/scripts/node.toml
```diff
@@ -1,19 +1,36 @@
 ########################################################################################################################
 # Node helpers
 ########################################################################################################################
+
+[tasks.bootstrap]
+env = { "NETWORK_NAME"="${NETWORK_NAME}", "CMT_DIR" = "${BASE_DIR}/bootstrap/cometbft", "CMT_CONTAINER_NAME" = "cometbft-${NETWORK_NAME}-bootstrap", "CMT_P2P_HOST_PORT" = "26650"}
```
aakoshh, line 6: I might be misunderstanding cargo make, but [the docs](https://github.com/consensus-shipyard/fendermint/pull/320/files#diff-9265726f2e486a1f8a407d6041312142989b59d7d59960399fc4e7085e203125R35-R37) say we can define the bootstrap host port, but here it's hardcoded.

infra/Makefile.toml
adlrocha: Correct. Good catch, Denis was allowing any chain name and completely missed this one. Thanks!

infra/Makefile.toml
adlrocha: Correct, so if you have more than one subnet anchored to `/r123` they will be in `/r123/subnet1` and `/r123/subnet2` (I already tested this one just in case :) )

infra/scripts/node.toml
adlrocha: Removed, it may be confusing. Let's leave users choose their port instead of forcing a different one for them.

docs/ipc.md
aakoshh: I'd use `ifconfig.me` as a reflection of the host IP and mention it. I'm not sure what else it can be, but I'm fairly certain if someone deploys to AWS they will take more control over the process.

docs/ipc.md
aakoshh: I can't even find where it's used right now. But yeah, an empty string with a comment above it saying what its' format is supposed to be would make it easily discoverable for someone who take a look at the makefile.

docs/ipc.md
adlrocha: Good point, will update accordingly, thanks!

aakoshh (approved):
LGTM! It's great to see are taking it through its paces and pushing it through the finish line to the hungry masses!

docs/ipc.md
aakoshh: Cool, I see you changed the script. Maybe change the IP here as well to something operators with a keen eye would not recognise as a docker IP. They usually start with 172.x.x.x

PR #321, opened 2023-10-23
Fix topdown

cryptoAtwill:
Contain fixes and updates from integration testing. Currently the happy path from this branch is working. Key changes:
- When top down messages are queried, we need to query the next block hash for lotus. See sync for more comments.
- Mint to gateway when cross messages are executed.
- Moving as much code as possible to `GatewayCaller`.


fendermint/vm/interpreter/src/chain.rs
```diff
@@ -238,48 +246,31 @@ where
...
-                    let msg = self
-                        .gateway_caller
-                        .store_validator_changes_msg(validator_changes)?;
-                    let (state, ret) = self
-                        .inner
-                        .deliver(state, VerifiableMessage::NotVerify(msg))
+                    let validator_changes = provider
+                        .validator_changes_from(prev_height + 1, p.height as u64)
+                        .await?;
+                    let (state, _, _) = self
```
aakoshh, line 259: The ignored variables are bothering me a little. Is one of them the `ret`? We should not ignore errors here. For now we should stick to returning an overall error if any of these fail.

fendermint/vm/interpreter/src/chain.rs
```diff
@@ -304,6 +295,90 @@ where
...
+        provider: &TopDownFinalityProvider,
+        finality: IPCParentFinality,
+    ) -> anyhow::Result<CommitFinalityResult<DB>> {
+        let msg = self.gateway_caller.commit_parent_finality_msg(finality)?;
+        let (state, ret) = self
+            .inner
+            .deliver(state, VerifiableMessage::NotVerify(msg))
+            .await?;
+        if is_error(&ret) {
+            return Err(anyhow!("cannot commit parent finality"));
```
aakoshh, line 327: It would be nice to attach some info about the error.

fendermint/vm/interpreter/src/chain.rs
aakoshh: Right, I see that inside these methods they return errors. It's a bit confusing that everything is ignored here. 

Also, shouldn't we take at least the events from these and attach them to the final returned result?

fendermint/vm/interpreter/src/chain.rs
```diff
@@ -238,48 +246,31 @@ where
...
                         .await?;
-                    let msg = self.gateway_caller.apply_cross_messages_msg(messages)?;
-                    let (state, ret) = self
-                        .inner
-                        .deliver(state, VerifiableMessage::NotVerify(msg))
-                        .await?;
-                    if ret.is_err() {
-                        return Err(anyhow!("failed to apply cross messages"));
-                    }
+                    let (state, ret, _) = self.execute_topdown_msgs(state, messages).await?;
```
aakoshh, line 267:
```suggestion
                    let (state, ret, ()) = self.execute_topdown_msgs(state, messages).await?;
```
Makes it clear what the return value is.

fendermint/vm/interpreter/src/chain.rs
aakoshh, line 259:
```suggestion
                    let (state, _ret, ()) = self
```
To make it clear that `()` is returned, and if not it wouldn't compile, so it can't be ignored.

fendermint/vm/interpreter/src/chain.rs
```diff
@@ -420,22 +495,24 @@ fn relayed_bottom_up_ckpt_to_fvm(
...
+    gateway: &GatewayCaller<DB>,
+    bytes: RawBytes,
+    provider: &TopDownFinalityProvider,
+) -> anyhow::Result<(u64, Option<IPCParentFinality>)> {
+    let (has_committed, prev_finality) = gateway.decode_commit_parent_finality_return(bytes)?;
+    Ok(if !has_committed {
+        (provider.genesis_epoch()?, None)
+    } else {
+        (prev_finality.height, Some(prev_finality))
+    })
```
aakoshh, line 517: I think it would be good to move most of these into a `topdown` module, similar to the `checkpoint` module. They are so numerous and complicated, they completely dominate the `chain.rs` file by now.

fendermint/vm/interpreter/src/fvm/state/ipc.rs
```diff
@@ -202,6 +204,34 @@ impl<DB: Blockstore> GatewayCaller<DB> {
...
 
+    pub fn decode_commit_parent_finality_return(
+        &self,
+        bytes: RawBytes,
+    ) -> anyhow::Result<(bool, IPCParentFinality)> {
+        let return_data = bytes
+            .deserialize::<BytesDe>()
+            .context("failed to deserialize return data")?;
+
+        let function = router::GATEWAYROUTERFACET_ABI
```
aakoshh, line 215:
```suggestion
        let function = self.router.contract().abi()
```

fendermint/vm/interpreter/src/fvm/state/ipc.rs
```diff
@@ -202,6 +204,34 @@ impl<DB: Blockstore> GatewayCaller<DB> {
...
+            .functions
+            .get("commitParentFinality")
+            .ok_or_else(|| anyhow!("broken abi"))?
+            .get(0)
+            .ok_or_else(|| anyhow!("function not found, abi wrong?"))?;
```
aakoshh, line 220: Would be good to at least mention what it was looking for.

fendermint/vm/interpreter/src/fvm/state/ipc.rs
```diff
@@ -202,6 +204,34 @@ impl<DB: Blockstore> GatewayCaller<DB> {
...
+
+        if let Token::Tuple(mut tuple) = decode_function_data(function, return_data.0, false)? {
```
aakoshh, line 222: `decode_function_data` is generic. I'm sure we can decode into the expected type instead of dealing with popping `Token`s.

aakoshh, line 222:
```suggestion
        match decode_function_data::<(bool, router::ParentFinality)>(function, return_data.0, false)? {
```

aakoshh (commented):
Impressive! :clap: 

Do you handle temporary network failures when looking up stuff from the parent? I can't remember where the backoff was applied.

cryptoAtwill:
@aakoshh There is a exponential back off implemented [here](https://github.com/consensus-shipyard/fendermint/blob/0d584fe33fa02da088f3b2ed06fde3f8514d7f8a/fendermint/vm/topdown/src/finality.rs#L98).

Some key follow up changes:
- Added `topdown` module and shifted top down parent finality execution logic to `topdown` module.
- Added `call_with_return` method in `ContractCaller`. This way we can obtain the `FvmApplyRet` from calling the contracts.
- Added `ChainMessageApplyRet::Ipc(FvmApplyRet)`, so that we dont have to used `Signed` for unsigned IPC messages.

Currently for top down execution result, only the `topdown` cross message execution apply result is returned. This is because:
- Merging `ApplyRet` seems difficult and awkward to do.
- Will merge `commitParentFinality` and `storeValidatorChanges` into one method in the future, so we can do the ret merging there.
- Currently if we dont return the `ApplyRet` from `commitParentFanlity` and `storeValidatorChanges` is still acceptable for UX (I could be wrong), but good to have for debugging. So for this iteration, I think it is still ok.

fendermint/vm/interpreter/src/chain.rs
```diff
@@ -56,6 +58,8 @@ pub struct IllegalMessage;
 // For now this is the only option, later we can expand.
 pub enum ChainMessageApplyRet {
     Signed(SignedMessageApplyRes),
+    /// The IPC chain message execution result
+    Ipc(FvmApplyRet),
```
aakoshh, line 62: I agree this is cleaner!

fendermint/vm/interpreter/src/fvm/state/ipc.rs
```diff
@@ -189,54 +186,65 @@ impl<DB: Blockstore> GatewayCaller<DB> {
...
-        let calldata = call
-            .calldata()
-            .ok_or_else(|| anyhow!("no calldata for apply cross messages"))?;
-
-        encode_to_fvm_implicit(calldata.as_ref())
+    ) -> anyhow::Result<FvmApplyRet> {
+        let messages = cross_messages
+            .into_iter()
+            .map(router::CrossMsg::try_from)
+            .collect::<Result<Vec<_>, _>>()?;
```
aakoshh, line 243:
```suggestion
            .collect::<Result<Vec<_>, _>>()
            .context("failed to convert cross messages")?;
```

I hope this cannot be triggered by malicious users.

aakoshh (approved):
Great! I also couldn't really find a justification for calling the inner interpreter with the the `NotValidate` kind and then trying to decode the return value when the context has been lost. Doing more in the `GatewayCaller` is much simpler :+1:

PR #326, opened 2023-10-24
improve `localnet.md`

raulk:
- Details system prerequisites (e.g. Docker, Rust, etc.) and provides instructions for Linux (Ubuntu).
- Adds missing details for when users start from scratch (which is the most probable scenario).
- Add some explanations and clean up.

### DIFF
--- a/docs/localnet.md
+++ b/docs/localnet.md
@@ -11,7 +11,7 @@ This guide offers two flavours:
 
 On Linux (links and instructions for Ubuntu):
 
-- Install developer tools: `sudo apt install build-essential`.
+- Install developer tools: `sudo apt install build-essential pkg-config`.
 - Install Rust. See [instructions](https://www.rust-lang.org/tools/install).
 - Install cargo-make: `cargo install --force cargo-make`.
 - Install Docker. See [instructions](https://docs.docker.com/engine/install/ubuntu/).
--- a/docs/localnet.md
+++ b/docs/localnet.md
@@ -1,16 +1,33 @@
 # Local Testnets
 
-Prerequisites:
+Setting up a Fendermint local testnet is a way to get started quickly with IPC.
+
+This guide offers two flavours:
+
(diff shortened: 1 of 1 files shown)


docs/localnet.md
```diff
@@ -21,41 +39,48 @@ cargo make --makefile ./infra/Makefile.toml testnode-down
...
-- Run CometBFT in 4 containers
-- Run Eth API in 4 containers
\ No newline at end of file
+## What's happening behind the scenes
+
+> For a 4-node deployment.
+
+The deployment process performs the following steps:
+
+- Remove all Docker containers, files, networks, etc. from any previous deployments.
```
adlrocha, line 78: The scripts should be handling this for you, but definitely useful as a sanity-check

docs/localnet.md
```diff
@@ -1,16 +1,34 @@
...
+This guide offers two flavours:
+
+- A single node deployment: useful for developing smart contracts and testing the APIs.
+- A 4 node testnet: useful for testing consensus, checkpointing, and more.
+
+## Prerequisites
+
+On Linux (links and instructions for Ubuntu):
+
+- Install system packages: `sudo apt install build-essential clang cmake pkg-config libssl-dev protobuf-compiler`.
```
adlrocha, line 14: We should maybe add this to a `make install-deps` to make users lives easier. I will track it in an issue.

adlrocha (approved):
This was long overdue. Thank you @raulk.

raulk:
Merged this branch into #329. Closing in favour of that.

docs/localnet.md
raulk: Can do, but this step is OS and distribution-dependent.

- Would need to check compatibility and abort if an unknown/unsupported Linux distro is being used.
- Would be great to make it support macOS.

Having said that, I'm migrating these scripts to use docker images, so this step will soon become obsolete/unnecessary.

PR #330, opened 2023-10-25
FM-316: Top down mint increase circ supply

### LINKED ISSUE #316: Adjust circulating supply during top-down message execution

aakoshh:
Closes #316 

The PR changes the top-down message execution so that it increases the circulating supply with the tokens minted and given to the gateway.

### DIFF
--- a/fendermint/vm/interpreter/src/fvm/state/exec.rs
+++ b/fendermint/vm/interpreter/src/fvm/state/exec.rs
@@ -80,6 +80,7 @@ where
     /// execution interpreter without having to add yet another piece to track at the app level.
     block_hash: Option<BlockHash>,
 
+    /// State of parameters that are outside the control of the FVM but can change and need to be persisted.
     params: FvmUpdatableParams,
 
     /// Indicate whether the parameters have been updated.
@@ -223,6 +224,23 @@ where
         }
 
         Ok(emitters)
+    }
+
+    /// Update the circulating supply, effective from the next block.
+    pub fn update_circ_supply<F>(&mut self, f: F)
+    where
+        F: FnOnce(&mut TokenAmount),
+    {
+        self.update_params(|p| f(&mut p.circ_supply))
+    }
+
+    /// Update the parameters and mark them as dirty.
+    fn update_params<F>(&mut self, f: F)
+    where
+        F: FnOnce(&mut FvmUpdatableParams),
+    {
+        f(&mut self.params);
+        self.params_dirty = true;
     }
 }
 
--- a/fendermint/vm/interpreter/src/fvm/state/exec.rs
+++ b/fendermint/vm/interpreter/src/fvm/state/exec.rs
@@ -55,7 +55,7 @@ pub struct FvmStateParams {
 /// TODO: `base_fee` should surely be here.
 #[derive(Debug)]
 pub struct FvmUpdatableParams {
-    /// The circulating supply changes if IPC is elabled and
+    /// The circulating supply changes if IPC is enabled and
     /// funds/releases are carried out with the parent.
     pub circ_supply: TokenAmount,
     /// Conversion between collateral and voting power.
--- a/fendermint/vm/interpreter/src/fvm/state/exec.rs
+++ b/fendermint/vm/interpreter/src/fvm/state/exec.rs
@@ -48,6 +48,22 @@ pub struct FvmStateParams {
     pub power_scale: PowerScale,
 }
 
+/// Parts of the state which can be updated by message execution, apart from the actor state.
+///
+/// This is just a technical thing to help us not forget about saving something.
+///
+/// TODO: `base_fee` should surely be here.
+#[derive(Debug)]
+pub struct FvmUpdatableParams {
+    /// The circulating supply changes if IPC is elabled and
+    /// funds/releases are carried out with the parent.
+    pub circ_supply: TokenAmount,
+    /// Conversion between collateral and voting power.
+    /// Doesn't change at the moment but in theory it could,
+    /// and it doesn't have a place within the FVM.
+    pub power_scale: PowerScale,
+}
+
 pub type MachineBlockstore<DB> = <DefaultMachine<DB, FendermintExterns> as Machine>::Blockstore;
 
 /// A state we create for the execution of all the messages in a block.
@@ -64,8 +80,10 @@ where
     /// execution interpreter without having to add yet another piece to track at the app level.
     block_hash: Option<BlockHash>,
 
-    /// Conversion between collateral and voting power.
-    power_scale: PowerScale,
+    params: FvmUpdatableParams,
+
+    /// Indicate whether the parameters have been updated.
+    params_dirty: bool,
 }
 
 impl<DB> FvmExecState<DB>
@@ -90,7 +108,7 @@ where
         // * base_fee; by default it's zero
         let mut mc = nc.for_epoch(block_height, params.timestamp.0, params.state_root);
         mc.set_base_fee(params.base_fee);
-        mc.set_circulating_supply(params.circ_supply);
+        mc.set_circulating_supply(params.circ_supply.clone());
 
         // Creating a new machine every time is prohibitively slow.
         // let ec = EngineConfig::from(&nc);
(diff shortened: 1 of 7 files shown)


fendermint/vm/interpreter/src/fvm/state/exec.rs
```diff
@@ -48,6 +48,22 @@ pub struct FvmStateParams {
...
 }
 
+/// Parts of the state which can be updated by message execution, apart from the actor state.
+///
+/// This is just a technical thing to help us not forget about saving something.
+///
+/// TODO: `base_fee` should surely be here.
+#[derive(Debug)]
+pub struct FvmUpdatableParams {
+    /// The circulating supply changes if IPC is elabled and
```
cryptoAtwill, line 58:
```suggestion
    /// The circulating supply changes if IPC is enabled and
```

fendermint/vm/interpreter/src/fvm/state/exec.rs
```diff
@@ -64,8 +80,11 @@ where
     /// execution interpreter without having to add yet another piece to track at the app level.
     block_hash: Option<BlockHash>,
 
-    /// Conversion between collateral and voting power.
-    power_scale: PowerScale,
+    /// State of parameters that are outside the control of the FVM but can change and need to be persisted.
+    params: FvmUpdatableParams,
+
+    /// Indicate whether the parameters have been updated.
+    params_dirty: bool,
```
cryptoAtwill, line 87: Just curious, I see this parameter updated but not read, is it used in other PRs?

cryptoAtwill (commented):
Probably this was discussed somewhere before, but why don't we do it in gateway/subnetActor contract but choose to track it in `fendermint`?

adlrocha:
> Probably this was discussed somewhere before, but why don't we do it in gateway/subnetActor contract but choose to track it in fendermint?

This is tracked by the parent every time funds move down, but the FVM also needs to track of its circulating supply. With this we have an accurate view of what is the circulating supply of a subnet. Even more, this is the accurate value for the circulating supply. Even if the parent says that the circulating supply of the subnet is 10FIL, that is just what was frozen to be injected in the subnet, but the real circulating supply in the subnet may be less because some top-down messages haven't been executed yet, or because funds have been burnt. 

Even more, this opens the door to something we have discussed for a while, the ability to reward with native tokens in the subnet, producing some inflation in the subnet (producing a dynamic exchange rate between the child subnet and the parent's token). This is completely out of scope in the short-term, but worth noting.

cryptoAtwill:
So the circulating supply will be tracked in both the parent and child. It is possible that these two values will be out of sync and eventually be consistent?

fendermint/vm/interpreter/src/fvm/state/exec.rs
```diff
@@ -48,6 +48,22 @@ pub struct FvmStateParams {
     pub power_scale: PowerScale,
 }
 
+/// Parts of the state which can be updated by message execution, apart from the actor state.
+///
+/// This is just a technical thing to help us not forget about saving something.
+///
+/// TODO: `base_fee` should surely be here.
```
adlrocha, line 55: Nice one, I like this approach! Yes, I was actually going to ask you why the `base_fee` wasn't included here. 

I would include it for now even if it doesn't currently change as we may want to expose a callback to update it for users in the future. But maybe you have a reason not to include it yet, so happy either way.

fendermint/vm/interpreter/src/fvm/topdown.rs
```diff
@@ -45,7 +48,14 @@ where
...
+
+    gateway_caller
+        .mint_to_gateway(state, total_value.clone())
+        .context("failed to mint to gateway")?;
+
+    state.update_circ_supply(|circ_supply| {
+        *circ_supply += total_value;
+    });
 
     gateway_caller.apply_cross_messages(state, messages)
```
adlrocha, line 60: I think we already discussed this when we were reviewing top-down, and we said we would defer it to the future, but what is your take for when executing the message fails?

adlrocha (approved):
Nice, we are starting to get really close to a beautiful end-to-end for the system.

adlrocha:
> So the circulating supply will be tracked in both the parent and child. It is possible that these two values will be out of sync and eventually be consistent?

Generally the should end up being consistent, but there may be cases where they are not (like funds being burnt directly in the child without being released to the parent).

* The circulating supply of the parent is what has been tranferred down in top-down messages (and the maximum amount that can be released from the subnet)
* The one in the child is the real circulating supply that the subnet has (depending on the use case this one may increase through internal processes in the subnet, like reward validators from minted token from thin air)

fendermint/vm/interpreter/src/fvm/state/exec.rs
aakoshh: You are right, it's a bit obscure. The only reason I added it was to use it in [genesis](https://github.com/consensus-shipyard/fendermint/pull/330/files#diff-8f0901d357789002b95cef8fd68d18055c0ef0b06a3873ee9cea6396f177ea3cR155) where the `commit`, like the one here used to, returns just a `CID` but it's a unified one coming either from the `StateTree` or the `FvmExecState`. I didn't want to change that `commit` too, but I wanted to have a sanity check that there hasn't been some change that I am discarding, in which case this design would have to change too, e.g. by moving the fields from `FvmGenesisOutput` to the commit return value.

It's a bit convoluted, but I haven't found a nice way to do things. I tried with maintaining a copy of `FvmStateParams`, but not all those fields can change during a run, and for the ones that can be retrieved from the executor, there is always that ambiguity of which one to return. Now there is only one, because the `circ_supply` in the updatable part is write-only.

fendermint/vm/interpreter/src/fvm/state/exec.rs
aakoshh: I didn't include it purely because there is no user story for how to update it. I created an issue, but haven't checked what's happening in Lotus or Forest. Maybe it's based on the last N blocks or something? Or maybe it's based on some voting, I don't know. I'd rather have a TODO stating it's not done, then code which suggests it might be done, which sends people looking all over to code just to confirm that nothing is happening. 

But the history of it is already maintained, so as soon as we add it here, the compiler will tell us to update the value in `app.rs` and all will be well.

aakoshh:
> Probably this was discussed somewhere before, but why don't we do it in gateway/subnetActor contract but choose to track it in fendermint?

The reason for tracking it in Fendermint is that the circulating supply is basically a constructor parameter for an FVM instance, so at the very least it has to be accessible to Fendermint. Technically it could live in the state of an actor, like the `Init` actor is used to look up actor ID to address associations, I don't have to implement that in the host. It could be updated there and I could read it when we do the new block. But it would have to be an actor different from any of the IPC ones, as IPC is optional, while the circulating supply is not.

fendermint/vm/interpreter/src/fvm/state/exec.rs
adlrocha: I would keep a static `base_fee` and not use what Lotus is doing, I would rather have a way for users to decide the `base_fee` function instead of giving them a really complicated one that they don't understand. So yes, let's leave it for the future. Thanks!

fendermint/vm/interpreter/src/fvm/topdown.rs
aakoshh: I had to re-read my comments:
1. [For crashing](https://github.com/consensus-shipyard/fendermint/pull/310#discussion_r1352791877)
2. [Against crashing](https://github.com/consensus-shipyard/fendermint/pull/310#discussion_r1352803799)

To reiterate the main points:
* Individual messages in the batch should be allowed to fail without failing the whole batch
* Failing the whole batch and moving on is unacceptable, especially if we can't roll back the validator changes and the parent height finality
* But, the whole match might fail if it goes over the block gas limit
* Therefore we should:
  * do all 3 parts as an atomic call to Solidity
  * let the whole thing fail as a transaction and be retried by another relayer
  * not execute immediately (and hit the block gas limit) but enqueue for execution and do it when we have room
  * then we can treat individual messages with their own gas limit

aakoshh:
> funds being burnt directly in the child without being released to the parent

If you send tokens to a burner address, the host will have no knowledge of this, and the circulating supply will not reflect the permanent locking of them. There is a [burntfunds](https://github.com/consensus-shipyard/fendermint/blob/main/fendermint/vm/actor_interface/src/burntfunds.rs) actor which we could use to adjust the circulating supply down, but people aren't obliged to use it.

PR #331, opened 2023-10-25
FM-287: Bottom up burn decrease circ supply

### LINKED ISSUE #287: Adjust circulating supply after releasing to parent
The `circ_supply` fields it is initialized [here](https://github.com/consensus-shipyard/fendermint/blob/main/fendermint/app/src/app.rs#L407) in the App, but it's currently [not updated](https://github.com/consensus-shipyard/fendermint/issues/19) during message execution. It is the number of tokens in the current subnet.

When we create a bottom-up checkpoint, if there is a release message inside it, it should affect this value in Fendermint. For this it looks like should not only hash the messages in https://github.com/consensus-shipyard/fendermint/issues/281 but inspect how much value they release.

The circulating supply can be adjusted on the [MachineContext](https://github.com/filecoin-project/ref-fvm/blob/fvm%40v3.8.0/fvm/src/machine/mod.rs#L253), however there is no way to get a mutable reference to it through the [DefaultMachine](https://github.com/filecoin-project/ref-fvm/blob/fvm%40v3.8.0/fvm/src/machine/default.rs#L148), so any adjustment has to be delayed until the end.

aakoshh:
Closes #287 

Changes the checkpoint creation process to include the decrease of circulating supply by the values that are included in the bottom-up messages. 

Introduced some functions to help keep things consistent:
* An `AbiHash` trait is implemented for anything we want to hash in Solidity and match that hash in Rust, to take out the guesswork from tuple or no-tuple. Seemingly unrelated but since I had to retrieve all the cross messages, whereas previously I got the hash directly from the gateway caller, I didn't want to use the naked `abi_hash` function from the interpreter. 
* `tokens_to_mint` and `tokens_to_burn` to sum up the value in cross messages.

### Questions
* We ignored the `fee` during minting, so I assumed it meant that the fees will be collected in the target network. => _Changed both the minting and the release to add both fees and values_.
* Do we have to modify the balance of the gateway to burn these tokens, or did the contracts already burn the appropriate amount? => _The contract already burned, so we just modify the circulating supply to match it._

### Caveat

The circulating supply will reflect the changes from IPC, but it will not show what is truly the unburnt token amount in the subnet, because people can use any inaccessible address for burning, not just the `BURNTFUNDS_ACTOR`.

### DIFF
--- a/fendermint/vm/actor_interface/src/ipc.rs
+++ b/fendermint/vm/actor_interface/src/ipc.rs
@@ -6,7 +6,9 @@
 // Solidity contracts during genesis.
 
 use anyhow::Context;
+use ethers::core::abi::Tokenize;
 use ethers::core::types as et;
+use ethers::core::utils::keccak256;
 use fendermint_vm_genesis::{Power, Validator};
 use fvm_shared::address::Error as AddressError;
 use fvm_shared::address::Payload;
@@ -195,6 +197,15 @@ pub fn subnet_id_to_eth(subnet_id: &SubnetID) -> Result<(u64, Vec<et::Address>),
         route.push(et::H160::from(addr.0))
     }
     Ok((subnet_id.root_id(), route))
+}
+
+/// Hash some value in the same way we'd hash it in Solidity.
+///
+/// Be careful that if we have to hash a single struct,
+/// Solidity's `abi.encode` function will treat it as a tuple,
+/// so it has to be passed as a tuple in Rust. Vectors are fine.
+pub fn abi_hash<T: Tokenize>(value: T) -> [u8; 32] {
+    keccak256(ethers::abi::encode(&value.into_tokens()))
 }
 
 pub mod gateway {
--- a/fendermint/vm/actor_interface/src/ipc.rs
+++ b/fendermint/vm/actor_interface/src/ipc.rs
@@ -208,6 +208,38 @@ pub fn abi_hash<T: Tokenize>(value: T) -> [u8; 32] {
     keccak256(ethers::abi::encode(&value.into_tokens()))
 }
 
+/// Types where we need to match the way we sign them in Solidity and Rust.
+pub trait AbiHash {
+    /// Hash the item the way we would in Solidity.
+    fn abi_hash(self) -> [u8; 32];
+}
+
+macro_rules! abi_hash {
+    (struct $name:ty) => {
+        // Structs have to be hashed as a tuple.
+        impl AbiHash for $name {
+            fn abi_hash(self) -> [u8; 32] {
+                abi_hash((self,))
+            }
+        }
+    };
+
+    (Vec < $name:ty >) => {
+        // Vectors can be hashed as-is
+        impl AbiHash for Vec<$name> {
+            fn abi_hash(self) -> [u8; 32] {
+                abi_hash(self)
+            }
+        }
+    };
+}
+
+abi_hash!(struct ipc_actors_abis::gateway_router_facet::BottomUpCheckpoint);
+abi_hash!(struct ipc_actors_abis::subnet_actor_manager_facet::BottomUpCheckpoint);
+abi_hash!(Vec<ipc_actors_abis::gateway_getter_facet::CrossMsg>);
+abi_hash!(Vec<ipc_actors_abis::subnet_actor_manager_facet::CrossMsg>);
+abi_hash!(Vec<ipc_actors_abis::subnet_actor_getter_facet::CrossMsg>);
+
 pub mod gateway {
     use super::subnet_id_to_eth;
     use ethers::contract::{EthAbiCodec, EthAbiType};
--- a/fendermint/vm/interpreter/src/fvm/checkpoint.rs
+++ b/fendermint/vm/interpreter/src/fvm/checkpoint.rs
@@ -5,6 +5,7 @@ use std::collections::HashMap;
 
 use anyhow::{anyhow, Context};
 use fendermint_crypto::PublicKey;
+use fendermint_vm_actor_interface::ipc::AbiHash;
 use fendermint_vm_genesis::Collateral;
 use fendermint_vm_genesis::PowerScale;
 use fendermint_vm_message::conv::from_eth;
@@ -74,9 +75,13 @@ where
                 .context("failed to apply validator changes")?;
 
             // Retrieve the bottom-up messages so we can put their hash into the checkpoint.
-            let cross_messages_hash = gateway
-                .bottom_up_msgs_hash(state, height.value())
-                .context("failed to retrieve bottom-up messages hash")?;
+            let cross_msgs = gateway
+                .bottom_up_msgs(state, height.value())
+                .context("failed to retrieve bottom-up messages")?;
+
+            // TODO: circ_supply
+
+            let cross_messages_hash = cross_msgs.abi_hash();
 
             // Construct checkpoint.
             let checkpoint = BottomUpCheckpoint {
--- a/fendermint/vm/interpreter/src/fvm/checkpoint.rs
+++ b/fendermint/vm/interpreter/src/fvm/checkpoint.rs
@@ -22,6 +22,7 @@ use fendermint_vm_genesis::{Power, Validator, ValidatorKey};
 use ipc_actors_abis::gateway_getter_facet as getter;
 use ipc_actors_abis::gateway_router_facet as router;
 
+use super::state::ipc::tokens_to_burn;
 use super::{
     broadcast::Broadcaster,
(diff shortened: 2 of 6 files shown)


adlrocha:
Let me first answer your questions here before the review: 
> We ignored the fee during minting, so I assumed it meant that the fees will be collected in the target network.

Correct. We just propagate them inside the message as-is, and the execution in the parent will worry about handling them and distributing them. That being said, every amount of tokens leaving the subnet must be burnt reducing the circulating supply, so the circulating supply so be decreased in `sum(msg.value + msg.fee)`. 

> Do we have to modify the balance of the gateway to burn these tokens, or did the contracts already burn the appropriate amount?

This is already handled by the contract. But thank you for pointing it out. I was checking the code to point you to the places where we were burning the funds, and for some reason someone decided to remove the line that were burning funds :) This is the PR with the fix: https://github.com/consensus-shipyard/ipc-solidity-actors/pull/263

fendermint/vm/actor_interface/src/ipc.rs
```diff
@@ -197,6 +199,47 @@ pub fn subnet_id_to_eth(subnet_id: &SubnetID) -> Result<(u64, Vec<et::Address>),
     Ok((subnet_id.root_id(), route))
 }
 
+/// Hash some value in the same way we'd hash it in Solidity.
+///
+/// Be careful that if we have to hash a single struct,
+/// Solidity's `abi.encode` function will treat it as a tuple,
+/// so it has to be passed as a tuple in Rust. Vectors are fine.
+pub fn abi_hash<T: Tokenize>(value: T) -> [u8; 32] {
```
adlrocha, line 207: @cryptoAtwill, we may need to use this one in the relayer when we check the hash of the messages for the checkpoint. It may make our lives significantly easier :)

fendermint/vm/interpreter/src/fvm/checkpoint.rs
```diff
@@ -74,9 +76,19 @@ where
...
-                .bottom_up_msgs_hash(state, height.value())
-                .context("failed to retrieve bottom-up messages hash")?;
+            let cross_msgs = gateway
+                .bottom_up_msgs(state, height.value())
+                .context("failed to retrieve bottom-up messages")?;
+
+            // Sum up the value leaving the subnet as part of the bottom-up messages.
+            let burnt_tokens = tokens_to_burn(&cross_msgs);
+
+            // TODO: Should we actually take this from the gateway balance? Or did it already get burnt?
```
adlrocha, line 86: It should already be burnt and discussed here: https://github.com/consensus-shipyard/ipc-solidity-actors/pull/263

fendermint/vm/interpreter/src/fvm/state/ipc.rs
```diff
@@ -270,10 +258,26 @@ impl<DB: Blockstore> GatewayCaller<DB> {
...
-/// Be careful that if we have to hash a single struct, Solidity's `abi.encode`
-/// function will treat it as a tuple.
-pub fn abi_hash<T: Tokenize>(value: T) -> [u8; 32] {
-    keccak256(ethers::abi::encode(&value.into_tokens()))
+/// Fees are ignored, which suggests that they will be paid similarly to gas,
+/// from the balance of the sender already present in the currents subnet.
+pub fn tokens_to_mint(msgs: &[ipc_sdk::cross::CrossMsg]) -> TokenAmount {
+    msgs.iter()
+        .fold(TokenAmount::from_atto(0), |mut total, msg| {
+            total += &msg.msg.value;
```
adlrocha, line 268: We should also add the fee here. It will generally be zero in the current state, but it should become part of the circulating supply if there are fees (even if we haven't decided yet how to distribute them for top-down messages). This is assuming that you are looking at the `CrossMsg` type, which seems to be the case.

fendermint/vm/interpreter/src/fvm/state/ipc.rs
```diff
@@ -270,10 +258,26 @@ impl<DB: Blockstore> GatewayCaller<DB> {
...
+}
+
+/// Total amount of tokens to burn as a result of bottom-up messages leaving the subnet.
+///
+/// Fees are ignored, which suggests that they will be paid similarly to gas,
+/// from the balance of the sender already present in the parent subnet.
+pub fn tokens_to_burn(msgs: &[gateway_getter_facet::CrossMsg]) -> TokenAmount {
+    msgs.iter()
+        .fold(TokenAmount::from_atto(0), |mut total, msg| {
+            total += from_eth::to_fvm_tokens(&msg.message.value);
```
adlrocha, line 280: Same here, we should reduce from the circulating supply the fees and the value of the cross-net message (you could also look at the `value` field for the wrapping release message, but this is not the case as you are directly looking at what is being propagated, but for context: https://github.com/consensus-shipyard/ipc-solidity-actors/blob/dev/src/gateway/GatewayManagerFacet.sol#L143-L150)

adlrocha (commented):
A few caveats, but overall LGTM. Thank you!

fendermint/vm/interpreter/src/fvm/checkpoint.rs
aakoshh: Ah great, thanks for the link. I see you are sending to the `BURNT_FUNDS_ACTOR`. In that case we have two options:
* We don't adjust the circulating supply here, rather, we look up the balance of the burnt funds actor on each block and subtract it from the circulating supply stored in the `AppState`. That way _anybody_ can send funds to the burnt funds actor, not just IPC, and we'll get the correct value *(see caveat below). 
* We decrease the circulaing supply here which makes the one in `AppState` be in sync with IPC itself, however, the actual circulating supply can be lower because anyone else can burn funds by sending to the burnt funds actor, and now unfortunately we can't subtract it as it would go negative. 

*Caveat: both cases can misrepresent the actual circulating supply if people use a random account for burning funds rather than the well known burnt funds address.

fendermint/vm/interpreter/src/fvm/state/ipc.rs
aakoshh: Okay, so you are saying that if there are `fees` collected in the parent subnet, then they will be taken there from the sender, locked in the subnet actor, and moved into the subnet, where they will be distributed to the subnet validators or something.

fendermint/vm/interpreter/src/fvm/state/ipc.rs
aakoshh: Great link, thanks!

fendermint/vm/actor_interface/src/ipc.rs
aakoshh: I didn't think you have to check anything, but if you do, you can also use the one [implemented on Vec<getter::CrossMsg>](https://github.com/consensus-shipyard/fendermint/pull/331/files#diff-ef5f1e66a6fde4f0d3bfec4815a10f0aa82faee5fd48a38f07a2bfed464bf720R241)

adlrocha: We currently don't, but in case we want to do a verification of some kind, so that we are aware of the existence of this :pray:

fendermint/vm/interpreter/src/fvm/checkpoint.rs
adlrocha: I think both options are good. If it is not hard to implement I personally like option 1. If we want to be super accurate, we could also check the balance for the EVM zero address and not only the `BURNT_FUNDS_ACTOR`. This gives us an IPC-agnostic solution :) 

That being said, I think this is the most accurate we could go, so I wouldn't worry about trying to be 100% accurate (not even Bitcoin has this knowledge, as there maybe funds lost forever on accounts for which private keys have been losted, and strictly speaking these don't belong to the circulating supply).

fendermint/vm/interpreter/src/fvm/state/ipc.rs
adlrocha: Correct. They will de deducted from the balance in the parent, increasing the circulating supply for the subnet. Then in the subnet, we can distributed however we want (although I am aware that we are currently not doing anything with them because we only support `fund`).

fendermint/vm/interpreter/src/fvm/checkpoint.rs
aakoshh: I left it as it is, which means it's 100% accurate with what IPC is doing, making at least that one a known quantity.

adlrocha: That's fine, I just realized one could always explicitly check the balances of those addresses if they want an accurate value.

adlrocha (approved):
Good one, thanks!

Issue #324, opened 2023-10-24
Integrate fevm-contract-tests in CI


raulk:
[This contract suite](https://github.com/filecoin-project/fevm-contract-tests) was built from OpenZeppelin + other test cases, to validate functional correctness of the Filecoin FEVM running on Lotus. It can be easily repurposed and integrated in CI here to validate functional correctness of Fendermint.

PR #332, opened 2023-10-26
FM-220: Remove tx.hash from the filter, index block.hash

### LINKED ISSUE #220: Filtering uses tx.hash for block hash
In `FilterKind::to_queries` the `tx.hash` is used as if it was a block hash, but I think that's just an oversight, it's actually the transaction hash. For example in https://docs.tendermint.com/v0.34/rpc/#/Websocket/subscribe it says `tm.event = 'Tx' AND tx.hash = 'XYZ' # single transaction`

To be able to search transactions by block hash we have to emit a `block.hash` event.

On the other hand `to_queries` is only used in subscriptions that listen to future events, and it's impossible to predict block hashes, so this filter will practically only be used in `eth_getLogs`, where it's handled completely differently by looking up the block by hash.

aakoshh:
Closes #220  

Initially I wanted to add `block.hash` to the indexes but decided against it, and just removed the incorrect filtering of `tx.hash`. 

Then I realised that we can use an index on `block.hash` in filters, where it didn't use to be available, but Ethereum clients expect it.

### DIFF
--- a/fendermint/app/src/app.rs
+++ b/fendermint/app/src/app.rs
@@ -303,6 +303,15 @@ where
         guard.take().expect("exec state empty")
     }
 
+    /// Apply a function on the state, if it exists.
+    fn map_exec_state<T, F>(&self, f: F) -> Option<T>
+    where
+        F: FnOnce(&FvmExecState<SS>) -> T,
+    {
+        let guard = self.exec_state.lock().expect("mutex poisoned");
+        guard.as_ref().map(f)
+    }
+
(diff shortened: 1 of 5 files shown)


adlrocha (approved):
Thanks! I am wondering how can we test these features like we did with the Lilypad integration. Raúl pointed me to this https://github.com/filecoin-project/fevm-contract-tests (tracked in #324)

aakoshh:
This could be added to `ethers.rs` where logs are returned.

PR #337, opened 2023-10-26
FIX: Remove unnecessary lib

aakoshh:
It was enough for the builder image to have `libssl3`. 

Follow up to https://github.com/consensus-shipyard/fendermint/pull/336

### DIFF
--- a/docker/local.Dockerfile
+++ b/docker/local.Dockerfile
@@ -4,7 +4,7 @@
 FROM rust:bookworm as builder
 
 RUN apt-get update && \
-  apt-get install -y build-essential clang cmake protobuf-compiler libssl3 && \
+  apt-get install -y build-essential clang cmake protobuf-compiler && \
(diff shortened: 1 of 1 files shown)


raulk:
Looks like the diff is doing the opposite of what the description says? This removed `libssl3` from the builder image.

raulk:
So the PR desc and commit message are contradictory ;-) And the Docker build passed on CI, so this commit was correct. I guess it's my env that's wonky.

aakoshh:
You are right, in my last attempt locally it seemed to work if only the runner had the library. Sorry for the confusion.

As discussed, this `local.Dockerfile` isn't tested on CI, that's what `ci.Dockerfile` is for which copies the built executable into an image that is based on `ubuntu`, same as what the CI runner is using. We should investigate again if it's practical to build and test this other image on CI, it's probably smaller in size. Question is how well it works with the caching mechanism used by the CI build.

PR #342, opened 2023-10-27
FIX: Set the Tendermint RPC URL for the fendermint container

aakoshh:
@adlrocha has some trouble configuring his fendermint container to connect to the cometbft one. 

This PR adds an example of configuring the URL in the smoke tests. They aren't used in the smoke tests, but we should see them printed in the logs.

The URL can be configure either as:
1. `TENDERMINT_RPC_URL`, because the `fendermint rpc` and `fendermint eth` commands take it in this format - I thought it might be something people dealing with CometBFT (Tendermint Core at the time) might have. 
2. `FM_TENDERMINT_RPC_URL`, because it's part of the config files and everything there can be overwritten with `FM_` prefix

### DIFF
--- a/infra/Makefile.toml
+++ b/infra/Makefile.toml
@@ -60,7 +60,7 @@ ACTORS_BUNDLE = "${CARGO_MAKE_WORKSPACE_WORKING_DIRECTORY}/../builtin-actors/out
 # If this wasn't present, any wait task is skipped.
 CARGO_MAKE_WAIT_MILLISECONDS = 5000
 # This wait time seems to work locally.
-CMT_WAIT_MILLIS = 10000
+CMT_WAIT_MILLIS = 30000
 # Keep example logs to a minimum.
 VERBOSITY = ""
 
--- a/infra/scripts/fendermint.toml
+++ b/infra/scripts/fendermint.toml
@@ -21,7 +21,7 @@ docker run \
 dependencies = ["docker-network-create", "fendermint-deps"]
 
 [tasks.fendermint-start-ipc]
-extend = "fendermint-run"
+extend = "fendermint-run-ipc"
 env = { "ENTRY" = "fendermint", "CMD" = "run", "FLAGS" = "-d" }
 
 [tasks.fendermint-run-ipc]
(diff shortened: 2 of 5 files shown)


adlrocha:
Added a few additional fixes: 
* Installed `ca-certificates` in runner so we can interact with https endpoints like calibration.
* Added top-down env variables expected by the top-down sync module
* Increase default wait to prevent the eth api from crashing when starting a validator

adlrocha (approved):

infra/scripts/fendermint.toml
```diff
@@ -40,9 +41,17 @@ docker run \
   --env FM_IPC__TOPDOWN__PARENT_HTTP_ENDPOINT=${PARENT_ENDPOINT} \
   --env FM_IPC__TOPDOWN__PARENT_REGISTRY=${PARENT_REGISTRY} \
   --env FM_IPC__TOPDOWN__PARENT_GATEWAY=${PARENT_GATEWAY} \
+  --env FM_IPC__TOPDOWN__EXPONENTIAL_BACK_OFF=5 \
+  --env FM_IPC__TOPDOWN__EXPONENTIAL_RETRY_LIMIT=5 \
+  --env FM_IPC__TOPDOWN__POLLING_INTERVAL=10 \
+  --env FM_ABCI__LISTEN__HOST=0.0.0.0 \
+  --env FM_ETH__LISTEN__HOST=0.0.0.0 \
```
aakoshh, line 48: You shouldn't have to set these, they are part of the [Dockerfile](https://github.com/consensus-shipyard/fendermint/blob/37bcaa42d7a9b9941fce10b36f309ad9b124edb4/docker/ci.Dockerfile#L24-L25)

infra/Makefile.toml
```diff
@@ -60,7 +60,7 @@ ACTORS_BUNDLE = "${CARGO_MAKE_WORKSPACE_WORKING_DIRECTORY}/../builtin-actors/out
 # If this wasn't present, any wait task is skipped.
 CARGO_MAKE_WAIT_MILLISECONDS = 5000
 # This wait time seems to work locally.
-CMT_WAIT_MILLIS = 10000
+CMT_WAIT_MILLIS = 30000
```
aakoshh, line 63: Wow that's long! 

https://github.com/consensus-shipyard/fendermint/pull/344 should help

infra/scripts/fendermint.toml
adlrocha: I saw it, but unfortunately for some reason they weren't being picked up. I will test it again though, now that I we know what was happening with the connections. Maybe it was something else and I was lucky and I added both changes at the same time (which could definitely be the case with all the try and error :) )

PR #343, opened 2023-10-27
FM-341: Delay broadcasting checkpoint sig after commit

### LINKED ISSUE #341: Signature broadcasted before block commit
@cryptoAtwill discovered a race condition where the validators try to broadcast their signatures for the checkpoint added in `end_block` _before_ the `commit` is finished, and therefore fail to estimate the gas requirement because the checkpoint is not in the actual ledger yet that `check_tx` will run on. 

We can fix this by waiting for the next block height to be ready in CometBFT.

Also add some actual sleep time between retries, for similar situations.

aakoshh:
Fixes #341 

The PR adds a step to `checkpoint::broadcast_incomplete_signatures` to call `wait_for_commit` with the highest checkpointed height +1, which should make sure that the block where the checkpoint has been added to the ledger has been executed. 

It also adds a `retry_delay` setting to the `Broadcaster` with a default 2s in the configuration file, so in case we run into an issue like this, it gives the thing a chance to at least create a block for the state to change. Previously there was no delay because I thought it's only the nonce that might be having a race condition, but this situation clearly won't resolve itself by an immediate restart.

### DIFF
--- a/fendermint/app/config/default.toml
+++ b/fendermint/app/config/default.toml
@@ -40,6 +40,8 @@ state_hist_size = 0
 [broadcast]
 # Maximum number of times to retry broadcasting a transaction after failure.
 max_retries = 5
+# Tome to wait between retries, in seconds. It should roughly correspond to the block interval.
+retry_delay = 2
 
 # FVM configuration
 [fvm]
--- a/fendermint/app/settings/src/lib.rs
+++ b/fendermint/app/settings/src/lib.rs
@@ -92,10 +92,14 @@ pub struct DbSettings {
 /// Settings affecting how we deal with failures in trying to send transactions to the local CometBFT node.
 /// It is not expected to be unavailable, however we might get into race conditions about the nonce which
 /// would need us to try creating a completely new transaction and try again.
+#[serde_as]
 #[derive(Debug, Deserialize, Clone)]
 pub struct BroadcastSettings {
     /// Number of times to retry broadcasting a transaction.
     pub max_retries: u8,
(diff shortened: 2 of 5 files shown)


cryptoAtwill (approved):

cryptoAtwill:
@aakoshh Btw, just a question, since the create checkpoint is called by the current validator, that means it's not really broadcasted. So when the signature is broadcasted, can we ensure the checkpoint is already created in other peers?

adlrocha (approved):

aakoshh:
> since the create checkpoint is called by the current validator, that means it's not really broadcasted.

No, the checkpoint is [created](https://github.com/consensus-shipyard/fendermint/blob/0d5d4452def4831b9dcc1e54538b5284c63c21dc/fendermint/vm/interpreter/src/fvm/checkpoint.rs#L116) by everyone who runs a full node by directly inserting into their ledger during [end_block](https://github.com/consensus-shipyard/fendermint/blob/0d5d4452def4831b9dcc1e54538b5284c63c21dc/fendermint/vm/interpreter/src/fvm/exec.rs#L135-L137). This is necessary so that every node arrives at the same state hash. There is nothing validator specific about the checkpoint, therefore it's not created by a broadcast.

> So when the signature is broadcasted, can we ensure the checkpoint is already created in other peers?

Yes, as soon as the current block is committed, everyone who is caught up with the chain can run the transaction that adds the signature to it. Any validator who isn't caught up with the chain would be unable to propose a block on top of the committed block, so they can't propose to execute a transaction on an earlier state.

PR #344, opened 2023-10-27
FIX: Add reconnection to Ethereum API startup

aakoshh:
The Ethereum API can fail to start if CometBFT is not ready for connection, which is why the smoke tests waited 10-20s after starting `cometbft` before starting the `ethapi` container. 

This PR adds a limited number of retries to the Ethereum API to help with this, by default 5 retries with 5 second delays between them, which should take it above the 20s that worked on CI. 

The `cometbft-wait` task is still present, but it's moved down, so we see the effect of of retries in the log and don't run tests until the API is ready.

### DIFF
--- a/fendermint/app/options/src/eth.rs
+++ b/fendermint/app/options/src/eth.rs
@@ -27,5 +27,13 @@ pub enum EthCommands {
         /// Tendermint node's RPC endpoint.
         #[arg(long)]
         proxy_url: Option<Url>,
+
+        /// Maximum number of times to try to connect to the websocket.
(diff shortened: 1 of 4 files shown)


cryptoAtwill (approved):

adlrocha (approved):
Note to self (hopefully I get a notification): I need to test once this is merged into master that the ETH api works with a lower wait. I tried to test it directly against this PR but we are missing some of the fixes that were merged on Friday.

aakoshh:
For me it always worked with 10s wait in the smoke tests. 30s would be really long wait while testing :flushed:

PR #345, opened 2023-10-27
FM-340: Don't send duplicate signatures

### LINKED ISSUE #340: Validators re-send their checkpoint signatures even if they signed
The checkpointing logic retrieves all incomplete checkpoints from the ledger to make sure signatures are eventually submitted. However it doesn't check whether the validator has already sent their signature before, therefore if the quorum takes a long time to reach, they will burn money re-sending their already existing signatures. 

This may not be true if the `call` we use to estimate the gas cost ends with a `revert SignatureReplay();` and we stop the sending because of that. Unfortunately that might prevent all other signatures from being sent too!

Make sure that validators don't send signatures for already signed checkpoints.

aakoshh:
Fixes #340 

Added a `checkpoint::unsigned_signatures` method that collects incomplete signatures that the current validator hasn't signed, so that they don't try to send their signatures again. 

It turns out that they would not have incurred cost because gas estimation would have failed, as the contract rejects duplicates. But this way they would skip it altogether.

### DIFF
--- a/fendermint/vm/interpreter/src/fvm/checkpoint.rs
+++ b/fendermint/vm/interpreter/src/fvm/checkpoint.rs
@@ -5,6 +5,7 @@ use std::collections::HashMap;
 
 use anyhow::{anyhow, Context};
 use fendermint_crypto::PublicKey;
+use fendermint_vm_actor_interface::eam::EthAddress;
 use fendermint_vm_actor_interface::ipc::AbiHash;
 use fendermint_vm_genesis::Collateral;
 use fendermint_vm_genesis::PowerScale;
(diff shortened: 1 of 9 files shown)


fendermint/vm/interpreter/src/fvm/checkpoint.rs
```diff
@@ -138,6 +139,38 @@ where
...
+    let mut unsigned_checkpoints = Vec::new();
+    let validator_addr = EthAddress::from(validator_key);
+
+    let incomplete_checkpoints = gateway
+        .incomplete_checkpoints(state)
+        .context("failed to fetch incomplete checkpoints")?;
+
+    for cp in incomplete_checkpoints {
+        let signatories = gateway
+            .checkpoint_signatories(state, cp.block_height)
```
cryptoAtwill, line 163: not for now, but a method `has_signed(state, height, addr)` can be more effective.

cryptoAtwill (approved):

adlrocha (approved):

fendermint/vm/interpreter/src/fvm/checkpoint.rs
```diff
@@ -138,6 +139,38 @@ where
...
+            .context("failed to get checkpoint signatories")?;
+
+        if signatories.contains(&validator_addr) {
+            unsigned_checkpoints.push(cp);
+        }
```
aakoshh, line 168: Waaah, it's the opposite!! :sob:

PR #348, opened 2023-10-30
Convert power to collateral

cryptoAtwill:
Convert power to collateral when creating checkpoint.

### DIFF
--- a/fendermint/vm/genesis/src/lib.rs
+++ b/fendermint/vm/genesis/src/lib.rs
@@ -7,9 +7,11 @@ use anyhow::anyhow;
 use fvm_shared::bigint::{BigInt, Integer};
 use serde::{Deserialize, Serialize};
 use serde_with::serde_as;
+use std::ops::Mul;
 
 use fvm_shared::version::NetworkVersion;
 use fvm_shared::{address::Address, econ::TokenAmount};
+use num_traits::ToPrimitive;
 
 use fendermint_crypto::{normalize_public_key, PublicKey};
 use fendermint_vm_core::Timestamp;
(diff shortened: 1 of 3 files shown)


adlrocha (commented):
Let's wait for @aakoshh's review that he has more context into this part of the code. So the issue was that we were passing the power value instead of the collateral to the contract and that's why it was being reverted?

fendermint/vm/genesis/src/lib.rs
```diff
@@ -107,6 +120,11 @@ impl Collateral {
         Power(power.try_into().expect("clipped to u64::MAX"))
     }
 
+    /// Convert the collateral into u64 representation. Will cap at u64::MAX.
+    pub fn u64(&self) -> u64 {
+        self.0.atto().to_u64().unwrap_or(u64::MAX)
+    }
```
aakoshh, line 126: What would be the possible purpose of this function? You can't cap atto to u64, it would be worth like less than $100. This is why the power scale was introduced.

fendermint/vm/genesis/src/lib.rs
```diff
@@ -92,6 +94,17 @@ pub struct Collateral(#[serde_as(as = "IsHumanReadable")] pub TokenAmount);
 #[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq, Copy)]
 pub struct Power(pub u64);
 
+impl Power {
+    /// Convert from [Power] to [Collateral] by specifying the number of significant
+    /// decimal places per FIL that grant 1 power.
+    pub fn into_collateral(self: Power, scale: PowerScale) -> Collateral {
```
aakoshh, line 100: I don't think we should have a function like this. The power scale works one way, but it's a lossy transformation, we can't expect it go back and be the same as the collateral. It should work without this.

aakoshh (commented):
This PR should not be necessary. The Merkle tree is built on the `Power` by design, the total is passed to the checkpoint creation function and it does not have to be in collaterals, because the only reason for it to exist is so that we can submit Merkle proofs, which are also based on `Power`. 

The `majority_percentage` is a value `[51, 100]` and should work on power too, not just collateral. 

I have said this for a long time that we can instead convert to power in the parent, if we have to, but just to check checkpoint signature we should just make sure that any rounding is happening up, not down, so it doesn't degenerate into zero. It would  help with someone trying to join with less than 0.001 FIL too.

fendermint/vm/interpreter/src/fvm/state/ipc.rs
```diff
@@ -110,9 +111,16 @@ impl<DB: Blockstore> GatewayCaller<DB> {
         let total_power = power_table.iter().fold(et::U256::zero(), |p, v| {
             p.saturating_add(et::U256::from(v.power.0))
         });
+        let collateral = Power(total_power.as_u64())
+            .into_collateral(power_scale)
+            .u64();
```
aakoshh, line 116: Why would you convert to `u64` here?

aakoshh (commented):
Can you explain why the contract revered without this? As I understand the threshold was 0, to which I posited that the power scale transformed collateral into a value where multiplying by 0.65 rounded to 0, whereas i think it should round up, like I do in Rust.

aakoshh (commented):
The PR also doesn't change the signature submission function, so with this change nobody can submit correct signatures, which are still based on power.

cryptoAtwill:
As discussed in slack, I misunderstood the idea of `power`, closing this for now.

PR #370, opened 2023-10-31
Fill gas limit for call

cryptoAtwill:
Fill the gas limit when it's 0 for `call` type txn. This should fix: #352.

### DIFF
--- a/fendermint/vm/interpreter/src/fvm/state/query.rs
+++ b/fendermint/vm/interpreter/src/fvm/state/query.rs
@@ -182,6 +182,14 @@ where
                     });
                 }
             }
+
+            // If the gas_limit is zero, set it to the block gas limit so that call will not hit
+            // gas limit not set error. It is possible, in the future, to estimate the gas limit
+            // based on the account balance and base fee + premium for higher accuracy.
+            if msg.gas_limit == 0 {
+                msg.gas_limit = fvm_shared::BLOCK_GAS_LIMIT;
+            }
+
             if msg.from == SYSTEM_ACTOR_ADDR {
                 // Explicit execution requires `from` to be an account kind.
                 s.execute_implicit(msg)


fendermint/vm/interpreter/src/fvm/state/query.rs
```diff
@@ -182,6 +182,14 @@ where
                     });
                 }
             }
+
+            // If the gas_limit is zero, set it to the block gas limit so that call will not hit
+            // gas limit not set error. It is possible, in the future, to estimate the gas limit
+            // based on the account balance and base fee + premium for higher accuracy.
```
aakoshh, line 188: Currently it's possible to set the gas prices to zero. Ethers does the estimation if you ask it, but I reckon we shouldn't complicate _this_ handler with estimations. That kind of shenanigans should really go into the Ethereum API.

fendermint/vm/interpreter/src/fvm/state/query.rs
```diff
@@ -182,6 +182,14 @@ where
                     });
                 }
             }
+
+            // If the gas_limit is zero, set it to the block gas limit so that call will not hit
```
adlrocha, line 186: Additional context for @aakoshh, when he reviews: this is what `ethers` does. Instead of estimating accurately it just uses the `BLOCK_GAS_LIMIT` and lets the validators be smart on what messages they include in their block proposal (I know we currently don't have message selection, but it seems like the most straightforward solution).

adlrocha (approved):

aakoshh (approved):
Thank you :pray:

fendermint/vm/interpreter/src/fvm/state/query.rs
adlrocha: hahaha there was a race in the review. I left a similar comment below, but you can disregard it.

fendermint/vm/interpreter/src/fvm/state/query.rs
aakoshh: Are you talking about `eth_call`, specifically? When I worked on the Ethereum API I saw it do the estimation if I tried to send a transaction, it's all [here](https://github.com/gakonst/ethers-rs/blob/efdc8d43318b818178c93a19a42b2b7e49e678eb/ethers-providers/src/rpc/provider.rs#L297-L323). Those use the `eth_estimateGas` and `eth_feeHistory` endpoints. I don't remember what it does for calls, but I haven't seen any block gas limit being applied.

AlcibiadesCleinias (approved):

fendermint/vm/interpreter/src/fvm/state/query.rs
cryptoAtwill: I think you guys misunderstood me by estimation, I'm referring more to [this](https://github.com/paradigmxyz/reth/blob/9ca44efe9f4e8768e84e855f8679e65c14a8b350/crates/rpc/rpc/src/eth/revm_utils.rs#L229) than the estimate gas.

PR #372, opened 2023-10-31
FM-355: Snapshot manager

### LINKED ISSUE #355: Snapshot manager
Create a snapshot manager component similar to `fendermint/vm/resolver` that handles the integration with the `App:
* receives notifications about newly committed blocks
* lists completed snapshots
* returns chunks of the snapshots when asked
* receives chunks of snapshots from remote nodes

We can again consider using STM for example to handle notifications by putting them into a `TVar` and using `retry` to get notifications when it changes, but also skip heights if the snapshotter is busy, instead of piling up work.

### LINKED ISSUE #358: Skip snapshots while syncing
While our node is catching up with other nodes on the network, there is no reason to create snapshots, it's not going to be news to anyone and we don't know how many more there will be.

We can add a `TVar<bool>` to the snapshot manager indicating sync status and kick off a background task to poll CometBFT; once it indicates it is no longer syncing we flip the value of the flag. Until then, we don't start snapshotting upon notifications, but rather use `retry` to be notified if there is a new snapshot _or_ the syncing status is finished, at which point we can snapshot the last eligible height.

aakoshh:
Closes #355 
Closes #358 

This is just a skeleton for a snapshot manager. It listens to notifications about new heights where snapshots should be taken and exports the data to a CAR file. It's not wired in, and doesn't implement all the functions which are covered by upcoming tasks, such as removing old files, listing available snapshots, etc.

### DIFF
--- a/fendermint/vm/snapshot/src/manager.rs
+++ b/fendermint/vm/snapshot/src/manager.rs
@@ -4,23 +4,17 @@
 use std::path::PathBuf;
 use std::time::Duration;
 
+use anyhow::Context;
 use async_stm::{atomically, retry, TVar};
-use fendermint_vm_interpreter::fvm::state::snapshot::{BlockHeight, BlockStateParams};
+use fendermint_vm_interpreter::fvm::state::snapshot::{BlockHeight, BlockStateParams, Snapshot};
 use fendermint_vm_interpreter::fvm::state::FvmStateParams;
+use fvm_ipld_blockstore::Blockstore;
 use tendermint_rpc::Client;
-
-const SYNC_POLL_INTERVAL: Duration = Duration::from_secs(60);
 
 /// State of snapshots, including the list of available completed ones
 /// and the next eligible height.
 #[derive(Clone)]
 struct SnapshotState {
-    /// Snapshotted heights are a multiple of the interval.
-    ///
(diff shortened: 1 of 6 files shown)


fendermint/vm/snapshot/src/manager.rs
```diff
@@ -0,0 +1,197 @@
...
+        &self,
+        height: BlockHeight,
+        params: FvmStateParams,
+    ) -> anyhow::Result<()> {
+        let snapshot = Snapshot::new(self.store.clone(), params, height)
+            .context("failed to create snapshot")?;
+
+        let file_name = format!("snapshot-{height}.car");
+        let file_path = self.snapshot_dir.join(file_name.clone());
+        let temp_dir = tempfile::tempdir().context("failed to create temp dir for snapshot")?;
```
adlrocha, line 145: Why do you need to have an intermediate temp file, in case it fails generating the car?

adlrocha (approved):

fendermint/vm/snapshot/src/manager.rs
aakoshh: Yes, I thought if the application crashes in the middle of exporting a snapshot, this should make sure the half-finished dump will be cleaned up.

PR #373, opened 2023-10-31
fix sed separator

dnkolegov:
This PR changes the sed separator to not collide with the subnet ID.

### DIFF
--- a/infra/Makefile.toml
+++ b/infra/Makefile.toml
@@ -18,7 +18,7 @@ default_to_workspace = false
 # General network-specific parameters
 SUBNET_ID = { value = "/r0", condition = { env_not_set = ["SUBNET_ID"] } }
 # The network name is derived from the SUBNET_ID, replacing slashes with dashes, and dropping the first dash if any.
-NETWORK_NAME = { script = ["echo $SUBNET_ID | sed -e 's/\\//-/g' -e 's/^-//1'"] }
+NETWORK_NAME = { script = ["echo $SUBNET_ID | sed -e 's|/|-|g' -e 's|^-||1'"] }
 
 BALANCE = { value = "1000", condition = { env_not_set = ["BALANCE"] }}
 BASE_FEE = { value = "1000", condition = { env_not_set = ["BASE_FEE"] }}
--- a/infra/scripts/cometbft.toml
+++ b/infra/scripts/cometbft.toml
@@ -32,6 +32,8 @@ docker run \
 dependencies = ["cometbft-pull", "docker-network-create"]
 
 [tasks.cometbft-node-id]
+# This script gets an external IP address, checks whether it is an IPv6 address,
+# and if so, adds square brackets to use it with a port number.
 script = """
 id=`docker exec ${CMT_CONTAINER_NAME} cometbft show-node-id`
 ip=`curl https://ifconfig.me`
--- a/infra/scripts/cometbft.toml
+++ b/infra/scripts/cometbft.toml
@@ -34,7 +34,15 @@ dependencies = ["cometbft-pull", "docker-network-create"]
 [tasks.cometbft-node-id]
 script = """
 id=`docker exec ${CMT_CONTAINER_NAME} cometbft show-node-id`
-ip=`curl ifconfig.me`
+ip=`curl https://ifconfig.me`
+
+set +e
+echo ${ip} | grep --quiet ":"
+if [ $? = 0 ]; then
+  ip="[$ip]"
+fi
+set -e
+
 echo $id@$ip:${CMT_P2P_HOST_PORT}
 """
 
--- a/infra/scripts/node.toml
+++ b/infra/scripts/node.toml
@@ -51,10 +51,10 @@ mkdir -p $CMT_DIR
 
 [tasks.seed-enable]
 script="""
-sed -i'bak' "s/seed_mode = false/seed_mode = true/" ${CMT_DIR}/config/config.toml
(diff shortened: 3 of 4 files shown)


infra/Makefile.toml
```diff
@@ -18,7 +18,7 @@ default_to_workspace = false
 # General network-specific parameters
 SUBNET_ID = { value = "/r0", condition = { env_not_set = ["SUBNET_ID"] } }
 # The network name is derived from the SUBNET_ID, replacing slashes with dashes, and dropping the first dash if any.
-NETWORK_NAME = { script = ["echo $SUBNET_ID | sed -e 's/\\//-/g' -e 's/^-//1'"] }
+NETWORK_NAME = { script = ["echo $SUBNET_ID | sed -e 's|/|-|g' -e 's|^-||1'"] }
```
aakoshh, line 21: What does it do to change `s/.../.../g` into `s|...|...|g`? 

(I have no idea what the whole expression does, but that's just me being lame).

infra/scripts/cometbft.toml
```diff
@@ -34,7 +34,15 @@ dependencies = ["cometbft-pull", "docker-network-create"]
...
 id=`docker exec ${CMT_CONTAINER_NAME} cometbft show-node-id`
-ip=`curl ifconfig.me`
+ip=`curl https://ifconfig.me`
+
+set +e
+echo ${ip} | grep --quiet ":"
+if [ $? = 0 ]; then
+  ip="[$ip]"
+fi
+set -e
```
aakoshh, line 44: What does this do?

infra/scripts/node.toml
```diff
@@ -51,10 +51,10 @@ mkdir -p $CMT_DIR
...
 [tasks.seed-enable]
 script="""
-sed -i'bak' "s/seed_mode = false/seed_mode = true/" ${CMT_DIR}/config/config.toml
+sed -i'' -e "s|seed_mode = false|seed_mode = true|" ${CMT_DIR}/config/config.toml
 """
 
 [tasks.set-seeds]
 script="""
-sed -i 's/seeds = ""/seeds = "${BOOTSTRAPS}"/g' ${CMT_DIR}/config/config.toml
+sed -i'' -e 's|seeds = ""|seeds = "${BOOTSTRAPS}"|g' ${CMT_DIR}/config/config.toml
```
aakoshh, line 59: What's the `-e` doing? (Sorry)

infra/Makefile.toml
dnkolegov: I added this to use a different separator than in the subnet names ("/").

infra/scripts/cometbft.toml
dnkolegov: It checks whether the external IP address is an ipv6 address, and if so, it adds square brackets. I will add a comment. In IPv6, you need to wrap an address to make an IP:PORT pair, as far as I am aware.

infra/scripts/node.toml
dnkolegov: It is used to add a sed script. The goal of this is to create a sed command that works on MacOS and GNU/Linux systems.

infra/Makefile.toml
aakoshh: The comment already says "replacing slashes with dashes". What are you replacing it with now? Or did it not work? 

I was also asking in more general: what does `|` do compared to `/` with `sed`?

infra/scripts/cometbft.toml
aakoshh: I see, thanks. Yeah, definitely worth a comment :pray:

infra/scripts/testnet.toml
```diff
@@ -107,7 +107,7 @@ export CMT_PERSISTENT_PEERS=`cat $BASE_DIR/peers`
 echo Persistent peers: $CMT_PERSISTENT_PEERS
 
 for i in $(seq 0 3); do
-    sed -i'bak' "s/persistent_peers = \\"\\"/persistent_peers = \\"$CMT_PERSISTENT_PEERS\\"/" $BASE_DIR/node${i}/cometbft/config/config.toml
+    sed -i'bak' "s|persistent_peers = \\"\\"|persistent_peers = \\"$CMT_PERSISTENT_PEERS\\"|" $BASE_DIR/node${i}/cometbft/config/config.toml
```
aakoshh, line 110: Isn't `-e` required here? 
Why do you keep `bak` here but not [here](https://github.com/consensus-shipyard/fendermint/pull/373/files#diff-c7acb1a1a82d3b2fa4b0b12e343395aee71560f17d0a888a933fc3f2ecea4c3cR59)?

infra/Makefile.toml
dnkolegov: `|` or `/` are just separators in sed script. The semantics is the same

infra/scripts/testnet.toml
dnkolegov: Will fix

aakoshh:
Okay so the problem wasn't with the subnet ID colliding with the sed separator and giving false results but that it didn't work on MacOS. 

I can confirm it works on Ubuntu:

```console
❯ echo /r123/f410123abd/f0123 | sed -e 's/\\//-/g' -e 's/^-//1'
r123-f410123abd-f0123

❯ echo /r123/f410123abd/f0123 | sed -e 's|/|-|g' -e 's|^-||1'
r123-f410123abd-f0123
```

aakoshh (approved):
Thanks for explaining. I can't try on MacOS but Ubuntu is fine :point_up_2:

PR #375, opened 2023-11-01
Fill sender if it's None

cryptoAtwill:
Fill the sender with SYSTEM_ACTOR_ADDR if it is None.

Reasons for doing this:
- `eth_call` will work as not every time `from` will be filled.

### DIFF
--- a/fendermint/vm/message/src/conv/from_eth.rs
+++ b/fendermint/vm/message/src/conv/from_eth.rs
@@ -37,12 +37,8 @@ pub fn to_fvm_message(tx: &Eip1559TransactionRequest) -> anyhow::Result<Message>
     // The `from` of the transaction is inferred from the signature.
     // As long as the client and the server use the same hashing scheme,
     // this should be usable as a delegated address.
-    let from = if let Some(from) = tx.from {
-        to_fvm_address(from)
-    } else {
-        // This is similar to https://github.com/filecoin-project/lotus/blob/master/node/impl/full/eth_utils.go#L124
-        system::SYSTEM_ACTOR_ADDR
-    };
+    // If none, use SYSTEM_ACTOR_ADDR. This is similar to https://github.com/filecoin-project/lotus/blob/master/node/impl/full/eth_utils.go#L124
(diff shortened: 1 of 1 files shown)


fendermint/vm/message/src/conv/from_eth.rs
```diff
@@ -37,7 +37,12 @@ pub fn to_fvm_message(tx: &Eip1559TransactionRequest) -> anyhow::Result<Message>
     // The `from` of the transaction is inferred from the signature.
     // As long as the client and the server use the same hashing scheme,
     // this should be usable as a delegated address.
-    let from = to_fvm_address(tx.from.unwrap_or_default());
+    let from = if let Some(from) = tx.from {
+        to_fvm_address(from)
+    } else {
+        // This is similar to https://github.com/filecoin-project/lotus/blob/master/node/impl/full/eth_utils.go#L124
```
adlrocha, line 43: Good one, that way we use the same as Lotus and the behavior is somewhat predictable.

adlrocha (approved):
LGTM. Feel free to merge

aakoshh (approved):
Cheers :+1:

fendermint/vm/message/src/conv/from_eth.rs
```diff
@@ -37,7 +37,12 @@ pub fn to_fvm_message(tx: &Eip1559TransactionRequest) -> anyhow::Result<Message>
     // The `from` of the transaction is inferred from the signature.
     // As long as the client and the server use the same hashing scheme,
     // this should be usable as a delegated address.
-    let from = to_fvm_address(tx.from.unwrap_or_default());
+    let from = if let Some(from) = tx.from {
```
raulk, line 40: Why not use [`Option::map_or_else()`](https://doc.rust-lang.org/std/option/enum.Option.html#method.map_or_else)?

cryptoAtwill: just a preference, but I think `map_or` will fit nicely here. thx for the tip.

raulk (approved):

raulk:
Unfortunately this caused a regression. The address we need to use here is the zero Eth address, not the address of the system actor. 0x00..00 should be an Ethereum account.

PR #376, opened 2023-11-01
FM-347: get child subnet genesis balance from parent

adlrocha:
Depends on https://github.com/consensus-shipyard/ipc/pull/361

This PR modifies the child subnet genesis generation to also pull information from the parent about the genesis balances to be initialized in genesis in the subnet.


aakoshh (approved):
LGTM!

adlrocha:
Test it manually in calibration and it works fine :partying_face: This is actually a really cool feature!

PR #379, opened 2023-11-02
DOCKER: Delete untagged images after build

aakoshh:
It turns out that https://github.com/consensus-shipyard/fendermint/pkgs/container/fendermint/versions has images going back months now. 

I think we are safe to remove everything that has been orphaned. Hopefully this action does just that: https://github.com/Chizkiyahu/delete-untagged-ghcr-action

### Note

I left the PR as _Draft_ because there is no point in testing it, it changes nothing apart from removing stuff if executed on `main`.


jsoares (approved):
lgtm

PR #380, opened 2023-11-02
FM-350: eth_getTransactionCount should return 0 for nonexistent senders

maciejwitowski:
Fixes https://github.com/consensus-shipyard/fendermint/issues/350

Consistent with Lotus https://github.com/filecoin-project/lotus/blob/master/node/impl/full/eth.go#L360


aakoshh (approved):
Highly appreciated :pray:

PR #382, opened 2023-11-02
Add IP tool

dnkolegov:
This PR adds a new IP tool to determine the external IP address of the bootstrap.

The idea of the tool is as follows:

1. Send a request to several web services, returning the IP address of the client in the HTTP body.
2. Record the returning IP address in a map.
3. The first IP address seen more than three times is the client's IP address.

In my opinion, the usage of only one `ifconfig.me` service is not reliable:

- the service can be down
- as we discussed earlier, it can be compromised
- it can return an HTML page instead of IP address in the body


infra/scripts/cometbft.toml
```diff
@@ -34,7 +34,7 @@ dependencies = ["cometbft-pull", "docker-network-create"]
 [tasks.cometbft-node-id]
 script = """
 id=`docker exec ${CMT_CONTAINER_NAME} cometbft show-node-id`
-ip=`curl ifconfig.me`
+ip=`./target/release/ip`
```
aakoshh, line 37: Do we really need a Rust crate for this?

We spent a lot of effort now to make sure that the build and the infra scripts can work without having to compile the project. It would be nice to keep it that way.

aakoshh (commented):
The code looks good, but perhaps instead of making it part of the workspace you could try using the RustScript facility of cargo-make to embed it?

https://github.com/sagiegurari/cargo-make#rust-code

infra/tools/ip/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "fendermint_infra_ip"
```
aakoshh, line 2: I don’t think this needs a `fendermint_` prefix, it’s even outside the directories.

infra/scripts/cometbft.toml
dnkolegov: Fixed.

dnkolegov:
> The code looks good, but perhaps instead of making it part of the workspace you could try using the RustScript facility of cargo-make to embed it?
> 
> https://github.com/sagiegurari/cargo-make#rust-code

Updated.

aakoshh (approved):
Cool, thank you for turning it into a Rust script :pray: 

How was your experience writing this? Are you just editing a string in the toml file with no idea if it will even compile until you try? Does it seem reasonable for something relatively small such as this use case?

dnkolegov:
> Cool, thank you for turning it into a Rust script 🙏
> 
> How was your experience writing this? Are you just editing a string in the toml file with no idea if it will even compile until you try? Does it seem reasonable for something relatively small such as this use case?

The main problem is to get output from scripts. It is not possible to do that using env variables directly. You have to use plugins.

> Does it seem reasonable for something relatively small such as this use case?
Yes, it makes sense.

Issue #377, opened 2023-11-01
Top-down finality crashing when facing `null` rounds in Calibration


adlrocha:
Reported by Longfei:

> Subnet will stop syncing from Calibration if there are null blocks or epoch, and there is no way to resync or fix the issue so that subnet can keep liveness. This will block moving fund from Calibration to subnet.

adlrocha:
Pasting it here from a Slack thread:

Will, I found where the null round thing is coming from, and it may have an easy fix :thread:
This is the error that we are seeing when there is a null round: https://github.com/filecoin-project/lotus/blob/7bb1f98ac6f5a6da2cc79afc26d8cd9fe323eb30/node/impl/full/eth.go#L164
And this is where is being returned: https://github.com/filecoin-project/lotus/blob/7bb1f98ac6f5a6da2cc79afc26d8cd9fe323eb30/node/impl/full/eth.go#L331
The TL;DR is that if we request the block for a round without block we get this error back.
Which means that every time we request a the hash for a null round, our call to get_block_hash returns an error ([1](https://github.com/consensus-shipyard/fendermint/blob/35e271d5fcf8a952a218e223bc76cd7f742fd0b4/fendermint/vm/topdown/src/sync.rs#L284), [2](https://github.com/consensus-shipyard/fendermint/blob/35e271d5fcf8a952a218e223bc76cd7f742fd0b4/fendermint/vm/topdown/src/sync.rs#L312)), which means that the parent syncer will get stuck after a null round because it will keep throwing errors and it won't be able to advance to the next height.
If we assume that we are waiting for things to be final to propagate them to the subnet (which is not currently the case, but  it will be in the future, especially when Granite is shipped, and in any case we have the same problem with any other block null or not). A null round will never have a block, which means that we can advance to the next height without proposing any proof of finality in the subnet (null rounds do not execute or commit any messages).
So TL;DR of the fix, if we see a null round, no proof of finality is committed and we move to the next height (//cc 
[@Akosh](https://filecoinproject.slack.com/team/U03KMQNT8MQ)
 just in case there are any blind spots on my reasoning).

PR #386, opened 2023-11-03
fix null round

cryptoAtwill:
Closes https://github.com/consensus-shipyard/fendermint/issues/377

Skip null round when syncing with parent.


fendermint/vm/topdown/src/sync.rs
```diff
@@ -22,6 +22,15 @@ const MAX_PARENT_VIEW_BLOCK_GAP: BlockHeight = 100;
 /// When polling parent view, if the number of top down messages exceeds this limit,
 /// the polling will stop for this iteration and commit the result to cache.
 const TOPDOWN_MSG_LEN_THRESHOLD: usize = 500;
+/// The null round error message
+const NULL_ROUND_ERR_MSG: &str = "(code: 1, message: requested epoch was a null round, data: None)";
```
aakoshh, line 26: I wouldn't make it this specific. What if they add some data or change the code? As long as `"requested epoch was a null round"` is present I don't think the other details such as tuple formatting are needed.

fendermint/vm/topdown/src/sync.rs
```diff
@@ -231,6 +240,29 @@ async fn sync_with_parent<T: ParentFinalityStateQuery + Send + Sync + 'static>(
     Ok(())
 }
 
+fn error_handling(
```
adlrocha, line 243: `error_handling` is too general and it may lead to confusion, right? 
```suggestion
fn handle_parent_view_error(
```

adlrocha (commented):
I think this could work, I will test it and let you know how it goes :)

fendermint/vm/topdown/src/sync.rs
adlrocha: Agree, let's have a regexp. An example of how things can go south if we use matching strings: 
* This is what returns the error: https://github.com/filecoin-project/lotus/blob/master/node/impl/full/eth_utils.go#L28
* But some function may add additional context info at the beginning of the error :face_exhaling: https://github.com/filecoin-project/lotus/blob/e8455567fd6a0d3e590f0f2b695b765526637954/node/impl/full/eth.go#L701-L703

It is not the case for the call we are using, but that way we are "safer" in the future.

fendermint/vm/topdown/src/sync.rs
```diff
@@ -267,23 +299,15 @@ async fn get_new_parent_views(
...
+) -> Result<GetParentViewPayload, Error> {
     let mut block_height_to_update = vec![];
     let mut total_top_down_msgs = 0;
 
     for h in start_height..=end_height {
         let block_hash_res = parent_proxy
             .get_block_hash(h)
             .await
-            .map_err(|e| Error::CannotQueryParent(e.to_string()))?;
+            .map_err(|e| Error::CannotQueryParent(e.to_string(), h))?;
```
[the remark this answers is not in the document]
adlrocha: @cryptoAtwill, after testing I can confirm that this fix is needed. Returning an empty vector makes the parent syncer to keep trying to propose the null round and while the syncer makes progress, it ends up getting stuck

cryptoAtwill:
@aakoshh @adlrocha I updated the null round fix, I think the current way of aborting the whole height is too `raw` and `brutal`. Instead of abort all the queries at height h, fixing at the `ipc-agent` proxy level is better, so that we can handle it at each query level instead. 

I also feel we should add a marker for null round block hash. This way to don't inherent the null block into fendermint contract state. Anything would do, such as `vec![]`, `vec![0; n]` or something like `keccak(vec![])`.

cryptoAtwill:
@aakoshh @adlrocha I updated the null round fix, I think the current way of aborting the whole height is too `raw` and `brutal`. Instead of abort all the queries at height h, fixing at the `ipc-agent` proxy level is better, so that we can handle it at each query level instead. 

I also feel we should add a marker for null round block hash. This way to don't inherent the null block into fendermint contract state. Anything would do, such as `vec![]`, `vec![0; n]` or something like `keccak(vec![])`.

aakoshh:
> I also feel we should add a marker for null round block hash. This way to don't inherent the null block into fendermint contract state. Anything would do, such as vec![], vec![0; n] or something like keccak(vec![]).

Where would this be used? Is it the provider API?

I wouldn't try hashing empty data to stand in for the hash of a height where there are no block in Filecoin if it can be avoided. Unless there is a type that you can't control and it demands that you give the hash of _something_, ie. if it's an internal thing for the syncer in Rust, then I think what you mentioned about using `Option` or even an enum such as [tendermint::Hash](https://docs.rs/tendermint/latest/tendermint/hash/enum.Hash.html) is better. Otherwise you will end up comparing to this "empty hash" all the time before trying to do anything. Letting the compiler help you out is better IMO.

aakoshh:
If you _must_ use an empty hash, then it might as well be all zeroes, to make it even easier to identify at glance. It's not like the keccak256 of an empty vector has any meaning; it can't be used for and it doesn't express uniqueness, but it's a source of errors in how you encode an empty vector.

cryptoAtwill:
@aakoshh sorry, my bad. I'm not using that anymore, it's from outdated comment. Now it's just skipping that block for proposal.

fendermint/vm/topdown/src/sync.rs
```diff
@@ -242,106 +239,217 @@ async fn reset_cache<T: ParentFinalityStateQuery + Send + Sync + 'static>(
...
+            //
+            // Currently the last recorded block is N1, so the next block to sync in parent is B2.
+            // The response from getting block hash at height B2 from fvm eth apis would return:
+            //
+            // Block height: B2, Block hash: hash(B2), Parent block hash: hash(B1)
+            //
+            // F0  B0  B1  N0  N1  B2
+            //     B0' N0' N1' N2' B2  B3  B4       <====== reorged chain case 1
+            //     B0  B1  B2' B3' B2               <====== reorged chain case 2
+            //     B0  B1  N0' B1' B2               <====== reorged chain case 3
```
aakoshh, line 289: It's not entire clear what's happening here. Are these observations on the syncer side, say, we have seen B1 followed by B2' then B3' then a switch back to B2 again?

fendermint/vm/topdown/src/sync.rs
```diff
@@ -242,106 +239,217 @@ async fn reset_cache<T: ParentFinalityStateQuery + Send + Sync + 'static>(
...
+            //
+            // F0  B0  B1  N0  N1  B2  B3  B4
+            //
+            // where F0 represents the last committed finality, B* represents non-null blocks and
+            // N* represents null blocks.
+            //
+            // Currently the last recorded block is N1, so the next block to sync in parent is B2.
+            // The response from getting block hash at height B2 from fvm eth apis would return:
+            //
+            // Block height: B2, Block hash: hash(B2), Parent block hash: hash(B1)
```
aakoshh, line 284: Thanks you for including these examples!

I think I misunderstood what a null block was, I thought it would be something like this:
```
B0 B1 N2 N3 B4 B5
```
That is, at height 2 and 3 there wouldn't be any blocks at all, and without a reorg that would not change. I was curious what the API would return for `getBlockByHeight(2)` and `getBlockByHeight(4)` in this case. Are you saying that it will actually return `B4` in my example at height 2?

cryptoAtwill: If you `getBlockByHeight(2)`, it would panic and error is null round sth. But if you `getBlockByHeight(4)`, it returns the block hash at height 4, but the parent block hash is actually `B1`'s block hash.

fendermint/vm/topdown/src/sync.rs
aakoshh: Okay so I got confused by the notation here. Above in `F0  B0  B1  N0  N1  B2  B3  B4` it looks like we use the `B*` numbers as identities for existing blocks, so the ID of the block at height 7 is 3. But then here we see `B3'` and `B1'` which I usually interpret as "a block at height 1 on a fork" to contrast it with `B1`. 

Perhaps it would help to draw down a chain with these forks and then mark them with identities as well. I think the above scenarios would translate to this:
```
F0 - B0 - B1 - N - N - B2 - B3 - B4
  |         |
  |         |- B8
  |         |
  |         |- B6 - B7              
  |
  |- B5 - N - N - N 
```

with the following translations, with `*` marking reorgs
```
F0  B0  B1  N0  N1  B2          => F0 B0 B1 N  N  B2
    B0' N0' N1' N2' B2  B3  B4  => F0 B5 N  N  N *B2 B3 B4
    B0  B1  B2' B3' B2          => F0 B0 B1*B6 B7*B2 
    B0  B1  N0' B1' B2          => F0 B0 B1 N  B8*B2      
```

adlrocha (commented):
Let's have a few tests with null rounds on the cache to be "safer" in case of a regression, and let's fix CI. I've tested it and it works fine for me, so let's fix those things and try to get them merged before we cut the release :pray:

adlrocha (approved):

PR #389, opened 2023-11-03
FM-357: Notify the snapshot manager about new commits

### LINKED ISSUE #357: Notify snapshot manager about committed blocks
Modify `App::commit` so that it notifies the snapshot manager about every new committed block event, including:
* `BlockHeight`, so it can decide whether to snapshot it
* `FvmStateParams` which is not in the `BlockStore` but has to go in the snapshot

The notification is immediate (just writes to STM) and doesn't block anything. We do it in `commit` so we know that it's already persisted and the notified component will be able to read it.

aakoshh:
Closes #357 

Changes the `App` to notify the snapshotter about newly committed blocks. The snapshotter is a new optional constructor dependency of the `App` currently unset, so we don't end up exporting data with nothing to clean it up yet.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2825,6 +2825,7 @@ dependencies = [
  "fendermint_vm_interpreter",
  "fendermint_vm_message",
  "fendermint_vm_resolver",
+ "fendermint_vm_snapshot",
  "fendermint_vm_topdown",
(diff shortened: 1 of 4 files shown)


adlrocha (approved):
LGTM :pray:

PR #391, opened 2023-11-03
FIX: pass settings to run bottom-up checkpoints in validator infra

adlrocha:
While testing the end-to-end I realized that no bottom-up checkpoints were being signed in the child subnet. It ended up being because I wasn't passing the `ValidatorContext` settings required for the validator to be able to sign and commit checkpoint signatures. This PR passes the settings to the validator infra scripts as env variables.


aakoshh (approved):
:+1:

PR #395, opened 2023-11-03
FM-377: Handle one or many JSON-RPC requests

aakoshh:
Fixes [#377](https://github.com/consensus-shipyard/ipc/issues/377)

Changes the HTTP server to accept a single JSON object or an array of objects as request, expecting that the latter are a batch of requests. The `jsonrpc_v2` server implementation actually handles this, I just didn't know it was part of the Ethereum API spec. 

I did not change the websocket implementation, that still expects results to arrive as single messages.

### DIFF
--- a/fendermint/eth/api/src/handlers/http.rs
+++ b/fendermint/eth/api/src/handlers/http.rs
@@ -6,60 +6,97 @@
 
 use axum::http::{HeaderMap, StatusCode};
 use axum::response::IntoResponse;
-use jsonrpc_v2::{Id, RequestObject as JsonRpcRequestObject};
+use jsonrpc_v2::{RequestObject, ResponseObjects};
+use serde::Deserialize;
 
-use crate::{apis, AppState, JsonRpcServer};
+use crate::{apis, AppState};
+
+type ResponseHeaders = [(&'static str, &'static str); 1];
+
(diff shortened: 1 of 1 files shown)


jsoares:
> I did not change the websocket implementation, that still expects results to arrive as single messages.

Can this become a problem? I believe Blockscout will also use WS:

> The WebSockets RPC endpoint used to subscribe to the newHeads subscription alerting the indexer to fetch new blocks.

But not sure how those requests are crafted; this seems to be handled by the EthereumJSONRPC client. Anyhow, we can solve this one, and I'll report back if we face other issues -- I assume we don't have block traces implemented, so it will probably crap out there.

raulk:
Shouldn't be a problem for subscriptions -- they won't use batched requests. In fact, they make little sense with WS since the client can pipeline requests.

@aakoshh I think we should investigate making processing parallel with a worker pool. Could we look at what other Eth clients do?

aakoshh:
@raulk thanks, good idea. Actually the library processes them in parallel, so we are covered: https://github.com/kardeiz/jsonrpc-v2/blob/16ec062ec3e9f83f95458bbb5e5bc6a599adbf75/src/lib.rs#L958-L964

jsoares (approved):
Tested, fixes the issue.

PR #397, opened 2023-11-06
Check if syncing with peer

cryptoAtwill:
Stops parent finality syncing with parent if the current node is syncing with peers.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3232,6 +3232,7 @@ dependencies = [
  "num-traits",
  "serde",
  "serde_json",
+ "tendermint-rpc",
  "thiserror",
  "tokio",
  "tracing",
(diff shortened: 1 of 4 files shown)


aakoshh (approved):
Nice. In practice I haven't found evidence that `catching_up` will go back from `false` to `true` unless CometBFT is restarted, but I also don't think there should be any harm in asking.

adlrocha (approved):

aakoshh:
Actually there is a hint in the [CometBFT config](https://docs.cometbft.com/v0.37/core/configuration):

```
# If this node is many blocks behind the tip of the chain, BlockSync
# allows them to catchup quickly by downloading blocks in parallel
# and verifying their commits
#
# Deprecated: this key will be removed and BlockSync will be enabled 
# unconditionally in the next major release.
block_sync = true
```

Issue #399, opened 2023-11-06
Blockscout compat tracking


jsoares:
Blockscout no longer fails the initial fetch if using #395. It can now get a list of blocks and starts fetching them in batches, as well as subscribing to new blocks. This means, of course, new issues -- which seem rather uniform across the different RPC flavours.

1. Timestamps

```
backend                         | 2023-11-06T23:17:32.840 application=explorer [info] tx/per day chart: collect records for txs per day stats
backend                         | 2023-11-06T23:17:32.840 application=explorer [info] tx/per day chart: earliest date 2023-11-05 00:00:00Z
backend                         | 2023-11-06T23:17:32.840 application=explorer [info] tx/per day chart: latest date 2023-11-05 23:59:59Z
backend                         | 2023-11-06T23:17:32.847 application=explorer [info] tx/per day chart: timestamp cannot be converted to min/max blocks, trying to find min/max blocks through a fallback option}
backend                         | 2023-11-06T23:17:32.850 application=explorer [warning] tx/per day chart: failed to get min/max blocks through a fallback option}
```

This is just a warning and doesn't seem to block anything. Also unclear whether it is related to our stack, maybe just normal init? I guess it should be doing something based on genesis and round period, so might even be configuration.

2. no pending_transaction/getPendingTransaction method available (ref: https://github.com/blockscout/blockscout/blob/master/apps/indexer/lib/indexer/fetcher/pending_transaction.ex)

```
backend                         | 2023-11-06T23:19:04.093 application=indexer fetcher=pending_transaction [error] %{code: -32601, message: "Method not found"}
```

This seems medium (pending transaction fetch can be disabled with INDEXER_DISABLE_PENDING_TRANSACTIONS_FETCHER) and happens for most flavours. We may need to implement it.

3. Blocks fail to validate on account of blank nonce

```
[...]

PR #403, opened 2023-11-07
FIX: Set block hash and number in to_eth_block

aakoshh:
Related to https://github.com/consensus-shipyard/fendermint/issues/399#issuecomment-1798758781 in #399 

Sets the `blockHash` and `blockNumber` on the transactions when they are embedded in a block. They are obvious, which is why I thought they don't need to be set in this context, but apparently that is not the case. They are probably optional for the sake of transactions that aren't included in any block.

### DIFF
--- a/fendermint/eth/api/src/conv/from_tm.rs
+++ b/fendermint/eth/api/src/conv/from_tm.rs
@@ -108,6 +108,8 @@ pub fn to_eth_block(
                 .context("failed to convert to eth transaction")?;
 
             tx.transaction_index = Some(et::U64::from(idx));
+            tx.block_hash = Some(et::H256::from_slice(block.header.hash().as_bytes()));
+            tx.block_number = Some(et::U64::from(block.header.height.value()));
 
             transactions.push(tx);
         }


jsoares (commented):
```
backend                         | 2023-11-07T15:26:55.178 application=indexer fetcher=block_catchup first_block_number=128 last_block_number=128 [error] ** (FunctionClauseError) no function clause matching in EthereumJSONRPC.Transaction.elixir_to_params/1
backend                         |     (ethereum_jsonrpc 5.3.1) lib/ethereum_jsonrpc/transaction.ex:173: EthereumJSONRPC.Transaction.elixir_to_params(%{nil => nil, "blockHash" => "0xe38c02e95fcc5099cd286c75c3cd3832d68ae39ecb8234b190b426583ffd10c7", "blockNumber" => 128, "chainId" => 3522868364964899, "from" => "0x8648f163750e3efed4f77d88a96d2685b694628e", "gas" => 1622716, "hash" => "0x0691501f73795704db45abe15d94168cb8379a2b3ddc24be91c5f099574b51cd", "input" => "0x40", "maxFeePerGas" => 1500001200, "maxPriorityFeePerGas" => 1500000000, "nonce" => 0, "r" => 110428157458856392277804628568161995515635561943346854452140107098481567656000, "s" => 33483381959205440497496170019303477864173447613849671740708889309703597879261, "to" => "0x8648f163750e3efed4f77d88a96d2685b694628e", "transactionIndex" => 0, "v" => 1, "value" => 10000000000000000000})
```

The params are now populated but still same issue. I think we need to match arity in addition to concrete params, except if they're marked as _. Unclear what's the best path forward here, but maybe check with Blockscout @maciejwitowski ?

maciejwitowski (approved):

PR #404, opened 2023-11-07
FM-366: Export snapshot to chunked CAR files

### LINKED ISSUE #366: Export snapshot
When the manager is notified about a new block at a snapshottable height, and CometBFT is no longer syncing, and it has finished any previous snapshots, it should write a new one. 

Consider the following:
* If the state hasn't changed, can we reuse the previous CAR file and only modify the meta-data in it (basically only the timestamp and the block height)?
* We should write to a temporary location first and only move the file the snapshot directory if it's successful
* Should we write some kind of checksum?

aakoshh:
Closes #366 

The idea is to break up the CAR file exported in #354 into files that fit into the ABCI limit of 10MB, which should be possible because the default IPLD block limit in FVM is 1MB. The files should be random access with a 0-based indexing the ABCI expects. 

The scheme being implemented in this PR is that the header of the CAR file goes into the `0.part` file, then we write the contents to `1.part`, `2.part`, starting a new file whenever we reach the limit. 

To try to reuse as much of the `fvm_ipld_car` API this PR implements the `WriteAsync` abstraction that rotates files transparently, instead of iterating over the CAR file contents which would require copying private stuff from `fvm_ipld_car::util` such as `ld_write` and `ld_read`. 

Once this is working, we will have the option to validate chunks on the fly in #364 by requiring that any CID we encounter has already been pointed at by a `Link` in a previous chunk, thus ensuring that what we are loading is at least a DAG.

- [x] File checksum function
- [x] Stream CAR contents
- [x] Write CAR contents with file rotation
- [x] Integration test with builtin actor bundle CAR
- [x] Write CAR level checksum

Deferred for later:
- [ ] Write snapshot manifest
- [ ] Write chunk level checksums
- [ ] Restore CAR file from chunks (the test does it, it's just concatentation, but I'm not sure how the import will work)
- [ ] Validate checksums (there is a separate ticket for validation)

The snapshot layout will look for example like this:
```
~/.fendermint/snapshots
├── snapshot-1000
│   ├── manifest.json
│   ├── parts.sha256
│   └── parts
│       ├── 0.part
│       └── 1.part
├── snapshot-2000
│   ├── manifest.json
│   ├── parts.sha256
│   └── parts
│       ├── 0.part
│       ├── 1.part
│       └── 2.part
...
```


fendermint/vm/snapshot/src/car/mod.rs
```diff
@@ -0,0 +1,128 @@
...
+        .with_context(|| format!("failed to open CAR file: {}", input_file.to_string_lossy()))?;
+
+    let reader: CarReader<_> = CarReader::new_unchecked(file.compat())
+        .await
+        .context("failed to open CAR reader")?;
+
+    // Create a Writer that opens new files when the maximum is reached.
+    let mut writer = ChunkWriter::new(output_dir.into(), max_size, file_name);
+
+    let header = CarHeader::new(reader.header.roots.clone(), reader.header.version);
```
adlrocha, line 45: Does the cid roots of the header currently point to the cid of the different parts? I don't think this is the case, right? We are just taking the whole CAR as a single DAG and then splitting it into parts.

fendermint/vm/snapshot/src/car/chunker.rs
```diff
@@ -0,0 +1,222 @@
...
+            Open { mut out, written } => match out.as_mut().poll_flush(cx) {
+                Poll::Pending => Open { out, written }.pending(),
+                Poll::Ready(Err(e)) => Open { out, written }.err(e),
+                Poll::Ready(Ok(())) => {
+                    // Close the file if either:
+                    // a) we have written the header, or
+                    // b) we exceeded the maximum file size.
+                    // The flush is ensured by `fvm_ipld_car::util::ld_write` called by `CarHeader::write_stream_async` with the header.
+                    // The file is closed here not in `poll_write` so we don't have torn writes where the varint showing the size is split from the data.
+                    let close = this.next_idx == 1 || written >= this.max_size && this.max_size > 0;
```
adlrocha, line 200: This makes sense, but quick sanity-check. The behavior is the following: 
- `write_stream_async` calls `poll_flush`.
- If went over the maximum chunk size, we open a new chunk and `poll_write` there. 
The goal of this is to keep clear boundaries between chunk for when we split the CAR file into chunks, right?

adlrocha (commented):
Man, this was a hard one. Thanks  :raised_hands:  Overall LGTM, and the tests seem to test that the behavior is correct. 

My main question (and something maybe worth clarifying in the comments) is if the CAR header for the snapshot assumes a single DAG with a single root, or in the header there is a root for each part.

adlrocha (approved):

fendermint/vm/snapshot/src/car/mod.rs
aakoshh: No, it doesn't. It is as you describe, we just split the body of the CAR with no regard to forming sub-dags or anything. The only criteria is the size, and that the content should stay with its CID, so we can validate parts as we get them:
1. Check that the CID corresponds to the content
2. Check that the CID is something that was pointed at by an earlier `Link`.

fendermint/vm/snapshot/src/car/chunker.rs
aakoshh: `CarHeader::write_stream_async` [calls](https://docs.rs/fvm_ipld_car/latest/src/fvm_ipld_car/lib.rs.html#43-47) `ld_write` with the header, then each CID+content pair; that in turn [writes](https://docs.rs/fvm_ipld_car/latest/src/fvm_ipld_car/util.rs.html#52-54) the data prefixed by its length, then flushes. 

The goal is to split after such content has been written, so it's like `<len0+header>|<len1+cid1+data1><len2+cid2+data2>|<len3+cid3+data3>...|...` where `|` is a split, we don't do split like `<len1+cid1|data1>`

aakoshh:
> My main question (and something maybe worth clarifying in the comments) is if the CAR header for the snapshot assumes a single DAG with a single root, or in the header there is a root for each part.

As far as I understand code in this PR doesn't assume anything. If you look at how @cryptoAtwill implemented the [snapshot export](https://github.com/consensus-shipyard/fendermint/blob/main/fendermint/vm/interpreter/src/fvm/state/snapshot.rs#L89-L98) itself you'll see that there will be a single root in the header, however it will be from a record that actually doesn't exist in the blockstore under normal circumstances, an artificial metadata piece pointing at the real state root, plus containing some other stuff. Arguably this same stuff also lives in the `manifest.json` file too, and I'm not sure if I should try to avoid importing it. Eventually it's going to be garbage collected anyway. 

I suppose the real state root could also be added to the roots, even if it's transitively referenced by the metadata. 

As far as this code is concerned we can have one root or multiple roots, it doesn't matter, they are all going to be in `0.part`, and validation should work too as long as they are a DAG.

PR #405, opened 2023-11-07
FIX: several fixes to cometbft config, bootstrap and validators in infra scripts

adlrocha:
This PR fixes several issues with the infra scripts when deploying in public networks: 
- We need to set several new fields in the cometbft config like `addr_book` or `external_address` so validators and bootstrap are discoverable. 
- `seed_mode` shouldn't be enabled in validators or full-nodes, only in bootstraps. 
- `set-seed` was not working properly and it was fixes. 
- Bootstraps need to use the genesis file of the subnet so it uses the same chainID of the validators. Without this any connection between them will be rejected. We can't run bootstraps without an ABCI application, which means that bootstraps need to be effectively full nodes that run Fendermint with `seed_mode` enabled.
- Include `restart` command for IPC nodes. 
- Return the node's CometBFT ID in the node report.
- Clean redundant code.


infra/scripts/cometbft.toml
```diff
@@ -171,4 +171,34 @@ env = { "CONTAINER_NAME" = "${CMT_CONTAINER_NAME}" }
...
+
+[tasks.addr-book-enable]
+script="""
+sed -i'' -e "s|addr_book_strict = false|addr_book_strict = true|" ${CMT_DIR}/config/config.toml
+"""
+
+[tasks.set-seeds]
+script = """
+sed -i'' -e 's|^seeds = ""$|seeds = "{{PLACEHOLDER}}"|g' ${CMT_DIR}/config/config.toml
+sed -i "s/{{PLACEHOLDER}}/$BOOTSTRAPS/g" ${CMT_DIR}/config/config.toml
```
dnkolegov, line 189: This sed script does not work on MAC. Please use the format as above

```suggestion
sed -i'' -e "s/{{PLACEHOLDER}}/$BOOTSTRAPS/g" ${CMT_DIR}/config/config.toml
```

infra/scripts/cometbft.toml
```diff
@@ -171,4 +171,34 @@ env = { "CONTAINER_NAME" = "${CMT_CONTAINER_NAME}" }
...
+"""
+
+[tasks.set-external-addr]
+script = """
+sed -i'' -e 's|^external_address = ""$|external_address = "{{PLACEHOLDER}}"|g' ${CMT_DIR}/config/config.toml
+sed -i "s/{{PLACEHOLDER}}/$CMT_EXTERNAL_ADDR/g" ${CMT_DIR}/config/config.toml
```
dnkolegov, line 195:
```suggestion
sed -i'' -e "s/{{PLACEHOLDER}}/$CMT_EXTERNAL_ADDR/g" ${CMT_DIR}/config/config.toml
```

dnkolegov (changes requested):

infra/scripts/cometbft.toml
adlrocha: Did you tested it in Mac? I will double-check if it works in Linux after the change. I had a really hard time to make this one work.

infra/scripts/cometbft.toml
adlrocha: Same as above

adlrocha:
@dnkolegov, I can confirm that your changes work in Linux, let me know if it works in Mac.

dnkolegov:
> @dnkolegov, I can confirm that your changes work in Linux, let me know if it works in Mac.

It works

adlrocha:
@dnkolegov, is it OK if I merge this one? I think I addressed all your comments.

dnkolegov (approved):

PR #409, opened 2023-11-08
FM-407: Insert 0x00..00 for system account

### LINKED ISSUE #407: Create EthAccount at `0x00..00` + fix default sender address
[This PR](https://github.com/consensus-shipyard/fendermint/pull/375) made us default to the system actor address in masked ID form as a sender when no `from` address is provided to `eth_call`. That address is `0xff00..00`. However, clients expect this address to be `0x00..00` (the null Ethereum address). We need to do two things here:

1. Create an EthAccount actor with an f410 address equivalent to `0x00..00`. The actor ID does not matter. See how Lotus does that [here](https://github.com/filecoin-project/lotus/blob/master/chain/gen/genesis/genesis_eth.go).
2. Fix `eth_call` to use that address as the default sender.

aakoshh:
Fixes #407 

Inserts into the `Init` actor an association from the "null" `f410` address `0x00..00` to the `SYSTEM_ACTOR_ID`. This way calls that do not set a `from` will continue to work with implicit semantics FWIW.

### DIFF
--- a/fendermint/eth/api/examples/ethers.rs
+++ b/fendermint/eth/api/examples/ethers.rs
@@ -490,6 +490,8 @@ where
     // Calling with 0 nonce so the node figures out the latest value.
     let mut probe_tx = transfer.clone();
     probe_tx.set_nonce(0);
+    // Calling with 0x00..00 address so we see if it world work for calls by clients that set nothing.
+    probe_tx.set_from(Address::default());
     let probe_height = BlockId::Number(BlockNumber::Number(bn));
 
     request(
--- a/fendermint/vm/actor_interface/src/eam.rs
+++ b/fendermint/vm/actor_interface/src/eam.rs
@@ -30,7 +30,7 @@ pub enum Method {
 
 // TODO: We could re-export `fil_evm_actor_shared::address::EvmAddress`.
 #[derive(
-    serde::Deserialize, serde::Serialize, Clone, Copy, PartialEq, Eq, Hash, PartialOrd, Ord,
+    serde::Deserialize, serde::Serialize, Clone, Copy, PartialEq, Eq, Hash, PartialOrd, Ord, Default,
 )]
 pub struct EthAddress(#[serde(with = "strict_bytes")] pub [u8; 20]);
 
--- a/fendermint/vm/actor_interface/src/init.rs
+++ b/fendermint/vm/actor_interface/src/init.rs
@@ -11,7 +11,7 @@ use fvm_ipld_encoding::tuple::*;
 use fvm_ipld_hamt::Hamt;
 use fvm_shared::{address::Address, ActorID, HAMT_BIT_WIDTH};
 
-use crate::eam::EthAddress;
+use crate::{eam::EthAddress, system};
 
 /// Defines first available ID address after builtin actors
 pub const FIRST_NON_SINGLETON_ADDR: ActorID = 100;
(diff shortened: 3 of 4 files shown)


raulk:
That's interesting. Doing this wouldn't work in Lotus, since I think we validate that the sender of an Eth message is nothing else but an Eth Account (EOA). It might just work here. You don't foresee any complications down the line in merging these types? You mentioned something about doing this to disable gas accounting. We do want gas accounting to run though, since I think some apps may call eth_estimateGas with the null sender.

aakoshh:
I'm not sure but intuitively the masked ID address of the system actor also didn't come with an ethereum account actor attached to it, and it worked in calls. As far as I can see as long as [the address can be turned into an ID](https://github.com/filecoin-project/ref-fvm/blob/e1931b057f85611758f4f580974e4e384ba409fd/fvm/src/executor/default.rs#L363-L366) it should be equivalent. 

The implicit execution still counts the gas and respects the gas limit, it just doesn't check that the sender has enough balance to cover it up front, so it's one less thing to worry about with calls. If you send a call in your own name, then it's an explicit execution and you need to have the funds, however currently there is no miner penalty for including transactions with 0 gas price, so that's another way to circumvent this. But tools often estimate the price and set it, which sometimes means a lot of stars have to align.

aakoshh:
It looks like you are right about there being some validation, the [tests failed](https://github.com/consensus-shipyard/fendermint/actions/runs/6792146312/job/18464946000?pr=409 ) with:
```
2023-11-08T01:23:36.415823Z  INFO ethers: example transfer height=1088 tx_hash=0xb338e8cd3676f2fa6e49b3315199cf6baee9fc1599df03adc69752ede5e7b939
Error: failed to call eth_call: (code: 1, message: pre-validation failed: Send not from valid sender
, data: Some(String("")))
```

fendermint/vm/actor_interface/src/system.rs
```diff
@@ -2,9 +2,23 @@
...
 use fvm_ipld_encoding::tuple::*;
+use fvm_shared::address::Address;
+use lazy_static::lazy_static;
+
+use crate::eam::EthAddress;
 
 define_singleton!(SYSTEM { id: 0, code_id: 1 });
 
+lazy_static! {
+    /// The Ethereum null-address 0x00..00 can also be used to identify the system actor.
```
adlrocha, line 13: @raulk, is this also the case in lotus?

fendermint/vm/interpreter/src/fvm/state/query.rs
```diff
@@ -190,7 +190,7 @@ where
                 msg.gas_limit = fvm_shared::BLOCK_GAS_LIMIT;
             }
 
-            if msg.from == SYSTEM_ACTOR_ADDR {
+            if is_system_addr(&msg.from) {
```
adlrocha, line 193: Maybe a dumb question, but what happens if a user sends an external transaction with the zero address in the `from`? I guess that it will have no signature and it will be rejected, right? I am just wondering if we are opening the door here to users being able to send free transactions that are executed implicitly by just setting `from = 0x00..`

fendermint/vm/actor_interface/src/system.rs
aakoshh: Based on [this](https://github.com/filecoin-project/lotus/blob/fbafe1c507f362c82d2cd5540b6c5d580dae8e74/chain/gen/genesis/genesis_eth.go#L25-L29) I think they just create Ethereum account actors with these addresses, they aren't the same as the system address. It is only me who thought since we have this notion of sending these calls as the system actor, and that's actually what Lotus does too, then why not make it official?

adlrocha (approved):

PR #410, opened 2023-11-08
FM-406: Return empty bytes from eth_call if it's a deploy

### LINKED ISSUE #406: Failed to deploy a simple contract through Remix IDE
When trying to deploy the simple Solidity contract pasted below through [Remix IDE](https://remix.ethereum.org/) compiled with Solidity 0.8.21, I am confronted with this error: "creation of MyToken errored: Error occured: Transaction has been reverted by the EVM."

I have traced the problem down to the following `eth_call`. In this call, Remix is trying to simulate the contract creation prior to submitting the transaction:

```sh
curl 'http://<ip>:8545/' \
  -H 'Accept: application/json' \
  -H 'Content-Type: application/json' \
  --data-raw '{"id":1460100713,"jsonrpc":"2.0","method":"eth_call","params":[{"from":"0xAF4E1Dfc45b23cBf0311541c0ac4c1E860d0AB9F","data":"0x608060405234801561000f575f80fd5b5060c88061001c5f395ff3fe608060405260043610601b575f3560e01c8063c298557814601f575b5f80fd5b60256039565b60405160309190607b565b60405180910390f35b5f33905090565b5f73ffffffffffffffffffffffffffffffffffffffff82169050919050565b5f6067826040565b9050919050565b607581605f565b82525050565b5f602082019050608c5f830184606e565b9291505056fea26469706673582212206789b2c9c7d635856db9515e4549a48afe20db8fece3fb9a8f0aca623dc9e0a864736f6c63430008150033","value":"0x0","gas":"0xe09d72","type":"0x2","maxPriorityFeePerGas":"0x9502f900","maxFeePerGas":"0x950300d0"},"0x5e8"]}'
{"jsonrpc":"2.0","error":{"code":0,"message":"error decoding data from deliver_tx in query: failed to deserialize bytes returned by FEVM: Serialization error for Cbor protocol: Mismatch { expect_major: 2, byte: 131 }"},"id":1460100713}
```

## Contract

```solidity
contract MyToken {
    function foo() external payable returns (address) {
        return msg.sender;
    }
}
```

aakoshh:
Fixes #406 

The PR changes the `eth_call` handler to first try to decode the return data as an EVM invocation result, and if that fails then try with an EAM contract creation as well. If the latter succeeds it means the call was a deployment, and we return the Ethereum address as bytes.


aakoshh:
I see that Lotus actually returns [empty bytes](https://github.com/filecoin-project/lotus/blob/5a1738c9ba99996c2cdc60a472ad5f149267aaf0/node/impl/full/eth.go#L1091-L1094) so maybe returning the address will upset tools. @raulk do you think Remix insists on empty?

There address returned cannot actually be used for anything like lining up another call targeting that account because the effects of calls aren't persisted, even when they run on `pending` state. But I suppose it's possible to send a call like this, mark down the address, then send it for real which affects the pending state even before it's included in a block, and line up a call to the contract - if the address is deterministic it can work, however convoluted and ill advised this is.

raulk:
An external contract creation transaction doesn't return a payload in Eth. The caller can compute the expected address of the contract deterministically. Returning the deployed address may be useful for manual debugging, but tools, libraries and wallets don't expect that behaviour and can lead to breakage.

raulk:
Actually, let me refine that answer. I think EVM initcode bytecode _can_ return a value, but Solidity constructors can't, AFAIK. So here you'd need to simply propagate the return you get from the FVM.

raulk:
Btw -- I'm not sure trial and error is a good idea. Is there no way to mark the "type" of operation at request time and correlate with expected behaviour at response time?

aakoshh:
I changed it to return empty bytes, like Lotus, based on the same condition of going to the EAM instead of an EVM actor. Also added a test to avoid trial and error, sorry, should have done that in the first place. 

> So here you'd need to simply propagate the return you get from the FVM.

I don't think the tools would be able to do anything with that, it's an IPLD encoded [CreateReturn](https://github.com/filecoin-project/builtin-actors/blob/cd9ac2bb0afcca7a59465e57cee6569e69070d7a/actors/eam/src/lib.rs#L83-L88). I could be wrong but I thought that was FVM specific.

adlrocha (approved):
I am probably not the best person to approve this one, so I will give the final call to @raulk, but the change LGTM :) That being said, I am impressed with the amount of "gotchas" one can find along the way haha

PR #414, opened 2023-11-08
FIX: Set transaction_type to 2

aakoshh:
Sets the `transaction_type` fields on the `ethers::core::type::Transaction` return values in the API to 2, which should signal that `gas_price` will not be set.


maciejwitowski (approved):

jsoares (approved):
It makes sense and tests fine. It doesn't solve all of our problems: to match geth expectations, we should also set gas price in line 171, to something like `gas_price: (Some(to_eth_tokens(&msg.gas_fee_cap)?+to_eth_tokens(&msg.gas_premium)?))` or whatever the appropriate Rust idiom.

PR #416, opened 2023-11-09
FM-360: List snapshots

### LINKED ISSUE #360: List completed snapshots in ABCI
Integrate the snapshot manager with the ABCI facility to list snapshots when remote nodes are looking for them. Only list completed ones.

aakoshh:
Closes #360 

- [x] Snapshot manifest type which we can store along the parts in a `manifest.json`
- [x] Export the manifest
- [x] List snapshots in a directory
- [x] Initialise the existing snapshots and add them to STM state
- [x] Add new snapshot manifests to the STM state
- [x] List snapshots from the STM data structure in the ABCI
- [x] Integration test that tests a snapshot being created and added to STM


fendermint/vm/interpreter/src/fvm/state/snapshot.rs
```diff
@@ -220,8 +227,15 @@ impl<BS: Blockstore> Stream for StateTreeStreamer<BS> {
 
             match this.bs.get(&cid) {
                 Ok(Some(bytes)) => {
-                    let ipld = from_slice::<Ipld>(&bytes).unwrap();
-                    walk_ipld_cids(ipld, &mut this.dfs);
+                    // Not all data in the blockstore is traversable, e.g.
+                    // Wasm bytecode is inserted as IPLD_RAW here: https://github.com/filecoin-project/builtin-actors-bundler/blob/bf6847b2276ee8e4e17f8336f2eb5ab2fce1d853/src/lib.rs#L54C71-L54C79
+                    if cid.codec() == DAG_CBOR {
+                        // XXX: Is it okay to panic?
```
adlrocha, line 233: Are snapshots only read when starting up the peer? If this is the case I think it is OK to panic.

adlrocha (approved):

fendermint/vm/interpreter/src/fvm/state/snapshot.rs
aakoshh: Here you are looking at the code from @cryptoAtwill that [streams the blockstore DAG](https://github.com/consensus-shipyard/fendermint/blob/3cf934c5fe7388814396b373184625938683b45d/fendermint/vm/interpreter/src/fvm/state/snapshot.rs#L215-L218). The stream either returns a pair of `Some` CID and content, or `None` to show that it's finished, but there is no way to signal errors. 

There was a problem with it (I can't remember now how it manifested itself exactly - something about EOF) when I tried to export the data that the genesis put in place, because the Wasm files are aren't actual IPLD data structures, so the `.unwrap()` at line 223 panicked. I think that every encoded by `DAG_CBOR` was placed by ourselves in this part of the code, so it should be okay to panic, it should parse. 

To answer your question: this isn't happening when we start the peer, this is happening at regular intervals when we are exporting the snapshots from the store.

Issue #412, opened 2023-11-08
Users report that running vanilla The Graph with a simple token kills Fendermint (panic?)


raulk:
(More info soon...)

PR #422, opened 2023-11-11
FIX: Pretend there is a block 0

aakoshh:
Creates a `BLOCK_ZERO` which is returned when the client is looking for one at height 0 or when it's looking for the `BLOCK_ZERO_HASH`. 

Testing through `ethers.rs` example.


aakoshh:
Related to https://github.com/consensus-shipyard/fendermint/issues/412

adlrocha (approved):

PR #424, opened 2023-11-13
FIX: Change checkpoint broadcast mode to sync

aakoshh:
Changes the mode of broadcasting checkpoint signatures from `TxCommit` to `TxSync`, that is, instead of waiting for `deliver_tx` it only waits for `check_tx`. 

The reason is that looking at the logs, the `broadcast_tx_commit` of the CometBFT API doesn't work the way I thought it did. I thought that it will wait for the commit (`deliver_tx`) of the transaction, but also return early if `check_tx` fails, since the response has the results for both. Instead what we see is that it times out waiting for delivery:

```console
2023-11-13T10:00:20.556795Z  INFO fendermint_vm_interpreter::fvm::query: query actor state height=40862 pending=true addr="t410frbdnwklaitcjsqe7swjwp5naple6vthq4woyfry" found=true
2023-11-13T10:00:20.557981Z  INFO fendermint_vm_interpreter::fvm::query: query estimate gas height=40862 pending=false to="t410fo6vebmifqq3sqcemaezoip6eineiqhnidng7rxi" from="t410frbdnwklaitcjsqe7swjwp5naple6vthq4woyfry" method_num=3844450837
2023-11-13T10:00:20.617176Z  INFO fendermint_vm_interpreter::fvm::check: check transaction exit_code=0 from="t410frbdnwklaitcjsqe7swjwp5naple6vthq4woyfry" to="t410fo6vebmifqq3sqcemaezoip6eineiqhnidng7rxi" method_num=3844450837 info=""
2023-11-13T10:00:21.398207Z  INFO fendermint_vm_interpreter::fvm::check: check transaction exit_code=33 from="t410frbdnwklaitcjsqe7swjwp5naple6vthq4woyfry" to="t410fo6vebmifqq3sqcemaezoip6eineiqhnidng7rxi" method_num=3844450837 info="message failed with backtrace:\n00: t064 (method 3844450837) -- contract reverted (33)\n01: t064 (method 6) -- out of gas (7)\n"
2023-11-13T10:00:30.618450Z ERROR fendermint_vm_interpreter::fvm::exec: error broadcasting checkpoint signature error=failed to broadcast checkpoint signature
Caused by:
  0: failed to broadcast signature
  1: failed to invoke contract
  2: response error
    
    Caused by:
      Internal error: timed out waiting for tx to be included in a block (code: -32603)
    
    Location:
      /root/.cargo/registry/src/index.crates.io-6f17d22bba15001f/flex-error-0.4.4/src/tracer_impl/eyre.rs:10:9
  3: response error
  4: Internal error: timed out waiting for tx to be included in a block (code: -32603)
Stack backtrace:
  0: <fendermint_rpc::client::BoundFendermintClient<C> as fendermint_rpc::tx::TxClient>::perform::{{closure}}
  1: fendermint_rpc::tx::TxClient::fevm_invoke::{{closure}}
  2: fendermint_vm_interpreter::fvm::checkpoint::broadcast_incomplete_signatures::{{closure}}
  3: fendermint_vm_interpreter::fvm::exec::<impl fendermint_vm_interpreter::ExecInterpreter for fendermint_vm_interpreter::fvm::FvmMessageInterpreter<DB,TC>>::end::{{closure}}::{{closure}}
  4: tokio::runtime::task::core::Core<T,S>::poll
  5: tokio::runtime::task::harness::Harness<T,S>::poll
  6: tokio::runtime::scheduler::multi_thread::worker::Context::run_task
  7: tokio::runtime::scheduler::multi_thread::worker::run
  8: tokio::runtime::task::core::Core<T,S>::poll
  9: tokio::runtime::task::harness::Harness<T,S>::poll
 10: std::sys_common::backtrace::__rust_begin_short_backtrace
 11: core::ops::function::FnOnce::call_once{{vtable.shim}}
 12: <alloc::boxed::Box<F,A> as core::ops::function::FnOnce<Args>>::call_once
       at /rustc/cc66ad468955717ab92600c770da8c1601a4ff33/library/alloc/src/boxed.rs:2007:9
   <alloc::boxed::Box<F,A> as core::ops::function::FnOnce<Args>>::call_once
       at /rustc/cc66ad468955717ab92600c770da8c1601a4ff33/library/alloc/src/boxed.rs:2007:9
   std::sys::unix::thread::thread::new::thread_start
       at /rustc/cc66ad468955717ab92600c770da8c1601a4ff33/library/std/src/sys/unix/thread.rs:108:17
 13: <unknown>
 14: clone height=40860
```

The problem is that this prevented the `retry!` mechanism I put in place because the [error case analysis](https://filecoinproject.slack.com/archives/C04JR5R1UL8/p1699876861338349?thread_ts=1699874787.947719&cid=C04JR5R1UL8) is circumvented by the unexpected error. We can't say what to do with a timeout so the best it can do is retry after the next checkpoint, if the checkpoint is still pending. The retry only works if there is an exit code, and a timeout is not it.

What I thought was happening was that a validator tries to submit their signature but because they are the first one they get a gas estimation that is below what it will be once it actually hits the chain and it's no longer alone in the epoch. Or it might just under estimate for some reason. 

Maybe I'll put it some over estimation on top of the estimation.

### Did it crash?

No, Fendermint didn't crash, the above is still just an error log. It just failed to send a checkpoint signature in a spawned background task, which is annoying. It will retry next time, but the estimates can keep being off the mark.

### What does CometBFT say?

Looking at the logs of https://docs.cometbft.com/v0.37/rpc/#/Tx/broadcast_tx_commit making this change is the right thing to do:

> IMPORTANT: use only for testing and development. In production, use BroadcastTxSync or BroadcastTxAsync. You can subscribe for the transaction result using JSONRPC via a websocket. 

However, I still don't understand the error behaviour:

> CONTRACT: only returns error if mempool.CheckTx() errs or if we timeout waiting for tx to commit.
> If CheckTx or DeliverTx fail, no error will be returned, but the returned result will contain a non-OK ABCI code.

It should not have returned an error, but a regular response with codes to look at :shrug:

### DIFF
--- a/fendermint/app/config/default.toml
+++ b/fendermint/app/config/default.toml
@@ -42,6 +42,8 @@ state_hist_size = 0
 max_retries = 5
 # Tome to wait between retries, in seconds. It should roughly correspond to the block interval.
 retry_delay = 2
+# Any over-estimation to apply on top of the estimate returned by the API.
+gas_overestimation_rate = 2
 
 # FVM configuration
 [fvm]
--- a/fendermint/vm/interpreter/src/fvm/broadcast.rs
+++ b/fendermint/vm/interpreter/src/fvm/broadcast.rs
@@ -56,6 +56,7 @@ pub struct Broadcaster<C> {
     addr: Address,
     gas_fee_cap: TokenAmount,
     gas_premium: TokenAmount,
+    gas_overestimation_rate: f64,
     max_retries: u8,
     retry_delay: Duration,
 }
@@ -70,6 +71,7 @@ where
         secret_key: SecretKey,
         gas_fee_cap: TokenAmount,
         gas_premium: TokenAmount,
+        gas_overestimation_rate: f64,
     ) -> Self {
         let client = FendermintClient::new(client);
         Self {
@@ -78,6 +80,7 @@ where
             addr,
             gas_fee_cap,
             gas_premium,
+            gas_overestimation_rate,
             max_retries: 0,
             // Set the retry delay to rougly the block creation time.
             retry_delay: Duration::from_secs(1),
@@ -149,7 +152,8 @@ where
                 .context("failed to estimate gas")?;
 
             if gas_estimate.value.exit_code.is_success() {
-                gas_params.gas_limit = gas_estimate.value.gas_limit;
+                gas_params.gas_limit =
+                    (gas_estimate.value.gas_limit as f64 * self.gas_overestimation_rate) as u64;
             } else {
                 bail!(
                     "failed to estimate gas: {} - {}",
--- a/fendermint/vm/interpreter/src/fvm/broadcast.rs
+++ b/fendermint/vm/interpreter/src/fvm/broadcast.rs
@@ -13,7 +13,7 @@ use tendermint_rpc::Client;
 use fendermint_crypto::SecretKey;
 use fendermint_rpc::message::GasParams;
 use fendermint_rpc::query::QueryClient;
-use fendermint_rpc::tx::{CallClient, TxClient, TxCommit};
+use fendermint_rpc::tx::{CallClient, TxClient, TxSync};
 use fendermint_rpc::{client::FendermintClient, message::MessageFactory};
 use fendermint_vm_message::query::FvmQueryHeight;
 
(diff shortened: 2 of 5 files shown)


fendermint/app/config/default.toml
```diff
@@ -42,6 +42,8 @@ state_hist_size = 0
 max_retries = 5
 # Tome to wait between retries, in seconds. It should roughly correspond to the block interval.
 retry_delay = 2
+# Any over-estimation to apply on top of the estimate returned by the API.
+gas_overestimation_rate = 2
```
adlrocha, line 46: :laughing: over-estimating the over-estimation

fendermint/vm/interpreter/src/fvm/broadcast.rs
```diff
@@ -96,12 +101,19 @@ where
         self.retry_delay
     }
 
+    /// Send a transaction to the chain and return is hash.
+    ///
+    /// It currently doesn't wait for the execution only, only that it has successfully been added to the mempool,
```
adlrocha, line 106:
```suggestion
    /// It currently doesn't wait for the execution, only that it has successfully been added to the mempool,
```

adlrocha (approved):
Let's merge and see if it solves some of the issues we see :crossed_fingers:

PR #428, opened 2023-11-14
skip null round retry

cryptoAtwill:
Skip retry if the height to query is a null round. This way we can advance the block height in fendermint faster.


fendermint/vm/topdown/src/finality/fetch.rs
```diff
@@ -34,6 +34,12 @@ macro_rules! retry {
                 tracing::warn!(
                     "cannot query ipc parent_client due to: {e}, retires: {retries}, wait: {wait:?}"
                 );
+
+                // there is no point in retrying if the current block is null round
+                if is_null_round_error(&e) {
```
aakoshh, line 39: Should this be moved above the warn, so it doesn't suggest there will be retries?

cryptoAtwill: I shifted the logging

aakoshh (approved):
Cheers :pray:

adlrocha (approved):
@cryptoAtwill, do you mind fixing the CI before merging? Thanks!

PR #430, opened 2023-11-14
Fix Fendermint folder

ArseniiPetrovich:
When you are starting validator through infra scripts - it will put cometbft data files into ~/.ipc/${SUBNET_ID}/${NODE_NAME}/cometbft, but Fendermint data will go into ~/.ipc/${SUBNET_ID}/fendermint for some reason, making it impossible to run multiple validators of the same subnet on 1 machine. This PR is intended to fix this behavior.


adlrocha (approved):
Really good catch, thank you!

PR #432, opened 2023-11-15
Tracing logs

cryptoAtwill:
Add more logging outputs options so that we still have default info log to console, but also have debug logs with rotation written to file.


docker/runner.Dockerfile
```diff
@@ -28,3 +28,7 @@ COPY docker/.artifacts/bundle.car $FM_HOME_DIR/bundle.car
 COPY docker/.artifacts/contracts $FM_HOME_DIR/contracts
 COPY --from=builder /app/fendermint/app/config $FM_HOME_DIR/config
 COPY --from=builder /app/output/bin/fendermint /usr/local/bin/fendermint
+
+RUN useradd ipc && chown -R ipc $FM_HOME_DIR
+
+USER ipc
```
adlrocha, line 34: Why do we need an `ipc` user instead of `root` here? Also, I don't think it is the case but I am wondering if this may break the execution of the `fendermint` container considering the directory structure that we use. Have you tried it?

fendermint/vm/topdown/src/sync.rs
```diff
@@ -340,7 +340,10 @@ async fn last_recorded_data(
             match atomically(|| provider.first_non_null_parent_hash(height)).await {
                 None => unreachable!("should have last committed finality at this point"),
                 Some(hash) => {
-                    tracing::info!("First non null parent hash: {hash:02x?} at height: {height}");
+                    tracing::info!(
```
adlrocha, line 343: Can we add also an info trace when we commit the finality? That way we know the last non-null parent and what is being committed in the chain as final.

fendermint/app/src/main.rs
```diff
@@ -14,14 +23,7 @@ async fn main() {
...
     if let Some(level) = opts.tracing_level() {
-        // Writing to stderr so if we have output like JSON then we can pipe it to something else.
-        let subscriber = FmtSubscriber::builder()
-            .with_max_level(level)
-            .with_writer(std::io::stderr)
-            .finish();
-
-        tracing::subscriber::set_global_default(subscriber)
-            .expect("setting default subscriber failed");
+        create_log(level, &opts.home_dir).expect("cannot create logging");
```
adlrocha, line 26: I think we should add a setting to configure the output path, and if we want the log to be output on std err/std output, or to a file.

fendermint/app/src/main.rs
```diff
@@ -14,14 +23,7 @@ async fn main() {
 
     // Log events to stdout.
```
adlrocha, line 24: Is this comment still relevant?

fendermint/app/src/main.rs
```diff
@@ -30,6 +32,59 @@ async fn main() {
...
+            maximum_file_size: "10mb".to_string(),
+            // we keep the last 5 log files, older files will be deleted
+            max_size_roll_backups: 5,
+            pattern: None,
+        },
+    };
+
+    // debug level logger
+    let topdown_logger = config::Logger {
+        level: config::LevelFilter::DEBUG,
```
adlrocha, line 56: Should we make this configurable too? Now we want this in debug mode for debugging purpose, but we may want to keep the same level for all in the future.

cryptoAtwill: I would say we just keep this top down logs just in case for something unexpected. The logs will be 50mb in total, it's not taking a lot of space. But we always have some logs to refer back to.

fendermint/app/options/src/lib.rs
```diff
@@ -75,6 +75,10 @@ pub struct Options {
     )]
     pub home_dir: PathBuf,
 
+    /// Set a custom directory for ipc log files.
+    #[arg(short = 'd', long, default_value = "/tmp/ipc/logs", env = "FM_LOG_DIR")]
```
aakoshh, line 79: Why not put these relative to the home directory into a `logs` sub-directory, e.g. `~/.fendermint/logs`?

Alternatively `/var/log` is the standard place.

I'd also avoid the `ipc` name, stick to `fendermint`.

fendermint/app/src/main.rs
```diff
@@ -30,6 +32,58 @@ async fn main() {
...
+        level: config::LevelFilter::from_str(level.as_str())?,
+        appenders: literally::hset! { AppenderId::from("console") },
+        format: Format::default(),
+    };
+
+    let config = Config {
+        default,
+        loggers: literally::hmap! {
+            Target::from("fendermint_vm_topdown") => topdown_logger.clone(),
+            Target::from("fendermint_vm_interpreter::chain") => topdown_logger,
```
aakoshh, line 71: Nice that we can specify the sources! Buty why is it restricted to topdown, why not everything under the fendermint packages? At least we wouldn't get the debug logs from 3rd party libraries, but see everything we produce, and topdown might not be the only interesting thing to look at.

fendermint/app/src/main.rs
```diff
@@ -30,6 +32,58 @@ async fn main() {
...
+        },
+        appenders: literally::hmap! {
+            AppenderId::from("console") => console_appender,
+            AppenderId::from("topdown") => topdown_appender
```
aakoshh, line 74: Note that this is in `main.rs` which is also the entry point for the `ethapi` as well as the CLI. Those won't have these emitters, but I can imagine that they would overwrite logs of each other when executed on the host directly, without a container. 

Maybe it would be better if the log directory wasn't specified by default, so that the operator has to explicitly set it, which would make any clashes obvious.

fendermint/vm/interpreter/src/chain.rs
```diff
@@ -250,12 +255,23 @@ where
                     )
                     .await
                     .context("failed to commit finality")?;
+                    tracing::debug!(
+                        "chain interpreter committed topdown finality, prev_height: {}, prev_finality: {})",
+                        prev_height,
+                        prev_finality.as_ref().map(|f| format!("{f}")).unwrap_or_else(|| String::from("None"))
+                    );
```
aakoshh, line 262: I do recommend structure logging instead of string interpolation.

fendermint/vm/interpreter/src/chain.rs
```diff
@@ -250,12 +255,23 @@ where
...
                     let validator_changes = provider
                         .validator_changes_from(prev_height + 1, finality.height)
                         .await
                         .context("failed to fetch validator changes")?;
+                    tracing::debug!(
+                        "chain interpreter received total validator changes msgs: {} from {} to {}",
+                        validator_changes.len(),
+                        prev_height + 1,
+                        finality.height
+                    );
```
aakoshh, line 274:
```suggestion
                    tracing::debug!(
                        from = prev_height + 1, 
                        to = finality.height,
                        msgs = validator_changes.len(),
                        "chain interpreter received total validator changes"
                    );
```

fendermint/vm/topdown/src/finality/null.rs
```diff
@@ -188,11 +188,11 @@ impl FinalityWithNull {
     ) -> StmResult<(), Error> {
         if !top_down_msgs.is_empty() {
             // make sure incoming top down messages are ordered by nonce sequentially
-            tracing::debug!("top down messages: {top_down_msgs:#?}");
+            tracing::debug!("top down messages: {top_down_msgs:?}");
```
aakoshh, line 191:
```suggestion
            tracing::debug!(?top_down_msgs);
```

fendermint/vm/topdown/src/sync.rs
aakoshh: Structured logging again would help keep logs consistent; for example if we use `block_height` consistently then a log aggregator would turn that into a searchable field and you can see everything that happened in a block.

fendermint/vm/topdown/src/sync.rs
```diff
@@ -340,7 +340,10 @@ async fn last_recorded_data(
...
+                        "First non null parent hash: {} at height: {height}",
+                        hex::encode(&hash)
+                    );
```
aakoshh, line 346:
```suggestion
                    tracing::info!(
                        block_height = height, 
                        parent_hash = hex::encode(&hash), 
                        "First non null parent",
                    );
```

docker/runner.Dockerfile
aakoshh: I never quite understood how to use users inside containers properly. I made images where I used `add-user` to create users with specific IDs and groups, but unless those IDs aligned with what exists on the host, I still had to end up setting the `-u` option to make sure I can remove files created into mounted volumes without `sudo`. Since this is a container, I never quite got why we shouldn't use `root` :shrug: I'm sure I'm just ignorant, but if it works...

fendermint/app/src/main.rs
aakoshh: I think it's right to do both. I had a [similar thing in my Scala programs](https://github.com/aakoshh/metronome/blob/e5b657b0ac68fe3bf37ee0c11fb8f6f55ea0e7ec/metronome/examples/resources/logback.xml#L7-L40) except I was able to configure them through XML, so we can enable the debug level on a package-by-package basis (third parties as well) without recompiling.

fendermint/app/src/main.rs
cryptoAtwill: removed

fendermint/app/src/main.rs
cryptoAtwill: yeap, I made it Optional as well. If not set, just log to console.

fendermint/app/options/src/lib.rs
cryptoAtwill: removed default value, see below comment.

docker/runner.Dockerfile
cryptoAtwill: I removed as it's not needed anymore. But I think one day we should not use root to execute docker

fendermint/vm/interpreter/src/chain.rs
cryptoAtwill: I changes as much as possible

adlrocha (approved):

fendermint/app/src/main.rs
```diff
@@ -30,6 +31,88 @@ async fn main() {
...
+            Target::from("fendermint_vm_interpreter::chain"),
+            topdown_logger,
+        );
+        config
+            .appenders
+            .insert(AppenderId::from("topdown"), topdown_appender);
+
+        config.loggers.insert(
+            Target::from("fendermint_vm_interpreter"),
+            interpreter_logger,
```
aakoshh, line 101: Are you still changing this so that it can be literally _any_ package under `fendermint*`? Is that possible without enumerating all of them?

cryptoAtwill: I need to test first, let me try

fendermint/app/src/main.rs
```diff
@@ -30,6 +31,88 @@ async fn main() {
...
+            Target::from("fendermint_vm_topdown"),
+            topdown_logger.clone(),
+        );
+        config.loggers.insert(
+            Target::from("fendermint_vm_interpreter::chain"),
+            topdown_logger,
+        );
+        config
+            .appenders
+            .insert(AppenderId::from("topdown"), topdown_appender);
```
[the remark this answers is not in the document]
cryptoAtwill: I havent looked very deep into `trace4rs`, my guess is when an event is emitted, it checks if there are any `Target` watching this event, then target has a logger, and logger has an appender.

fendermint/app/src/main.rs
```diff
@@ -30,6 +31,88 @@ async fn main() {
...
+
+        config
+            .loggers
+            // "" matches everything
```
aakoshh, line 101: 3rd parties too? So now wildcard available for us?

fendermint/vm/topdown/src/sync.rs
```diff
@@ -315,7 +326,7 @@ async fn last_recorded_data(
         LastRecordedBlock::Empty => Ok(None),
         LastRecordedBlock::FilledBlock { height, hash } => Ok(Some((height, hash))),
         LastRecordedBlock::NullBlock(height) => {
-            tracing::info!("last recorded height {height} is a null block");
+            tracing::info!(height = height, "last recorded height is a null block");
```
aakoshh, line 334:
```suggestion
            tracing::info!(height, "last recorded height is a null block");
```

fendermint/app/src/main.rs
cryptoAtwill: yeah, it matches too many, I had to specify one by one.

Cargo.toml
```diff
@@ -85,6 +85,8 @@ tracing = "0.1"
 tracing-subscriber = "0.3"
 url = "2.4.1"
 zeroize = "1.6"
+trace4rs = "0.5.1"
```
aakoshh, line 88: Do you know if this is available on Github somewhere? I can't find a link at https://crates.io/crates/trace4rs

fendermint/app/src/main.rs
```diff
@@ -30,6 +31,91 @@ async fn main() {
...
+            .appenders
+            .insert(AppenderId::from("topdown"), topdown_appender);
+
+        // maybe there are better ways to handle this, but * does not seem to be supported
+        config
+            .loggers
+            .insert(Target::from("fendermint_app"), debug_logger.clone());
+        config
+            .loggers
+            .insert(Target::from("fendermint_vm_interpreter"), debug_logger);
```
aakoshh, line 105: https://docs.rs/trace4rs/0.5.1/src/trace4rs/handle/loggers.rs.html#91 suggests that a simple `Target::from("fendermint")` should work.

Cargo.toml
cryptoAtwill: https://github.com/imperva/trace4rs

fendermint/vm/topdown/src/sync.rs
```diff
@@ -403,37 +425,54 @@ async fn parent_views_in_block_range(
...
     let block_hash_res = parent_proxy
         .get_block_hash(height)
         .await
         .map_err(|e| Error::CannotQueryParent(e.to_string(), height))?;
     if block_hash_res.parent_block_hash != *previous_hash {
         tracing::warn!(
-                "parent block hash at {height} is {:02x?} diff than previous hash: {previous_hash:02x?}",
-                block_hash_res.parent_block_hash
-            );
+            heigit = hex::encode(&block_hash_res.parent_block_hash),
```
aakoshh, line 472:
```suggestion
            height = hex::encode(&block_hash_res.parent_block_hash),
```

fendermint/app/src/main.rs
cryptoAtwill: funny thing is, with this, all logs are directly to `debug` and `console` is empty

fendermint/app/src/main.rs
```diff
@@ -30,6 +31,91 @@ async fn main() {
...
+        config
+            .appenders
+            .insert(AppenderId::from("debug"), debug_appender);
+    }
+
+    let handle = Arc::new(Handle::try_from(config).unwrap());
```
aakoshh, line 111:
```suggestion
    let handle = Arc::new(Handle::try_from(config)?);
```

fendermint/app/src/main.rs
aakoshh: Maybe because it's set as a "default" which could mean "if nothing else matched"? Perhaps insert it explicitly with the "" target?

aakoshh: Yes :point_right:  https://docs.rs/trace4rs/0.5.1/src/trace4rs/handle/layers.rs.html#139-149

fendermint/app/src/main.rs
```diff
@@ -30,6 +31,88 @@ async fn main() {
...
+            .appenders
+            .insert(AppenderId::from("topdown"), topdown_appender);
+
+        // maybe there are better ways to handle this, but * does not seem to be supported
+        config
+            .loggers
+            .insert(Target::from("fendermint"), debug_logger);
+        config
+            .appenders
+            .insert(AppenderId::from("debug"), debug_appender);
```
aakoshh, line 105: Yeah so why don't we insert a clone of the `default` console logger with an `Target::from("fendermint")` and the `INFO` level?

cryptoAtwill: yeah, but it seems to me this is a bug or sth, we should not be able to do that though. But I added this.

aakoshh: What's a bug? Why shouldn't we be able to set a logger that captures everything and logs to the console if we want to?

cryptoAtwill: I mean we need to do this:
```rust
        config
            .loggers
            .insert(Target::from("fendermint"), debug_logger);
        config.loggers.insert(Target::from("fendermint"), default);
```
but ideally we should not need `config.loggers.insert(Target::from("fendermint"), default);` as we already set default.

aakoshh (approved):
Thank you for all the changes!

PR #434, opened 2023-11-15
DEBUG: Gas logging and a tweak to set prices to 0 for initial estimation

aakoshh:
The initial estimation sets the gas limit to `BLOCK_GAS_LIMIT` which could result in "insufficient funds". This PR changes it to zero for the duration of `estimate_gassed_msg` and then restores it at the end.


fendermint/vm/interpreter/src/fvm/query.rs
```diff
@@ -163,10 +164,22 @@ where
         // Setting BlockGasLimit as initial limit for gas estimation
         msg.gas_limit = BLOCK_GAS_LIMIT;
 
+        // With unlimited gas we are probably better off setting the prices to zero.
```
adlrocha, line 167: Good catch!

adlrocha (approved):

dnkolegov (approved):

PR #438, opened 2023-11-17
FM-367: Handle snapshot offer

### LINKED ISSUE #367: Choose snapshot to load
Implement the method in ABCI which picks a snapshot from the ones offered by peers.

aakoshh:
Closes #367 

Implement the `offer_snapshot` ABCI method by parsing the manifest back from the snapshot and accepting it if the application hash in it (coming from a light client) matches what the FVM state parameters in the snapshot. 

For now if the offer cannot be validated we return `Reject` rather than `RejectSender`; it might just be a bug, rather than any bad intentions. 

The PR moves the notification of the manager from `commit` to `begin_block`, so that the snapshot height is associated with the block where the `app_hash` will match the snapshot content. This is so we can validate the `OfferSnapshot::app_hash` which is something CometBFT acquired for us at `OfferSnapshot::height` against the FVM parameters in the manifest. 

This is a change from https://github.com/consensus-shipyard/fendermint/pull/389 but also reflects what I did in https://github.com/consensus-shipyard/fendermint/pull/420 (which is not merged into this branch).


fendermint/app/src/app.rs
```diff
@@ -672,6 +652,16 @@ where
         let db = self.state_store_clone();
         let state = self.committed_state()?;
         let mut state_params = state.state_params.clone();
+
+        // Notify the snapshotter. We don't do this in `commit` because *this* is the height at which
+        // this state has been officially associated with the application hash, which is something
+        // we will receive in `offer_snapshot` and we can compare. If we did it in `commit` we'd
```
adlrocha, line 658: This is done to ensure that the `state_params` that we are requesting belongs to the right height because in `end_block` we haven't yet computed the up to date `state_params`, right?

adlrocha (approved):

PR #439, opened 2023-11-17
FIX: Update ethers with new fields and use cargo install --locked

aakoshh:
Started getting errors in CI about new fields in ethers: https://github.com/consensus-shipyard/fendermint/actions/runs/6906307060/job/18791062964 

Perhaps the `Cargo.lock` file doesn't apply on the test? That would explain why it takes long to compile even after the application has been built :thinking: 

Actually in [both](https://github.com/consensus-shipyard/fendermint/actions/runs/6906529196/job/18791777648?pr=435) failed CI cases it failed in the docker build, so it might be related to https://github.com/rust-lang/cargo/issues/7169

Okay, the PR will also update the `cargo install` commands to use `--locked` so that it respects the `Cargo.lock` file. Going to test that out in https://github.com/consensus-shipyard/fendermint/pull/435 which doesn't include the `cargo update` I did here.

### DIFF
--- a/fendermint/eth/api/src/conv/from_tm.rs
+++ b/fendermint/eth/api/src/conv/from_tm.rs
@@ -195,6 +195,9 @@ pub fn to_eth_block(
         size: Some(size),
         gas_limit,
         gas_used,
+        blob_gas_used: None,
+        excess_blob_gas: None,
+        parent_beacon_block_root: None,
     };
 
     Ok(block)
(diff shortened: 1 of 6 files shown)


cryptoAtwill (approved):

fendermint/eth/api/src/conv/from_tm.rs
```diff
@@ -360,6 +363,9 @@ where
         withdrawals_root,
         withdrawals,
         other,
+        blob_gas_used,
```
adlrocha, line 366: It looks like https://eips.ethereum.org/EIPS/eip-4844 has already been implemented in ethers. Hopefully it doesn't break anything for us :crossed_fingers:

adlrocha (approved):

PR #441, opened 2023-11-21
FM-362: Save snapshot chunks

### LINKED ISSUE #362: Take snapshot chunks in ABCI
Implement a mechanism to build up a snapshot CAR file from chunks sent to us by remote peers. Consider how we could verify them on the fly.

aakoshh:
Closes #362 

* Split the snapshot crate into multiple submodules to manage code size
* Change `App::offer_snapshot` to put into the state a new `SnapshotDownload` record which shows which manifest we are downloading to what temporary directory
* `SnapshotDownload` cleans up the temporary directory when the last clone of it is dropped, e.g. if CometBFT restarts with a new snapshot
* Add `App::apply_snapshot_chunk` to ingest new chunks of a snapshot, saving them to the temporary directory and validating the checksum when they are all present

Not done in this PR:
* Tests are deferred to https://github.com/consensus-shipyard/fendermint/issues/363 
* Any locking - the docs say CometBFT calls `apply_snapshot_chunk` sequentially, but I'm not sure if it waits before passing the next chunk, or is it similar to `end_block` which was called while `deliver_tx` was still running. We'll see.
* Here we just save all the snapshot parts and verify the checksum at the end; the import of the snapshot is deferred to https://github.com/consensus-shipyard/fendermint/issues/365


fendermint/vm/snapshot/src/client.rs
```diff
@@ -0,0 +1,151 @@
...
+                    Ok(()) => {
+                        let next_index = index + 1;
+                        cd.next_index.write(next_index)?;
+
+                        if next_index == cd.manifest.chunks {
+                            // Verify the checksum then load the snapshot and remove the current download from memory.
+                            match manifest::parts_checksum(cd.download_dir.as_ref()) {
+                                Ok(checksum) => {
+                                    if checksum == cd.manifest.checksum {
+                                        // TODO: Import Snapshot.
```
aakoshh, line 122: I'm not sure yet if we should import the snapshot here, or make another call from the `App`. In any case, the `App` will have to write to its app state history, so we'll have to return the full `SnapshotManifest` so it can construct the necessary record.

aakoshh: This change was done in https://github.com/consensus-shipyard/fendermint/pull/443/commits/6e3c5873d536a858e5b255e7f4198fbbadc87ab2

adlrocha: I had a few questions on how you import and if you already persist this snapshot so it is available to offer it to other peers, but I see this is addressed in the follow-up PR. Will ask my questions there if they are still not solved. :pray:

adlrocha (approved):

PR #443, opened 2023-11-22
FM-365: Commit snapshot

### LINKED ISSUE #365: Commit snapshots to the blockstore
Consider importing the CAR file to a temporary blockstore before committing potential spam to storage. This is a similar issue to bitswap: maybe the CAR file we receive has data that isn't supposed to be in the actor storage; definitely the metadata part shouldn't be in there.

We probably can't store everything in memory, but we could write it into a different namespace in the KV store and then import them starting from the state root hash, finally delete the namespace we imported into.

aakoshh:
Closes #365 

* When the last chunk is saved we return a `SnapshotItem`
* The `SnapshotItem::import` is used to load the data into a blockstore and validate that the imported metadata matches.
* If successful the `App` state is updated to the manifest. 

Doesn't contain: 
* Tests, deferred to https://github.com/consensus-shipyard/fendermint/issues/363
* The clearing of the `/tmp` directory - we can still investigate the snapshot contents after we're done
* The offering of the imported snapshot for others - we'll offer the next snapshot we make, but not this, although we could if we copied the contents from `/tmp` to the persistent place
* Anything clever like importing the CAR file into a temporary namespace in RocksDB, or trying to _not_ import the artificial metadata field. This is a security vulnerability because we allow potentially bogus CIDs to enter the state store, which could lead to consensus failure if some smart contract computation depends on whether something is present or not.

### DIFF
--- a/fendermint/app/src/app.rs
+++ b/fendermint/app/src/app.rs
@@ -887,7 +887,14 @@ where
                     tracing::info!(?manifest, "received snapshot offer");
                     // We can look at the version but currently there's only one.
                     match atomically_or_err(|| client.offer_snapshot(manifest.clone())).await {
-                        Ok(()) => {
+                        Ok(path) => {
+                            tracing::info!(
+                                download_dir = path.to_string_lossy().to_string(),
+                                height = manifest.block_height,
+                                size = manifest.size,
+                                chunks = manifest.chunks,
+                                "downloading snapshot"
+                            );
                             return Ok(response::OfferSnapshot::Accept);
                         }
                         Err(SnapshotError::IncompatibleVersion(version)) => {
@@ -919,13 +926,14 @@ where
 
         if let Some(ref client) = self.snapshots {
             match atomically_or_err(|| {
-                client.apply_chunk(request.index, request.chunk.clone().into())
+                client.save_chunk(request.index, request.chunk.clone().into())
             })
             .await
             {
-                Ok(completed) => {
-                    if completed {
+                Ok(snapshot) => {
+                    if let Some(_snapshot) = snapshot {
                         tracing::info!("received all snapshot chunks");
+                        // TODO: Import the snapshot.
                     }
                     return Ok(response::ApplySnapshotChunk {
                         result: response::ApplySnapshotChunkResult::Accept,
--- a/fendermint/app/src/app.rs
+++ b/fendermint/app/src/app.rs
@@ -958,6 +958,15 @@ where
                         state.block_height = snapshot.manifest.block_height;
                         state.state_params = snapshot.manifest.state_params;
                         self.set_committed_state(state)?;
+
+                        // TODO: We can remove the `current_download` from the STM
+                        // state here which would cause it to get dropped from /tmp,
+                        // but for now let's keep it just in case we need to investigate
+                        // some problem.
+
+                        // We could also move the files into our own snapshot directory
+                        // so that we can offer it to others, but again let's hold on
+                        // until we have done more robust validation.
                     }
                     return Ok(response::ApplySnapshotChunk {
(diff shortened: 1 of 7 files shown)


fendermint/vm/snapshot/src/state.rs
```diff
@@ -70,6 +74,59 @@ impl SnapshotItem {
...
+
+        // 2. Import the contents.
+        let result = Snapshot::read_car(&car_path, store).await;
+
+        // 3. Remove the restored file.
+        std::fs::remove_file(&car_path).context("failed to remove CAR file")?;
+
+        let snapshot = result.context("failed to import the snapshot into the blockstore")?;
+
+        // 4. See if we actually imported what we thought we would.
```
adlrocha, line 106: What would happen if the validation fail and we've already imported the snapshot? I guess you would need to import a valid snapshot, and the invalid blocks would still be stored in the blockstore, right? I don't think is critical at this point, but worth tracking it somewhere, as several import failures could make the blockstore size increase  significantly. Can't we do the validation before importing the CAR?

adlrocha (approved):

fendermint/app/src/app.rs
```diff
@@ -870,13 +877,47 @@ where
...
+                        state.block_height = snapshot.manifest.block_height;
+                        state.state_params = snapshot.manifest.state_params;
+                        self.set_committed_state(state)?;
+
+                        // TODO: We can remove the `current_download` from the STM
+                        // state here which would cause it to get dropped from /tmp,
+                        // but for now let's keep it just in case we need to investigate
+                        // some problem.
+
+                        // We could also move the files into our own snapshot directory
```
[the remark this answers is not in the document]
aakoshh: To be honest it's mostly laziness, but I also think we should first try to resume syncing with the chain before we re-transmit this snapshot. I'm not exactly sure whether it's possible to do too much validation on the snapshot contents. I validate the checksum, but I am not sure whether a particular snapshot chosen by CometBFT necessarily has more than one validator behind it, or it might be just a one-off, which would allow somebody to smuggle in illegal CIDs into the state store, and then later cause consensus failure by deploying a contract that looks it up, finding it one the store of some, but not other validators. 

IIRC this would only affect Wasm actors, not EVM, and by the time it's relevant the FVM might offer protection against it. But it reminds me that I should try to remove even the metadata record from the blockstore after the import.

fendermint/vm/snapshot/src/state.rs
aakoshh: That's right, the unwanted blocks would be left in the blockstore. Because they have been imported into the state store directly, the only way to remove them would be to implement garbage collection. 

I actually included such thoughts into this ticket as loading the data into a staging area inside RocksDB (a separate namespace/collection) and only if successful copying it over to their final destination, or dropping the collection if the import failed. However, I don't think I have the necessary API exposed at the moment and didn't want to complicate things while we don't even have the happy path working. 

> Can't we do the validation before importing the CAR?

To be honest I was only following what Will's `Snapshot` lets me do easily. I see it uses `load_car_unchecked` which doesn't even verify CIDs; I'll change that. By the looks of it it's generally not easy to validate a CAR file: in our case the only CID in the header is the one of the metadata record, which only points at the CID of some data records, which is a tuple that is an internal implementation detail of the `Snapshot`. It is actually the [last thing](https://github.com/consensus-shipyard/fendermint/blob/3cf934c5fe7388814396b373184625938683b45d/fendermint/vm/interpreter/src/fvm/state/snapshot.rs#L192) it puts in the file I think, but to get there you'd have to enumerate everything and then parse it correctly - and this is just the V1 snapshot. 

So I landed on the "trust, but verify" approach.

I checked what kind of validations I noted down in https://github.com/consensus-shipyard/fendermint/issues/364 but they are also post-import.

PR #446, opened 2023-11-27
FIX: Use SubnetRegistryDiamond

aakoshh:
Use the new `SubnetRegistryDiamond` and its facets. 

Depends on https://github.com/consensus-shipyard/ipc/pull/406 and https://github.com/consensus-shipyard/ipc-solidity-actors/pull/298


adlrocha:
@aakoshh, this is the PR that introduced the upgradable registry, and here's the deployment logic: https://github.com/consensus-shipyard/ipc-solidity-actors/pull/278/files#diff-7ab9152e789f8cdb3c93eaf4f0ea12329f38479078f153f6b4f7cca0096d91fd

adlrocha (approved):

PR #449, opened 2023-11-30
Reset provider

cryptoAtwill:
We should reset provider, i.e. clean cache, reset last committed and polling trackers after syncing with peer and also a potential reorg.

For potential reorg, we had this for block height check but not after block hashes are different. We should reset syncer and try from scratch (what else can we do beside crashing app).

After peer syncing, we need to reset provider because with peer syncing, we might have the last committed finality way ahead of the cache polled heights. Clean cache and restart syncer will make sure syncer is up to date.

### DIFF
--- a/fendermint/vm/topdown/src/sync/pointers.rs
+++ b/fendermint/vm/topdown/src/sync/pointers.rs
@@ -31,6 +31,11 @@ impl SyncPointers {
     pub fn set_tail(&mut self, height: BlockHeight, hash: BlockHash) {
         self.tail = Some((height, hash));
     }
+
+    pub fn reset(&mut self, head: BlockHeight) {
+        self.tail = None;
+        self.head = head;
+    }
 }
 
 impl Display for SyncPointers {
--- a/fendermint/vm/topdown/src/sync/syncer.rs
+++ b/fendermint/vm/topdown/src/sync/syncer.rs
@@ -89,7 +89,7 @@ where
                 chain_head,
                 "reorg detected from height"
             );
-            return self.reset_cache().await;
+            return self.reset().await;
         }
 
         if !self.has_new_blocks(chain_head) {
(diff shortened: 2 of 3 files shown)


fendermint/vm/topdown/src/sync/pointers.rs
```diff
@@ -31,6 +31,11 @@ impl SyncPointers {
     pub fn set_tail(&mut self, height: BlockHeight, hash: BlockHash) {
         self.tail = Some((height, hash));
     }
+
+    pub fn reset(&mut self, head: BlockHeight) {
+        self.tail = None;
+        self.head = head;
```
adlrocha, line 37: As discussed sync, I think with the new model where we enforced the defer execution of side-effects for top-down there is no need to use a `head` and a `tail` here.

fendermint/vm/topdown/src/sync/pointers.rs
```diff
@@ -31,6 +31,11 @@ impl SyncPointers {
     pub fn set_tail(&mut self, height: BlockHeight, hash: BlockHash) {
         self.tail = Some((height, hash));
     }
+
+    pub fn reset(&mut self, head: BlockHeight) {
```
adlrocha, line 35: Here you are just resetting the pointers. Is there no need to clean the content from the cache? We may be leaking outdated information by just changing the pointers, right?

fendermint/vm/topdown/src/sync/syncer.rs
```diff
@@ -118,11 +107,54 @@ where
...
+                    prev_non_null_height
+                )
+            };
 
+            Ok((latest_height, hash))
+        })
+        .await
+    }
+
+    /// Poll the next block height. Returns finalized and executed block data.
```
adlrocha, line 152: Maybe it is worth adding a comment here and be explicit about the height and data returned. 
- We get `height` and return the state at `parent_block` right?

adlrocha (approved):
LGTM. Thank you for simplifying the cache logic a lot :pray:

PR #455, opened 2023-12-06
FIX: Do not fail if status doesn't respond in the interpreter

aakoshh:
Follow up for https://github.com/consensus-shipyard/fendermint/pull/426

I missed this part when making that change: it still tries to connect to CometBFT in `end_block`. This change ignores any error as a synonym for catching up.


cryptoAtwill:
nice, this should resolve the issue I encountered in localnet.

cryptoAtwill (approved):

PR #457, opened 2023-12-07
Fall back in proposal checking

cryptoAtwill:
Add fallback in proposal checking, i.e. if the proposal is not found in cache, do io.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4679,7 +4679,7 @@ checksum = "59ce5ef949d49ee85593fc4d3f3f95ad61657076395cbbce23e2121fc5542074"
 [[package]]
 name = "ipc-identity"
 version = "0.1.0"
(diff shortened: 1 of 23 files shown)


aakoshh:
Can you perhaps rebase your feature branch so that this PR doesn't contain snapshot related changes?

aakoshh:
I strongly disagree with this PR, as I have said the other day. We should not do IO during voting.

fendermint/vm/topdown/src/finality/fetch.rs
```diff
@@ -108,15 +109,55 @@ impl<T: ParentQueryProxy + Send + Sync + 'static> ParentViewProvider for CachedF
...
+        .await
+        {
+            return Ok(check);
+        }
+
+        match retry!(
+            self.config.exponential_back_off,
+            self.config.exponential_retry_limit,
+            self.parent_client.get_block_hash(proposal.height).await
+        ) {
```
aakoshh, line 143: Doing exponential backoff during voting would be detrimental to liveness.

PR #458, opened 2023-12-07
Patch concurrent update error

cryptoAtwill:
This fixes a concurrent update error. It is possible that a validator is catching up and received a finality from peer that is higher than the value stored in cache. In this case, there is no need to update cache.

### DIFF
--- a/fendermint/vm/topdown/src/finality/null.rs
+++ b/fendermint/vm/topdown/src/finality/null.rs
@@ -63,6 +63,17 @@ impl FinalityWithNull {
     /// Clear the cache and set the committed finality to the provided value
     pub fn reset(&self, finality: IPCParentFinality) -> Stm<()> {
         self.cached_data.write(SequentialKeyCache::sequential())?;
+
+        if let Some(p) = self.last_committed_finality.read_clone()? {
+            if p.height > finality.height {
+                tracing::warn!(
+                    last = p.to_string(),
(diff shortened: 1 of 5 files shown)


fendermint/vm/topdown/src/finality/null.rs
```diff
@@ -63,6 +63,17 @@ impl FinalityWithNull {
     /// Clear the cache and set the committed finality to the provided value
     pub fn reset(&self, finality: IPCParentFinality) -> Stm<()> {
         self.cached_data.write(SequentialKeyCache::sequential())?;
+
+        if let Some(p) = self.last_committed_finality.read_clone()? {
```
aakoshh, line 67: I think `.read()?` should suffice here.

fendermint/vm/topdown/src/finality/null.rs
```diff
@@ -71,6 +82,17 @@ impl FinalityWithNull {
         height: BlockHeight,
         maybe_payload: Option<ParentViewPayload>,
     ) -> StmResult<(), Error> {
+        if let Some(p) = self.last_committed_finality.read_clone()? {
```
aakoshh, line 85:
```suggestion
        if let Some(p) = self.last_committed_finality.read()? {
```
Here too, the value isn't returned, so you can access it through the `Arc`

PR #460, opened 2023-12-07
FIX: Accept 'input' or 'data' or both in transaction request

aakoshh:
Equivalent of https://github.com/filecoin-project/lotus/pull/11471

Introduces an `Eip1559TransactionRequestCompat` type next to our `TypedTransactionRequestCompat` which we use to deserialize JSON requests. It accepts `input` or `data`, preferring `input`, before being mapped to the `ethers` type. 

Instead of adding both fields I used the approach in I found [here](https://users.rust-lang.org/t/how-can-i-handle-duplicate-fields-when-specifying-multiple-aliases-using-serde/46426/6). 

I also tried with `#[serde(rename="input", alias = "data")]` but that does not accept duplicate fields.

Eventually ended up with Raul's suggestion of adding `input` but using `flatten` to wrap the original, which requires the minimum amount of code. 

Also opened a PR in `ethers` but ended up only adding an `alias = "input"` to keep things simple. I think the trend is that clients will only send one value or the other, at least the double sending is being fixed in `web3.js`, so the added complexity of supporting both fields seems like an overkill. We can keep our `Compat` stuff to support both fields though.

### DIFF
--- a/fendermint/eth/api/src/apis/eth.rs
+++ b/fendermint/eth/api/src/apis/eth.rs
@@ -1057,19 +1057,77 @@ pub async fn unsubscribe<C>(
     uninstall_filter(data, Params((filter_id,))).await
 }
 
-use params::{EstimateGasParams, SubscribeParams};
-
-use self::params::TypedTransactionCompat;
+use params::{EstimateGasParams, SubscribeParams, TypedTransactionCompat};
 
 mod params {
     use ethers_core::types::transaction::eip2718::TypedTransaction;
     use ethers_core::types::Eip1559TransactionRequest;
     use ethers_core::types::{self as et, Eip2930TransactionRequest, TransactionRequest};
-    use serde::{Deserialize, Serialize};
+    use serde::{Deserialize, Deserializer};
 
     use crate::state::WebSocketId;
 
-    #[derive(Serialize, Deserialize, Clone, PartialEq, Eq, Debug)]
+    /// Deserialize either the `input` or the `data` field, preferring the former.
+    fn deserialize_input_or_data<'d, D: Deserializer<'d>>(
+        d: D,
+    ) -> Result<Option<et::Bytes>, D::Error> {
+        #[derive(Deserialize)]
+        struct InputOrData {
+            input: Option<et::Bytes>,
+            data: Option<et::Bytes>,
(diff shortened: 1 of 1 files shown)


aakoshh:
Opened an upstream PR: https://github.com/gakonst/ethers-rs/pull/2697

fendermint/eth/api/src/apis/eth.rs
```diff
@@ -1057,19 +1057,77 @@ pub async fn unsubscribe<C>(
...
+        #[serde(rename = "maxPriorityFeePerGas", default)]
+        pub max_priority_fee_per_gas: Option<et::U256>,
+
+        #[serde(rename = "maxFeePerGas", default)]
+        pub max_fee_per_gas: Option<et::U256>,
+
+        #[serde(default, rename = "chainId")]
+        /// Chain ID (None for mainnet)
+        pub chain_id: Option<et::U64>,
+    }
```
raulk, line 1111: In my local version of this this fix, I had followed a simpler approach of creating Compat structures for all transaction types, and "embedding" the original type in an `orig` field with `#[deref]` and `#[serde(flatten)]`, then only adding the field for input. This solution has us copying the entire type from upstream, which is not bad, but a bit duplicative.

raulk (approved):
LGTM but would add `input` parsing for all transaction types supported by Fendermint.

fendermint/eth/api/src/apis/eth.rs
```diff
@@ -1079,7 +1137,7 @@ mod params {
         #[serde(rename = "0x00", alias = "0x0")]
         Legacy(TransactionRequest),
```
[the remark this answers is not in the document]
aakoshh: Ah, thanks, I forgot we accept this format for gas estimation.

fendermint/eth/api/src/apis/eth.rs
aakoshh: Hm, that's a nice idea, thanks, I'll try.

aakoshh: Changed it

raulk (commented):
Final implementations is basically what I had locally, so post factum LGTM again.

PR #468, opened 2023-12-21
build: lock actors to last pre-refactor commit.

raulk:
Hopefully this will fix the build for now. Not worth spending too much time discussing this because these out-of-sync problems will vanish with the monorepo.


raulk:
Self-merging to unblock the build and resume Docker image publishing.

PR #469, opened 2023-12-21
BUILD: Point at pre-audit versions of dependencies

aakoshh:
Pointing the code at dependencies before contracts were radically rewritten and broke everything.


aakoshh:
Hm, just seen that https://github.com/consensus-shipyard/fendermint/pull/468 passed. Maybe it didn't work for me because I also fixed the `ipc` at an earlier point :thinking:

